PR #5, opened 2022-06-17
add co2 box

alexandreoliv:


ArneTR:
Have automatically closed the branch by manual merge of the email.

Please use not .toPrecison() but .toFixed() as the former will alter the length of the number in total

PR #6, opened 2022-06-17
save usage_scenario to projects table

alexandreoliv:


ArneTR:
With this change the `obj` variable will not be defined on *cron* mode. 

Let's discuss the refactoring in the office on Monday

PR #18, opened 2022-08-15
Bugfixes rapl c improvements

dan-mm:


dan-mm:
not main, should be making PR to dev

PR #19, opened 2022-08-15
Bugfixes rapl c improvements

dan-mm:
- Get user_id dynamically
- refactor rapl code so that the main loop is leaner
- use #define variables instead of booleans


ArneTR:
Perfect

PR #20, opened 2022-08-30
Tdp reporter

dan-mm:
Delimiter for the csv file is |

Only pipe.

Don't make the same mistake I did and accidentally leave on ; as a delimiter when viewing the csv file in excel or whatever.


ArneTR:
Has been moved to separate repository: https://github.com/green-coding-berlin/spec-power-model

PR #57, opened 2022-12-05
Powerspy2

ribalba:
Pretty much the first initial version. We should see how much more work we want to put into this.

### DIFF
--- a/frontend/js/stats.js
+++ b/frontend/js/stats.js
@@ -63,6 +63,11 @@ const metrics_info = {
       unit: 'us',
       SI_conversion_factor: 1,
       unit_after_conversion: 'us'
+  },
+  psu_energy_powerspy2: {
+    unit: 'mW',
+    SI_conversion_factor: 1000,
+    unit_after_conversion: 'W'
   }
 }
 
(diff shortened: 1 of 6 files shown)


ribalba:
We should think about doing the rfcomm connections in our python code. See https://github.com/pybluez/pybluez/blob/master/examples/simple/rfcomm-server.py unfortunately it looks like pybluez isn't maintained anymore.

ArneTR:
I could get it running and it works fine.

However the Wattage values seem to be incorrect.

I have run a test in our lab with the DC Measurement card attached:
- Stress: https://metrics.green-coding.org/stats.html?id=03053bc5-3321-4c4f-a9fb-f8f33b624850
- Idle: https://metrics.green-coding.org/stats.html?id=55701a88-c81e-4fe0-a09e-7d1ba6d88182

As you can see the Powerspy2 claims to sense 4 W on ilde and 5.5 W on load.

However the DC provider tells 12.86 W on idle and ~25 W on load. 

I have verified with an external power meter and AC should be around 29 W on load for my test machine. See also this old measurement on the same machine with another AC provider that verifies this: https://metrics.green-coding.org/stats.html?id=0f0e89ab-f7f5-4bf8-a6ba-40efeeca50a6

ArneTR:
I ran the tests as discussed with the *stress-step* example application to see if the wrong offset of the powerspy2 is of quadratic nature or more complex.

You can see the results of the initial tests here:
- https://metrics.green-coding.org/stats.html?id=67a6131e-a6f6-4101-a644-59501223bdaa
Sadly the `psu_energy_dc_system` graph has a bit of a display proble, but by hovering you see that the values seem to about quadratic in difference to the powerspy2 ones.

I looked at the code and changed the line 120:
```
# old
                rmsPower = math.sqrt(int(res[19:27], 16) * self.uscale * self.iscale )
# new
                rmsPower = math.sqrt( (int(res[19:27], 16))**2 * self.uscale * self.iscale )
```

The result is here and looks quite good:
- https://metrics.green-coding.org/stats.html?id=0a6c6bcc-dd77-4c3d-83b5-8c7a60816d81

**Please make a pull, as I merged dev in  to incorporate the latest changes and gave the powerspy2 a unit attribute**

When looking at the original documentation it looks like the implementation is actually specifying that it should be squared, but this is not done.
This probably is also the case for all the other values like `rmsCurrent`, `rmsVoltage` etc., but I wanted to double check with you ... maybe the value was squared beforehand somewhere and I am just missing something.

However the resulting value is now by 3 orders of magnitude to high ... 

Up for discussion on Wednesday.

PR #98, opened 2022-12-28
Makes code pylint

ribalba:


ribalba:
@ArneTR took a little longer as thought and was quite some manual editing necessary which normally introduces some errors. I also tried to change all the quotes to single for strings that don't change and double for the others. 

I also added some description to contributing.md to write down most of my thoughts. All is up for discussion :)

ArneTR:
Thx, I will reviews it ASAP.

One question on the single-quote / double-quote thing. Is this the current standard to still use both, but change to double quotes for interpolatable strings? Technically Python can do f-strings with single- or double quotes, doesn't it?

ribalba:
Yes, I think python doesn't really care if single or double. I used the standard that any string that will not change is in single and anything that will is in double. So f-strings are in double as they are changed by the interpreter.

PR #108, opened 2023-01-04
Makes the hardware_info script run certain parts as root

### LINKED ISSUE #107: Make getting the sched info a sudo command in get_hardware_info.py

ribalba:
Closes #107


ArneTR:
Looks good!

What is the rationale for running the adding to the `/etc/sudoers` file in the `Makefile` under `lib` rather than doing it directly in the `install.sh`?

The reason why we are doing something similar with the metric providers is that they are thought of the be modular. The same design decision does not hold for the hardware_info.py I would say. And even when it does the sudo hack is rather something we need to make it work with our Python code.
What do you say?

ribalba:
None, refactored. Was something from olden days

ArneTR:
When I merge `dev` in I get a conflict that the kernel scheduling has been moved to the file with sudo need.

This is expected. but also this line was removed:
````
    [rfwr, 'IO scheduling', '/sys/block/sda/queue/scheduler', r'(?P<o>.*)'],

````

What is the reason for that?

I left it in in my test merge and it works fine. Results here: 

Shall we put that back in? if so: Please merge `dev` into this branch from your side and add the line, than merge-squash the pull request.

PR #124, opened 2023-01-13
Install.sh script should show failure more noticeable

djesic:


djesic:
Closing - to be handled in the eventual `bash -> python` rewrite task

PR #131, opened 2023-01-16
126 - Domain and port changes

djesic:
`.org` to `.berlin`
`:8000` to `:9142`


ArneTR:
Good work on the port stuff.

The domain stuff is however not quite how I wanted it to be:
- There we very many references to .org still missing. Did you do a grep?
- The api still had the setting to .org in place. Did you place a fake hostname for testing?

Also it is not domain agnostic in this form. I have to admit that my specification was a bit too short maybe. We will discuss the next steps on Friday, where I would like to explain this more throrough.

Please do not delete the branch for now.

PR #135, opened 2023-01-19
Specify branch

djesic:


ArneTR:
this works just nicely! Also the migration added, I am stoked!

I added your TBD recommendation for the shallow checkout already

PR #148, opened 2023-01-30
Implement IPMI power sensor

djesic:
I got this working on the server, the script output is in the `timestamp value` format as expected.
I'm not 100% sure about the expected unit and there could be some improvements to the bash script in terms of resolution.
I'd like to discuss that and what to write in the readme


djesic:
TODO: readme >> 1 second delay warning
fix the resolution to be passed in in ms

ArneTR:
Also reminder for me in QA: The file should not be called metric_provider_binary but a generic name should be able to be defined in the class itself.

PR #152, opened 2023-01-31
146 port the gmt to mac

### LINKED ISSUE #146: Port the GMT to Mac
Currently a Ubuntu Linux machine is needed to run the GMT. This limits the amount of people that can actually use our tool. So we can port it to Mac so that more people can run the tool. We are aware that the energy measurements will never be any good. See blog article.

ribalba:
Closes #146


ArneTR:
I could get it working, encountered a few roadbumps on the way though:

- the mac provider is not automatically turned on. This is due to the TODO refactoring of the config.yml.example I guess, where we wanted to introduce the new key
- the sed replacement leaves me with "config.yml-e" etc. files. On mac the command must be "sed -i '' -e ..." otherwiese temp files are created
  + I added this to your branch as commit 
- the replacement in the /etc/hosts has not worked for me. The first line was just added to the last line in my file. A newline has to be introduced before here. tee on macos behaves differently. It does not add a newline.
  + I added this to your branch as commit
- On Intel macs the keys are named differently. I added the "package_joules" key and transformed the calculation. Added it as commit to your branch


Now the juicy part! I made a measurement on my MBP 13" on linux and on macOS with your provider. You can find the results in the Dashboard online.

Qualitatively the results look nice! However the time keying is wrong. I did not look into this further and would play it back to you.

For instance in https://metrics.green-coding.berlin/stats.html?id=bbf23cc1-687d-4526-8150-ef762df8a603 it says that the start_measurement (which is set by python) is at 1675333274001283.
The last `time` entry for the metric in the `stats` table is 1675329685285178 . 
As you can see the values do not even overlap. I do not know exactly why the times are so heavily shifted. It seems like this is a timezone issue.

ArneTR:
Replacing line 77 with the following fixed the timeshifiting issue for me:
`cum_time = int(data['timestamp'].replace(tzinfo.utc).timestamp() * 1e9)`

You must include `from datetime import timezone` at the top

Please check if it also works for you.

ribalba:
Just for completeness my solution to the timezone issue:
```
            if cum_time is None:
                # Get the system's local timezone
                local_tz = pytz.timezone(datetime.datetime.now().astimezone().tzname())

                # Set the timezone on the timestamp to UTC
                utc_time = pytz.timezone('UTC').localize(data['timestamp'])

                # Convert the datetime object to the system's local timezone
                local_dt = utc_time.astimezone(local_tz)

                # Convert seconds to nano seconds
                cum_time = int(local_dt.timestamp() * 1e9)
```

This is due to `Lib/plistlib.py` not attaching timezone info. PR to fix this can be found under https://github.com/python/cpython/pull/23384/files

ribalba:
I can't get the tzinfo to work on my system. I keep getting `AttributeError: type object 'datetime.tzinfo' has no attribute 'utc'`

ArneTR:
I will merge this in tomorrow and then also tackle the macos / linux keys in the usage_scenario.yml

ArneTR:
I merged everything in to the new location of the metric_providers folder.

The fixes of the tests I will do tomorrow with Dan.

What is left open is the keying of the providers with linux/macos. I forgot why we needed it exactly, but it works for now.

I also seem to have gotten you the wrong line for the timezone replacement. Maybe I copied a broken version.

The working one is: 
```
                cum_time = int(data['timestamp'].replace(tzinfo=timezone.utc).timestamp() * 1e9)
````

I refactored it to use this one-liner

PR #156, opened 2023-02-05
Changes the packages to be distro dependent

ribalba:


ArneTR:
Had to merge it by hand. Did not work on github

Issue #155, opened 2023-02-05
import in usage_scenario


ribalba:
Great tool: https://github.com/tanbro/pyyaml-include/ but because it allows relative paths this could lead to people reading files they shouldn't

Issue #32, opened 2022-11-07
Should it be possible to use the compose.yml file?


ribalba:
Currently we are duplicating code from the `compose.yml` in the `usage_scenario.yml`. Following ideas were discussed with @ArneTR 

1) Allow an import in the `usage_scenario.yml`. This will increase code complexity in the runner.py and could potentially make the yml file very had to read if we want to use keywords like `setup-commands` which we would have to interweave. 
2) Separate the flow into a new file and allow the usage of `compose.yml` for setup.

PR #157, opened 2023-02-06
Adds include functinality to usage_scenario.yml

ribalba:
Closes: #155 
Closes: #32 

We should still:
- Update one example to include this 
- Test that if we include a compose.yml unwanted settings are ignored.


ArneTR:
Works very nicely. I have done a bit of rework. Details in the team meeting, when we meet next time

ArneTR:
Please do not delete the branch until then

PR #168, opened 2023-02-13
Configurable domain

djesic:


djesic:
Todo docs for this and the macos install script
Wanted to open to PR now to see if this what we we're aiming for.

ArneTR:
Yes, this looks exactly what we are going for!

Is the PR ready for QA? Or what is left to be done?

From a quick look I would say that the ./lib/email_helpers.py:__METRICS_URL__ is not handled yet. Everything else looks fine ... but have not tested without your GO yet

djesic:
I think this is ready for review.
I have not made changes to the `install_mac.sh` - I have no way of testing this

PR #169, opened 2023-02-13
added linux, macos, and common keys for metric providers; gmt now che…

dan-mm:
…cks system architecture and only loads the appropriate metric provider; added get_achitecture to utils

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -33,10 +33,18 @@ def get_pascal_case(in_string):
 def get_metric_providers(config):
     architecture = get_architecture()
     if 'common' in config['measurement']['metric-providers']:
+        if config['measurement']['metric-providers']['common'] is None:
+            raise RuntimeError('No metric providers under \'common\' key in config.yml')
+        if config['measurement']['metric-providers'][architecture] is None:
+            raise RuntimeError(f"No metric providers under '{architecture}' key in config.yml")
         metric_providers = {**config['measurement']['metric-providers'][architecture],\
             **config['measurement']['metric-providers']['common']}
     else:
         metric_providers = config['measurement']['metric-providers'][architecture]
+    return metric_providers
+
+def get_metric_providers_names(config):
+    metric_providers = get_metric_providers(config)
     metric_providers_keys = metric_providers.keys()
     return [(m.split('.')[-1]) for m in metric_providers_keys]
 
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -32,15 +32,23 @@ def get_pascal_case(in_string):
 
 def get_metric_providers(config):
     architecture = get_architecture()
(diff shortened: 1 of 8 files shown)


metric_providers/lm_sensors/abstract_provider.py
```diff
@@ -3,15 +3,19 @@
...
+current_dir = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(f"{current_dir}/../../lib")
+import utils
 from metric_providers.base import BaseMetricProvider
 
 class LmSensorsProvider(BaseMetricProvider):
 
     def _create_options(self):
-        provider_config = GlobalConfig().config['measurement']['metric-providers'][self._provider_config_path]
+        provider_config = GlobalConfig().config['measurement']['metric-providers'][utils.get_architecture()]\
```
ArneTR, line 17: I changed this in the PR to have the static key "linux". I think it makes no sense to have this provider architecture agnostic, when it is only available on linux anyway (see: https://formulae.brew.sh/formula/lm-sensors)

ArneTR (approved):
nice work, tests all work for me. Merged

lib/utils.py
```diff
@@ -29,5 +31,20 @@ def get_pascal_case(in_string):
     return ''.join([s.capitalize() for s in in_string.split('_')])
 
 def get_metric_providers(config):
-    metric_providers_keys = config['measurement']['metric-providers'].keys()
+    architecture = get_architecture()
+    if 'common' in config['measurement']['metric-providers']:
```
ArneTR, line 35: This function will fail if the "common" key does not contain any active providers.

runner.py
```diff
@@ -178,17 +175,25 @@ def import_metric_providers(self):
         config = GlobalConfig().config
 
         print(TerminalColors.HEADER, '\nImporting metric providers', TerminalColors.ENDC)
+
+        # if there's a more elegant way to do this, I don't know it
+        if 'common' in config['measurement']['metric-providers']:
```
ArneTR, line 180: Why don't you use the get_metric_providers function here?

As I see it has to be reduced a bit and the splitting to be done manually afterwards, but then you could reuse that part of the code

ArneTR (changes requested):

metric_providers/lm_sensors/abstract_provider.py
dan-mm: changed to ['linux']

lib/utils.py
dan-mm: good catch, pushed a check against this (and the same was true if there were no providers under the architecture key as well)

runner.py
dan-mm: good call, did the refactor and pushed

PR #177, opened 2023-02-20
161 eco ci badges

dan-mm:


docker/structure.sql
```diff
@@ -57,3 +57,13 @@ CREATE TABLE notes (
     time bigint,
```
ribalba, line 57: I think this belongs into the migration folder.

PR #178, opened 2023-02-20
Setup command debugging

djesic:
If the `debug` flag is enabled, the runner will now pause before each setup command and also allow the user to ignore a failing setup command due to a non-zero return code from the setup command subprocess run.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,8 +76,8 @@ guidelines on when to use or avoid this directive follow:
 
 - instead of disabling `unused-argument`, prefix the arg (or kwarg) with `_`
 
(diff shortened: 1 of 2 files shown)


ArneTR:
I do understand why you needed this for your use-case, however I see this as unintuitive as functionality in the GMT.

This functionality encourages people to write usage_scenario files that only work if the switch is active.

I however want to encourage to have usage_scenario files that always work.

In order to your particular code running you could have just post-fixed the command in the setup-commands with "|| true".
Atm our tool cannot handle shell executions through the setup-commands, but it could be done by starting a shell instead of sending the command pure to docker exec:
https://stackoverflow.com/questions/31438112/bash-docker-exec-file-redirection-from-inside-a-container

PR #186, opened 2023-02-28
Graph for badges

djesic:
The curl call to add data: `curl -X POST "http://api.green-coding.example:9142/v1/ci/badge/add/" -H 'Content-Type: application/json' -d '{"value":"12365","unit":"mJ","repo":"owner-name/test-repo","branch":"test_branch","workflow":"test-workflow","run_id":"2312315","project_id":""}'`

The URL: `http://metrics.green-coding.example:9142/ci.html?repo=owner-name%2Ftest-repo&branch=test_branch&workflow=test-workflow`

What it looks like:
image

I feel like this might not be 100% what we were aiming at, but I'm opening a PR to start a discussion :)

### DIFF
--- a/frontend/ci.html
+++ b/frontend/ci.html
@@ -14,13 +14,13 @@
     <meta name="referrer" content="no-referrer-when-downgrade" />
     <meta name="description" content="Detailed energy statistics for software measurement" />
     <script src="/dist/js/jquery.min.js" defer></script>
+    <script src="/dist/js/echarts.min.js" defer></script>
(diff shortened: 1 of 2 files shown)


ArneTR:
Close to where we want to go. This is a bar chart however. Can you do a line chart?

djesic:
I pushed the code for a line chart, but in my opinion the same data looks nicer with a pie chart.
Maybe the test values used are not representative of the real values coming in?

PR #194, opened 2023-03-03
Refactoring ipmi reporter

### LINKED ISSUE #188: Refactoring IPMI reporter
bash script call and parameter

djesic:
Closes #188


ArneTR:
This code is untested and contained a bug! Please test it on the IPMI server before pushing

djesic:
I did test it on the blockheating server, but only by running the script itself while making the changes

PR #197, opened 2023-03-09
added tests and fixtures for volume loading; refactored test function…

dan-mm:
…s into own file


ribalba:
Is this still up to date? Or can it be closed?

@dan-mm @ArneTR

ArneTR:
This is up to date. I have reviewed it, but it can currently not be merged due to issues with the volume loading

ribalba:
Just to have it written down. The question was why the `test_load_volume_references` was working even if we don't have the code implemented to do so. It turns out that pyyml resolves anchors and references when loading a yml file. God to know for the future. 

@dan-mm @ArneTR

PR #199, opened 2023-03-09
Raise value error on invalid note timestamp

djesic:


djesic:
@dan-mm added you as a reviewer because I've added some unit tests

test/tools/test_save_notes.py
```diff
@@ -0,0 +1,73 @@
...
+        save_notes(
+            "test_project_id",
+            [
+                {
+                    "note": "This is my note",
+                    "timestamp": 'string_instead_of_time',
+                    "detail_name": "Test_Container",
+                }
+            ],
+        )
```
dan-mm, line 28: I would call an explicit assert here based on the expected error. 

an example of how I do this can be found in the test_usage_scenario.py, in test_env_variable_no_skip_or_allow()

something like:

with pytest.raises(ValueError) as err:
        [... does the thing ...]
expected_exception = ''Note timestamp did not match expected format"
assert expected_exception in str(err.value), \
        utils.assertion_info(f"Exception: {expected_exception}", str(err.value))

test/tools/test_save_notes.py
```diff
@@ -0,0 +1,73 @@
...
+        save_notes(
+            "test_project_id",
+            [
+                {
+                    "note": "This is my note",
+                    "timestamp": int(time.time()),
+                    "detail_name": "Test_Container",
+                }
+            ],
+        )
```
dan-mm, line 57: same as above here, i'd do an explicit assert

test/tools/test_save_notes.py
```diff
@@ -0,0 +1,73 @@
...
+        "test_project_id",
+        [
+            {
+                "note": "This is my note",
+                "timestamp": str(int(time.time_ns() / 1000)),
+                "detail_name": "Test_Container",
+            }
+        ],
+    )
+    mock_db.query.assert_called_once()
```
dan-mm, line 44: I actually wouldn't mock this, I'd make the DB call and make sure the note is inserted properly. Our test suite is small enough that the time savings on mocking here is minimal, and this way we make sure the DB call is correct in save_notes

test/tools/test_save_notes.py
```diff
@@ -0,0 +1,73 @@
...
+        "test_project_id",
+        [
+            {
+                "note": "This is my note",
+                "timestamp": int(time.time_ns() / 1000),
+                "detail_name": "Test_Container",
+            }
+        ],
+    )
+    mock_db.query.assert_called_once()
```
dan-mm, line 73: same as above, wouldn't mock this

tools/save_notes.py
```diff
@@ -9,6 +9,16 @@
 def save_notes(project_id, notes):
 
     for note in notes:
+        try:
+            if len(str(note['timestamp'])) != 16:
```
ribalba, line 13: Thinking about this, this could also be done with a regex:
```
    if re.fullmatch(r'\d{16}', note['timestamp']):
```

test/tools/test_save_notes.py
```diff
@@ -0,0 +1,70 @@
...
+CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(f"{CURRENT_DIR}/../../tools")
+sys.path.append(f"{CURRENT_DIR}/../../lib")
+
+# pylint: disable=import-error,wrong-import-position
+from save_notes import save_notes
+import utils
+
+
+def test_invalid_string_timestamp():
```
ribalba, line 16: I think it is good practice to group all related tests together into a unittest. See https://github.com/green-coding-berlin/green-metrics-tool/blob/dev/test/test_yml_parsing.py

test/tools/test_save_notes.py
```diff
@@ -0,0 +1,70 @@
...
+    assert expected_exception in str(err.value), \
+        utils.assertion_info(f"Exception: {expected_exception}", str(err.value))
+
+def test_string_timestamp():
+    save_notes(
+        "e55675c5-8f7b-4d21-a4f7-d2417e23a44b",
+        [
+            {
+                "note": "This is my note",
+                "timestamp": str(int(time.time_ns() / 1000)),
```
ribalba, line 38: In general dynamic content like time shouldn't be used in tests as it might be hard to reproduce an error. This doesn't apply for fuzzing tests. In this case it is ok but just as a note.

test/tools/test_save_notes.py
djesic: I've added parameterization, but am not sure if adding them to a class is needed. I know pytest is ok with both cases, but unsure what is preferred

PR #214, opened 2023-03-23
XSS in frontend

djesic:
Project name is no longer clickable on the Project data/General tab
Neither is anything inside of the usage scenario tab
Notes are also escaped and don't create a link in the database, or frontend.

I used the Python builtin `html.escape()`, so no dependency got introduced.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -4,9 +4,9 @@
 # pylint: disable=wrong-import-position
 
 import faulthandler
-from html import escape
 import sys
 import os
+from html import escape
 
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
@@ -24,8 +24,9 @@ import email_helpers
 import error_helpers
 import psycopg
 import anybadge
-from api_helpers import rescale_energy_value, is_valid_uuid, \
-                        determine_comparison_case, get_phase_stats, get_phase_stats_object, add_phase_stats_statistics
+from api_helpers import (add_phase_stats_statistics, determine_comparison_case,
+                         escape_dict, get_phase_stats, get_phase_stats_object,
+                         is_valid_uuid, rescale_energy_value, safe_escape)
 
 
 # It seems like FastAPI already enables faulthandler as it shows stacktrace on SEGFAULT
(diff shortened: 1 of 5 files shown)


ribalba:
Yes please also add escaping to the API

runner.py
```diff
@@ -266,14 +267,14 @@ def update_and_insert_specs(self):
 
             machine_specs.update(machine_specs_root)
 
-        # Insert auxilary info for the run. Not critical.
+        # Insert auxiliary info for the run. Not critical.
         DB().query("""UPDATE projects
             SET machine_specs=%s, measurement_config=%s, usage_scenario = %s, last_run = NOW()
             WHERE id = %s
             """, params=(
             json.dumps(machine_specs),
```
ribalba, line 275: Please also escape the machine_specs as someone could inject an XSS through adding parameters to the process that are html and then these would be dumped here and also rendered into the frontend

ribalba (commented):
Please also add escaping to 
`async def post_project_add(project: Project):` in `api.py` as the name could be an XSS

As anyone can add a badge to any project we should make sure that the repo is taken from the project and checked with the repo supplied. Otherwise I could just submit a badge to any project and then put the uri to a malicious site. Please quickfix this here. I will write a ticket for next Thursday how to make this more secure.

ribalba:
@djesic Please rebase with dev and check that the checks are still ok now that @dan-mm has merged his new badge stuff.

djesic:
@ArneTR @ribalba 
rebased, escaped the badge stuff and now it's ready to merge

djesic:
Messed up the last rebase - not ready to merge anymore..

api/api.py
```diff
@@ -361,36 +394,55 @@ class CI_Measurement(BaseModel):
...
+            continue
+        if key == 'workflow':
+            measurement.workflow = escape(value)
+            continue
+        if key == 'run_id':
+            measurement.run_id = escape(value)
+            continue
+        if key == 'source':
+            measurement.source = escape(value)
+            continue
```
ribalba, line 436: This could be rewritten to be
```
else_keys = ['repo', 'branch', 'workflow', 'run_id', 'source']
if key in else_keys:
        setattr(measurement, key, escape(value))
```

or you could use:
```
    match value:
        case 'case1' | 'case2':
            return 'This is case 1 or 2'
        case 'case3':
            return 'This is case 3'
        case _:
            return 'This is default case'

```
for the whole block

api/api.py
```diff
@@ -441,6 +493,24 @@ async def get_ci_badge_get(repo: str, branch: str, workflow:str):
...
+        if isinstance(value, list):
+            for item in value:
+                if isinstance(item, str):
+                    dictionary[key] = escape(item, quote=False)
+                    continue
+                if isinstance(item, dict):
+                    escape_dict(item)
+
+    return dictionary
+
```
ribalba, line 513: I don't think this will work properly as you sometimes just call `escape_dict` without assigning it?

```
def escape_dict(dictionary):
    for key, value in dictionary.items():
        if isinstance(value, str):
            dictionary[key] = escape(value, quote=False)
        elif isinstance(value, dict):
            dictionary[key] = escape_dict(value)
        elif isinstance(value, list):
            dictionary[key] = [escape_dict(item) if isinstance(item, dict) else escape(item, quote=False) if isinstance(item, str) else item for item in value]
    return dictionary
```

Something like this might work better. I haven' tested my code just as a reference

ribalba (changes requested):
Could you please also add some tests

api/api.py
```diff
@@ -434,5 +467,56 @@ async def get_ci_badge_get(repo: str, branch: str, workflow:str):
         default_color='green')
     return Response(content=str(badge), media_type="image/svg+xml")
 
+<<<<<<< HEAD
```
ArneTR, line 470: There is a merge marker here

api/api.py
```diff
@@ -434,5 +467,56 @@ async def get_ci_badge_get(repo: str, branch: str, workflow:str):
...
+    return energy_rescaled
+
+def is_valid_uuid(val):
+    try:
+        uuid.UUID(str(val))
+        return True
+    except ValueError:
+        return False
+
+>>>>>>> 0d9c6cd (Escape CI measurement add)
```
ArneTR, line 520: Another merge marker here

PR #229, opened 2023-04-03
Add missing 'provider' to config.yml.example

djesic:
Always enabled metric provider is broken, was missing `.provider` in import

Installs from main will result broken runner.

I will also make this commit to dev


djesic:
I made the PR into main because it was causing the installation to result in a broken state at the time. I could push the fix into dev but not main, and when you merged in dev to main, it made this PR obsolete.

PR #236, opened 2023-04-05
Adds the = notatio for env vars

ribalba:


runner.py
```diff
@@ -398,29 +398,35 @@ def setup_services(self):
...
                             print(TerminalColors.WARNING, warn_message, TerminalColors.ENDC)
                             continue
                         raise RuntimeError(f"Docker container setup environment var key had wrong format. \
-                            Only ^[A-Z_]+$ allowed: {docker_env_var} - Maybe consider using --allow-unsafe \
+                            Only ^[A-Z_]+$ allowed: {env_key} - Maybe consider using --allow-unsafe \
                                 or --skip-unsafe")
 
                     if not self._allow_unsafe and \
-                        re.search(r'^[a-zA-Z_]+[a-zA-Z0-9_-]*$', str(service['environment'][docker_env_var])) is None:
+                        re.search(r'^[a-zA-Z0-9_]+[a-zA-Z0-9_-]*$', env_value) is None:
```
ribalba, line 418: We need to discuss this. Is this a security problem? @ArneTR

ArneTR: The security issue is that you could reference a variable from the host. Checking both key and value with a whitelist is what I choose.
You specify this yourself in the comment in the changed files ...? Or am I missing something?

ArneTR:
Just quickly looked at the code and not tested yet. But what happens if env_var does not contain a "=" and ist not a key value pair, but just a string? will env_key then ever be set?

runner.py
ribalba: Currently every value needs to start with a char or _. But let's say you have a password that starts with number or has special characters. For example you could have an env var:
```
MYSQL_PASSWD=123$%^FOO
```
which will currently fail because we only allow `r'^[a-zA-Z_]+[a-zA-Z0-9_-]*$'` I now changed this to also allow 0-9 as first char as my std password for dev  work is `123`. I am not sure how to handle the value part from the key. Because if we allow anything people could do `$(SOMETHING)` or similar but if we are too restrictive we might run into issues with certain parameters. For now I would also allow 0-9 for the first char and we can see if someone complains. I would rather have it a little more secure per default.

ribalba:
It will fail. I didn't specifically catch this but have now made it more explicit.

ArneTR:
That is nice. Acutally I was under the impression that this is a feature in Docker, that you can set an empty env and just the presence of it provides some info to the app.

The actual behaviour is very nasty though:

> You can choose not to set a value and pass the environment variables from your shell straight through to a service’s containers. It works in the same way as docker run -e VARIABLE ...:
> The value of the DEBUG variable in the container is taken from the value for the same variable in the shell in which Compose is run.

From https://docs.docker.com/compose/environment-variables/set-environment-variables/

You probably have read that, for the this was news. Good that we block it. I would even encourage to make the error more specific and say "We do not allow the feature of forwarding variables from the host OS".

A thought that came to my head is allowing this feature with "--allow-unsafe", but since this is a dev only feature anyway and just leads to usage scenarios that will never work in production the user can as well just change the local usage_scenario and we save on code complexity

ArneTR:
I merged it in and changed the wording to "We do not allow the feature of forwarding variables from the host OS"

runner.py
```diff
@@ -398,29 +398,45 @@ def setup_services(self):
...
+                    #   - image: "postgres: ${POSTGRES_VERSION}"
+                    # will fail as this could expose env vars from the host system.
+                    if isinstance(docker_env_var, str) and '=' in docker_env_var:
+                        env_key, env_value = docker_env_var.split('=')
+                    elif isinstance(service['environment'], dict):
+                        env_key, env_value = str(docker_env_var), str(service['environment'][docker_env_var])
+                    else:
+                        raise RuntimeError(f"Environment variable needs to be a string with = or dict!")
+
+                    True, True
```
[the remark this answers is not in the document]
ribalba: Yes, that was me. Sorry.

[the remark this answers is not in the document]
ArneTR: I removed it already. Also ignore my question  in my email then

PR #253, opened 2023-04-24
Configuration check

djesic:


ribalba:
@djesic could you please also add 2 simple unit tests. One pass one fail. Otherwise ready to merge.

djesic:
@ribalba adding tests for this opened more questions than expected. Please see the `TODO` I've left in the test file. It can be done with copying config files to repo root during tests and cleaning them up, but I'm not sure if there was a design decision for allowing a config file to come only from the repo root?

ArneTR:
I changed the base to dev. Please recheck if all ok

PR #254, opened 2023-04-25
Added description field to usage scenarios

djesic:
As mentioned in the changelog, I've added the description field to GMT as well.
I have a whole bunch of failing tests at the moment on the `blue-angel-compare` branch, but no longer any `Key error: 'description'` at least after this commit.

@dan-mm I don't know the details of `test/data/usage_scenarios/`, so I've added where it made sense to me. Could you please double check them?

### DIFF
--- a/test/data/usage_scenarios/basic_stress.yml
+++ b/test/data/usage_scenarios/basic_stress.yml
@@ -2,6 +2,7 @@
 name: Test Stress
 author: Dan Mateas
 version: 1
+description: test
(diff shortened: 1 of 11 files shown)


dan-mm:
Looks good to me - I have a local branch for fixing tests that made the same change amongst others, but I'll update my description name to match yours to avoid conflicts. I think this can be merged in perfectly safely into the blue angel branch.

PR #256, opened 2023-04-25
184 schema checker

dan-mm:


lib/schema_checker.py
```diff
@@ -0,0 +1,118 @@
...
+        raise SchemaError(f"{filename} does not exist")
+    return value
+
+def valid_service_types(value):
+    if value != 'container':
+        raise SchemaError(f"{value} is not 'container'")
+    return value
+
+usage_scenario_schema = Schema({
+    Optional("name"): str,
```
ribalba, line 65: I think we need description here too. Are name etc optional? Or is this just the way Schema does this?

ribalba:
@dan-mm could you please also write some unit tests for the checker ;)

lib/schema_checker.py
```diff
@@ -0,0 +1,118 @@
...
+    if not value.startswith('!include '):
+        raise SchemaError(f"{value} does not begin with !include")
+    if (not value.endswith('.yml') and not value.endswith('.yaml')):
+        raise SchemaError(f"{value} does not end with a valid yaml filename")
+
+    # Extract the filename from the value
+    filename = value[len('!include '):]
+
+    # Make sure that the compose yaml exists
+    if not os.path.exists(filename):
```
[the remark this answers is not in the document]
dan-mm: At the moment its optional - if the compose-file line is missing, the validation will still pass (as its not necessary anyways)

Instead of having a flag, I think we can have the validate_compose_include function which validates this line handle the different cases. I think I need a bit more details about what would be different exactly to understand how to best handle this case

lib/schema_checker.py
dan-mm: Added a comment - but yes, everything that is Optional() is not needed - took this from our docs requirements. Lemme know if any of these are actually required

dan-mm:
> @dan-mm could you please also write some unit tests for the checker ;)

Good call - added a couple of quick tests just to see if this code basically passes/fails.

lib/schema_checker.py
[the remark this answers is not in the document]
ribalba: Here is the ticket: https://github.com/green-coding-berlin/green-metrics-tool/issues/270
We will still have the case that the `!include` line will be present but the file will not be found as we don't check out the whole repo. Only the one file.

lib/schema_checker.py
ribalba: I think they are actually required. Or they are in the code. [1] Could you please update the documentation accordingly. 

[1] https://github.com/green-coding-berlin/green-metrics-tool/blob/main/runner.py#L299

dan-mm: Made these values mandatory now

lib/schema_checker.py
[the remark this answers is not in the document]
dan-mm: Updated this - turns out there's no easy way to just pass in a flag through the validator down the chain, so instead I turned the schema checker into a class in order to store the flag as a class variable. That does the trick. You can now set a flag for whether it will actually check if the .yaml file is there or if the line is just syntactically correct

ArneTR:
@dan-mm Checks have failed. Can you please check

dan-mm:
Tests fixed - added schema to the requirements file for our docker containers. How that was handled was changed on dev at some point

PR #264, opened 2023-05-03
Extend the error helper to log to file

saiteja13427:
1. Adds an error file option in the config
2. If the error file is not None in config, then writes the errors to the error file as well

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -27,7 +27,8 @@ cluster:
 machine:
   id: 1
   description: "Development machine for testing"
-  error_log_file: None
+  # Takes a file path to log all the errors to it. This is disabled if False
+  error_log_file: False
 
 measurement:
   idle-time-start: 10
--- a/config.yml.example
+++ b/config.yml.example
@@ -27,9 +27,7 @@ cluster:
 machine:
   id: 1
   description: "Development machine for testing"
-
-error:
-  file: None
+  error_log_file: None
 
 measurement:
   idle-time-start: 10
--- a/config.yml.example
+++ b/config.yml.example
@@ -35,6 +35,9 @@ measurement:
   phase-transition-time: 1
   metric-providers:
 
+error:
+  file: None
+
   # Please select the needed providers according to the working ones on your system
   # More info https://docs.green-coding.berlin/docs/measuring/metric-providers
   # Please activate and deactivate any provider in this list by uncommenting it.
--- a/config.yml.example
+++ b/config.yml.example
@@ -28,15 +28,15 @@ machine:
   id: 1
   description: "Development machine for testing"
 
+error:
+  file: None
+
 measurement:
   idle-time-start: 10
   idle-time-end: 5
   flow-process-runtime: 1800
   phase-transition-time: 1
   metric-providers:
-
-error:
-  file: None
 
   # Please select the needed providers according to the working ones on your system
   # More info https://docs.green-coding.berlin/docs/measuring/metric-providers
--- a/lib/error_helpers.py
+++ b/lib/error_helpers.py
@@ -1,4 +1,5 @@
 import sys
+import os
 import traceback
 from terminal_colors import TerminalColors
 from global_config import GlobalConfig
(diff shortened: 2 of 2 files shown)


saiteja13427:
Even after the running install script it doesn't seem to be working

Here is the complete log
```
[2023-05-04 12:49:06 +0000] [7] [INFO] Worker exiting (pid: 7)
[2023-05-04 12:49:06 +0000] [8] [ERROR] Exception in worker process
  File "/usr/lib/python3/dist-packages/gunicorn/arbiter.py", line 589, in spawn_worker
    worker.init_process()
  File "/usr/local/lib/python3.10/dist-packages/uvicorn/workers.py", line 66, in init_process
    super(UvicornWorker, self).init_process()
  File "/usr/lib/python3/dist-packages/gunicorn/workers/base.py", line 134, in init_process
    self.load_wsgi()
  File "/usr/lib/python3/dist-packages/gunicorn/workers/base.py", line 146, in load_wsgi
    self.wsgi = self.app.wsgi()
  File "/usr/lib/python3/dist-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/usr/lib/python3/dist-packages/gunicorn/app/wsgiapp.py", line 58, in load
    return self.load_wsgiapp()
  File "/usr/lib/python3/dist-packages/gunicorn/app/wsgiapp.py", line 48, in load_wsgiapp
    return util.import_app(self.app_uri)
  File "/usr/lib/python3/dist-packages/gunicorn/util.py", line 384, in import_app
    mod = importlib.import_module(module)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/var/www/green-metrics-tool/api/api.py", line 78, in <module>
    GlobalConfig().config['cluster']['metrics_url'],
  File "/var/www/green-metrics-tool/api/../lib/global_config.py", line 16, in __init__
    self.config = yaml.load(config_file, yaml.FullLoader)
  File "/usr/local/lib/python3.10/dist-packages/yaml/__init__.py", line 81, in load
    return loader.get_single_data()
  File "/usr/local/lib/python3.10/dist-packages/yaml/constructor.py", line 49, in get_single_data
    node = self.get_single_node()
  File "/usr/local/lib/python3.10/dist-packages/yaml/composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "/usr/local/lib/python3.10/dist-packages/yaml/composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
  File "/usr/local/lib/python3.10/dist-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/usr/local/lib/python3.10/dist-packages/yaml/composer.py", line 133, in compose_mapping_node
    item_value = self.compose_node(node, item_key)
  File "/usr/local/lib/python3.10/dist-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/usr/local/lib/python3.10/dist-packages/yaml/composer.py", line 127, in compose_mapping_node
    while not self.check_event(MappingEndEvent):
  File "/usr/local/lib/python3.10/dist-packages/yaml/parser.py", line 98, in check_event
    self.current_event = self.state()
  File "/usr/local/lib/python3.10/dist-packages/yaml/parser.py", line 438, in parse_block_mapping_key
    raise ParserError("while parsing a block mapping", self.marks[-1],
yaml.parser.ParserError: while parsing a block mapping
  in "/var/www/green-metrics-tool/lib/../config.yml", line 39, column 3
expected <block end>, but found '<block mapping start>'
  in "/var/www/green-metrics-tool/lib/../config.yml", line 48, column 5
[2023-05-04 12:49:06 +0000] [8] [INFO] Worker exiting (pid: 8)
[2023-05-04 12:49:06 +0000] [1] [WARNING] Worker with pid 8 was terminated due to signal 15
[2023-05-04 12:49:06 +0000] [1] [INFO] Shutting down: Master
[2023-05-04 12:49:06 +0000] [1] [INFO] Reason: Worker failed to boot.
```

saiteja13427:
Seems like somehow if I place my change in config.yml above `measurement:` it works fine. Let me change this.

But after that when I import the demo data and try to load, one of the API calls fail.
`v1/measurements/single/fee753bb-b881-4874-a221-d0f56e864a72`

config.yml.example
```diff
@@ -35,6 +35,9 @@ measurement:
   phase-transition-time: 1
   metric-providers:
 
+error:
```
ArneTR: When the metric-providers block starts you need to at least fill in one value with the correct indentation. You cannot start from no indentation.

Moving the error block up was the right decision

saiteja13427: Yeah exactly, realised that!

ArneTR:
Hey @saiteja13427 
The YAML in your first commit is not valid. I made a comment here: https://github.com/green-coding-berlin/green-metrics-tool/pull/264/commits/af4ccb15b51900f7e7f1ef92377e84e83e118595#r1185085934

2. The demo data is not current anymore I see.

I just pushed a fix so that the demo data is accurate.

@dan-mm Can you please put an issue into your backlog that there needs to be a test that also checks if the demo data can be correctly imported into the current schema

saiteja13427:
@ArneTR Yes the yaml isn't valid in the first commit. I hope with the second commit it is fine now, right?

ArneTR:
I will have a look at it by the end of this week!

ArneTR:
Hey @saiteja13427 
I now had chance to look at it.

The implementation is quite nice!

In terms of clarity I would suggest to change the name of the key to `error-log-file` and place it under the `machine` key

example:
```
machine:
  id: 1
  description: "Development machine for testing"
  error-log-file: /tmp/arne.txt
````

What do you think?

2. Also it would be nice if the function fails nicely if the file is not writeable or cannot be created in the location

saiteja13427:
@ArneTR Those points makes sense, let me make those changes and get back!

saiteja13427:
@ArneTR I have made the suggested changes.

ArneTR:
Hey @saiteja13427 , here my remarks:

- Thanks for chaning the naming
- The variable should be either `false`or have a value I would suggest. This make it also easier to check against in Python
- Thank you for being so careful with the guard clauses. I had something way easier in mind though. I would have made no guard clause but rather just caught the exceptions for FileNotfound and access error and then made a nice error message like "Could not activate logging to file XY, because file not found or not writable". The way you have set it up now when the second and third guard clause (https://github.com/green-coding-berlin/green-metrics-tool/blob/9abdee96710c75c33f85770473a41d8ecf4f5b1b/lib/error_helpers.py#L34) fails, then the user will not even know that the logging is not working.
So my suggestion: Remove the guard clauses and just except from the three exceptions, IOError, File not found, access error

saiteja13427:
@ArneTR oh yes, that way it will be way cleaner. I should have given more thought to it, my bad!

Will make changes accordingly and push it.

In `The variable should be either false or have a value I would suggest. This make it also easier to check against in Python`, by variable are you referring to `error_log_file`

ArneTR:
Jap, I meant `error_log_file `. Using YAML has native true / false support. So I would suggest to use that. 
you can also use `null`, which has also native support, if that makes more sense to you. I personally am fine with either.

saiteja13427:
@ArneTR made the suggested changes. This time I have tested it as well and seems to work fine on my setup.

Thanks a lot for the guard suggestion, it looks really clean now.

ArneTR:
@djesic Could you give this a test drive please from your side and merge if everything is ok.

djesic:
@saiteja13427 @ArneTR 
Tested with local path, a bad path, and an absolute path. Working as expected!
Good work!

PR #282, opened 2023-05-05
Refactor phase state to use `copy_from`

djesic:
Ok, opening this up as a draft, because it currently fails to enter the data into the database.
Reason being type difference - `phase_stats.value` is `bigint` and some of the measurements are coming in as floats.
Does it make sense to change it from `bigint` to `numeric` or to round the numbers coming in?

### DIFF
--- a/lib/db.py
+++ b/lib/db.py
@@ -63,12 +63,6 @@ class DB:
         return self.__query(query, params=params, return_type='all', row_factory=row_factory)
 
     def copy_from(self, file, table, columns, sep=','):
-        from time import time_ns
-        if table == 'phase_stats':
-            print(table)
-            print(columns)
-            print(file.read())
-        start = time_ns()
         try:
             cur = self._conn.cursor()
             statement = f"COPY {table}({','.join(list(columns))}) FROM stdin (format csv, delimiter '{sep}')"
@@ -80,7 +74,6 @@ class DB:
             cur.close()
             raise exception
         cur.close()
-        print(f"Duration: {(time_ns() - start) / 1_000_000} ms\n")
 
 
 if __name__ == '__main__':
--- a/lib/db.py
+++ b/lib/db.py
@@ -64,7 +64,11 @@ class DB:
 
     def copy_from(self, file, table, columns, sep=','):
         from time import time_ns
-        print(f"Copy from start: {time_ns()}")
+        if table == 'phase_stats':
+            print(table)
+            print(columns)
+            print(file.read())
+        start = time_ns()
         try:
             cur = self._conn.cursor()
             statement = f"COPY {table}({','.join(list(columns))}) FROM stdin (format csv, delimiter '{sep}')"
@@ -76,7 +80,7 @@ class DB:
             cur.close()
             raise exception
         cur.close()
-        print(f"Copy from end: {time_ns()}")
+        print(f"Duration: {(time_ns() - start) / 1_000_000} ms\n")
 
 
 if __name__ == '__main__':
--- a/lib/db.py
+++ b/lib/db.py
@@ -63,6 +63,8 @@ class DB:
         return self.__query(query, params=params, return_type='all', row_factory=row_factory)
 
     def copy_from(self, file, table, columns, sep=','):
+        from time import time_ns
+        print(f"Copy from start: {time_ns()}")
         try:
             cur = self._conn.cursor()
             statement = f"COPY {table}({','.join(list(columns))}) FROM stdin (format csv, delimiter '{sep}')"
@@ -74,6 +76,7 @@ class DB:
             cur.close()
             raise exception
         cur.close()
+        print(f"Copy from end: {time_ns()}")
 
 
 if __name__ == '__main__':
--- a/tools/phase_stats.py
+++ b/tools/phase_stats.py
@@ -1,4 +1,5 @@
 #pylint: disable=import-error,wrong-import-position
+from io import StringIO
 import sys
 import os
 import faulthandler
@@ -11,6 +12,10 @@ sys.path.append(f"{CURRENT_DIR}/../lib")
 
 from db import DB
 
+
+
+def generate_csv_line(project_id, metric, detail_name, phase_name, value, value_type, max_value, unit):
+    return f"{project_id},{metric},{detail_name},{phase_name},{value},{value_type},{max_value or ''},{unit},NOW()\n"
 
 def build_and_store_phase_stats(project_id):
     query = """
@@ -39,19 +44,13 @@ def build_and_store_phase_stats(project_id):
 
         network_io_bytes_total = [] # reset; # we use array here and sum later, because checking for 0 alone not enough
 
-        insert_query = """
-            INSERT INTO phase_stats
-                (project_id, metric, detail_name, phase, value, type, max_value, unit, created_at)
-            VALUES
-                (%s, %s, %s, %s, %s, %s, %s, %s, NOW())
-        """
-
         select_query = """
             SELECT SUM(value), MAX(value), AVG(value), COUNT(value)
             FROM measurements
             WHERE project_id = %s AND metric = %s AND detail_name = %s AND time > %s and time < %s
         """
 
+        csv_buffer = StringIO()
 
         # now we go through all metrics in the project and aggregate them
         for (metric, unit, detail_name) in metrics: # unpack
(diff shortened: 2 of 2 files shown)


ArneTR:
1. Can you clarifiy which values exactly are coming in as float?
2. When using non-copy insertion: How do the values look before and how after?
3. I would keep the bigint type and rather round. I see in the code that some rounding should occur. I am just suprised hat psycopg does this directly ... I was under the impression it should not be doing that ...
4. Please adjust the timing code to subtract the first timestamp from the latter. Otherwise it is hard to read how long it really took

djesic:
@ArneTR 
I've added some debug logging to figure this out, here is what is coming in to the `DB().copy_from()`:

```
phase_stats
('project_id', 'metric', 'detail_name', 'phase', 'value', 'type', 'max_value', 'unit', 'created_at')
77bd78ae-ce07-4e10-a933-697bf7b66232,cpu_frequency_sysfs_core,0,000_[BASELINE],3175476.800000000000,MEAN,3401028,Hz,NOW()
77bd78ae-ce07-4e10-a933-697bf7b66232,cpu_frequency_sysfs_core,1,000_[BASELINE],3131214.288888888889,MEAN,3399530,Hz,NOW()
77bd78ae-ce07-4e10-a933-697bf7b66232,cpu_frequency_sysfs_core,2,000_[BASELINE],3100322.044444444444,MEAN,3400727,Hz,NOW()
77bd78ae-ce07-4e10-a933-697bf7b66232,cpu_frequency_sysfs_core,3,000_[BASELINE],3172632.644444444444,MEAN,3364257,Hz,NOW()
77bd78ae-ce07-4e10-a933-697bf7b66232,cpu_utilization_procfs_system,[system],000_[BASELINE],402.8400000000000000,MEAN,1500,Ratio,NOW()
77bd78ae-ce07-4e10-a933-697bf7b66232,phase_time_syscall_system,[SYSTEM],000_[BASELINE],5005890,TOTAL,,us,NOW()

Duration: 1.340768 ms
```
This is for one of the phases, I've just run stress

Also, the exception does not get printed if there are these debug print statements, which is odd.
But this was the exception before I added the debug prints:

```
Error:  Base exception occured in runner.py:  invalid input syntax for type bigint: "3119742.511111111111"
CONTEXT:  COPY phase_stats, line 1, column value: "3119742.511111111111" 233aec2e-a415-45dd-92d4-377d57ca49ac
```

djesic:
The query and params from before my commits reveled something:

```
(UUID('fff6581d-dd47-4e3a-b114-7d36cae2d6d3'), 'cpu_frequency_sysfs_core', '0', '000_[BASELINE]', Decimal('3184809.044444444444'), 'MEAN', 3400070, 'Hz')

            INSERT INTO phase_stats
                (project_id, metric, detail_name, phase, value, type, max_value, unit, created_at)
            VALUES
                (%s, %s, %s, %s, %s, %s, %s, %s, NOW())
```

So the values did already get cast into `bigint`: `Decimal('3156120.555555555556')` and in the DB, this value was bigint and rounded `3156121`.

ArneTR:
Ok, understand.

I do not really understand why th rounding is done automatically when you run line-by-line statements and copy_from complains, but maybe this is why it is faster :)

That the values are float is expected through the average and a rounding is in order. Thanks for spotting.

ArneTR:
@ribalba will review the PR in detail. My response was just for this particular issue

ribalba:
Somehow the tests are still failing. It looks like that is the rounding issue. 

Otherwise good to go from my side

ribalba (changes requested):
test fix

djesic:
@ribalba all tests are passing on my machine, can you share which ones are failing on your end?

ribalba:
Here in GitHub. Right under this comment. I reran the tests just now

djesic:
@ribalba I'm not figuring out why the tests are failing on GitHub. I see from the log that it's trying to enter floats into the db, but what changed in https://github.com/green-coding-berlin/green-metrics-tool/pull/282/commits/a2b1e2018c64cbf0d63bc1786707706949102de6
The same tests are passing locally, I'll ask Dan if he has some insight

djesic:
@dan-mm @ribalba 
The case for `max_value` being None is fixed, tests are passing in the workflow and locally

ArneTR:
Seeing that the issue seems now resolved. Could you, @dan-mm @djesic specificy how it was ever possible that the tests worked locally at some point, before the fix?

See @djesic comment here: https://github.com/green-coding-berlin/green-metrics-tool/pull/282#issuecomment-1545530196

dan-mm:
> Seeing that the issue seems now resolved. Could you, @dan-mm @djesic specificy how it was ever possible that the tests worked locally at some point, before the fix?
> 
> See @djesic comment here: [#282 (comment)](https://github.com/green-coding-berlin/green-metrics-tool/pull/282#issuecomment-1545530196)

I don't have the full picture yet, but what I can tell is that the value for max_value here is quite different between local and what I see in the VM.  Setting some print statements locally on the vm for both the value of max_value, as well as some variables before I expect it to be called, I see:

Local:
metric: cpu_frequency_sysfs_core
phase: {'start': 1685104437523642, 'name': '[BASELINE]', 'end': 1685104442526285}
result: (Decimal('75610588'), 3944908, Decimal('2160302.514285714286'), 35)
max_value NO ROUND: 3944908

metric: cpu_frequency_sysfs_core
phase: {'start': 1685104437523642, 'name': '[BASELINE]', 'end': 1685104442526285}
result: (Decimal('75994577'), 3939018, Decimal('2171273.628571428571'), 35)
max_value NO ROUND: 3939018

metric: cpu_frequency_sysfs_core
phase: {'start': 1685104437523642, 'name': '[BASELINE]', 'end': 1685104442526285}
result: (Decimal('85508548'), 4166151, Decimal('2443101.371428571429'), 35)
max_value NO ROUND: 4166151

metric: cpu_frequency_sysfs_core
phase: {'start': 1685104437523642, 'name': '[BASELINE]', 'end': 1685104442526285}
result: (Decimal('97225221'), 3952029, Decimal('2777863.457142857143'), 35)

max_value NO ROUND: 3952029
metric: cpu_frequency_sysfs_core
phase: {'start': 1685104437523642, 'name': '[BASELINE]', 'end': 1685104442526285}
result: (Decimal('100147416'), 4118624, Decimal('2861354.742857142857'), 35)
max_value NO ROUND: 4118624

[ this goes on for quite a few more providers, but it follows the same pattern ... ]

VM:
metric: cpu_utilization_procfs_system
phase: {'start': 1685104479030916, 'name': '[BASELINE]', 'end': 1685104484036111}
result: (Decimal('3478'), 1500, Decimal('69.5600000000000000'), 50)
max_value NO ROUND: 1500

metric: psu_energy_ac_sdia_machine
phase: {'start': 1685104479030916, 'name': '[BASELINE]', 'end': 1685104484036111}
result: (None, None, None, 0)

metric: psu_energy_ac_xgboost_machine
phase: {'start': 1685104479030916, 'name': '[BASELINE]', 'end': 1685104484036111}
result: (Decimal('136034'), 3613, Decimal('2720.6800000000000000'), 50)
max_value NO ROUND: None

max_value NO ROUND: 36092.49989261158

max_value NO ROUND: None
DMM: SANITY CHECK
max_value NO ROUND: None

What's interesting to me is that when the max_value is 36092, the other print statements did not get hit... which means its getting to that generate_csv_line function through a different codepath than what's happening on local. (For reference, my debug print statements are in the main loop of build_and_store_phase_stats, right around where it makes this db call:

```
        select_query = """
            SELECT SUM(value), MAX(value), AVG(value), COUNT(value)
            FROM measurements
            WHERE project_id = %s AND metric = %s AND detail_name = %s AND time > %s and time < %s
        """
        csv_buffer = StringIO()

        # now we go through all metrics in the project and aggregate them
        for (metric, unit, detail_name) in metrics: # unpack
            # -- saved for future if I need lag time query
            #    WITH times as (
            #        SELECT id, value, time, (time - LAG(time) OVER (ORDER BY detail_name ASC, time ASC)) AS diff, unit
            #        FROM measurements
            #        WHERE project_id = %s AND metric = %s
            #        ORDER BY detail_name ASC, time ASC
            #    ) -- Backlog: if we need derivatives / integrations in the future

            print(f"metric: {metric}")
            print(f"phase: {phase}")
            results = DB().fetch_one(select_query,
                (project_id, metric, detail_name, phase['start'], phase['end'], ))
            print(f"result: {results}")

            value_sum = 0
            value_max = 0
            value_avg = 0
            value_count = 0

            value_sum, value_max, value_avg, value_count = results
```
That's all the info I have so far, but I'm still digging. @djesic do you have any insights as to how it could be getting this value differently?

djesic:
@dan-mm @ArneTR 
I just re-ran the `test_jobs.py` tests locally on the latest commit (passing), and on this commit `a2b1e2018c64cbf0d63bc1786707706949102de6` which is passing locally but failed previously on the VM - https://github.com/green-coding-berlin/green-metrics-tool/actions/runs/4946827255/jobs/8935362413 , and the fail on the VM is because of `max_calue`

I do manage to get failing tests locally on this commit `ea3f248` complaining about non rounded `value`.
But I still didn't manage to reproduce the failing tests on `max_value` locally.

Let's talk about this in person today

djesic:
@ArneTR @ribalba 
It was failing on the VM because of `max_value` not being rounded for the metric provider `xgboost`. (`phase_stats.py`:106)
This provider was enabled on the VM but not locally, which is why there was a difference in test output.
This was before the latest commits, but we've identified the cause!

ArneTR:
The csv_buffer is implemented that it gets reset every phase loop iteration. Why is that?

it should be outside of `for idx, phase in enumerate(phases[0]):`

What was the rationale for placing it inside the function and thus calling the copy_from method 7 times?

djesic:
@ArneTR 
> The csv_buffer is implemented that it gets reset every phase loop iteration. Why is that?
> 
> it should be outside of `for idx, phase in enumerate(phases[0]):`
> 
> What was the rationale for placing it inside the function and thus calling the copy_from method 7 times?

An oversight of indentation, pushing a fix!

ArneTR:
looks good!

PR #295, opened 2023-05-14
Adds number of runs selected for comparison

### LINKED ISSUE #274: Display number of runs selected for comparison
Before clicking on the `Compare` button on `/index.html`, display the number of runs that are going to be compared.

Inside of `Project Data` on `/compare.html` display the number of compared runs

saiteja13427:
Adds number of runs selected for comparison. Fixes #274 

1. If there are no selected runs, the button will just say `Compare.`
image

2. If there is 1 run selected, the button says `Compare: 1 Run`
image

3. Else it says `Compare: n Runs`
image

### DIFF
--- a/frontend/js/index.js
+++ b/frontend/js/index.js
@@ -9,10 +9,8 @@ const compareButton = () => {
 }
 const updateCompareCount = () => {
     const countButton = document.getElementById('compare-button');
-    const checkedCount = document.querySelectorAll('input#chbx-proj:checked').length;
-    countButton.textContent = checkedCount === 0 ? "Compare" 
-                              : checkedCount === 1 ?`Compare: ${checkedCount} Run` 
-                              : `Compare: ${checkedCount} Runs`;
+    const checkedCount = document.querySelectorAll('input[name=chbx-proj]:checked').length;
+    countButton.textContent = `Compare: ${checkedCount} Run(s)`;
 }
 
 function allow_group_select_checkboxes(checkbox_wrapper_id){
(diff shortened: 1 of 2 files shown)


shekhar-chauhan:
@ArneTR would you rather want 3 cases:
checkedCount == 0, 1 or >1 
or 2 cases, ==0, >0 with the button having the label as 'Run(s)'

djesic:
@saiteja13427 
I just reviewed it and the functionality is there!


> @ArneTR would you rather want 3 cases: checkedCount == 0, 1 or >1 or 2 cases, ==0, >0 with the button having the label as 'Run(s)'

Could we please do 'Run(s)'? Reasoning is that the entire table shifts to expand when the 's' is added for 2nd selection

saiteja13427:
Sure @djesic, 

There will be a shift in the beginning also when the first run is selected. Would you like to show `Compare: 0 Run(s)` at the beginning as well?

djesic:
That sounds ok to me, lets do that please

saiteja13427:
Alright then, will push the change asap!

frontend/js/index.js
```diff
@@ -89,19 +97,23 @@ const compareButton = () => {
             <td class="td-index" title="${filename}">${filename}</td>
             <td class="td-index">${machine}</td>
             <td class="td-index" style="width: 120px"><span title="${last_run}">${dateToYMD(new Date(last_run))}</span></td>
-            <td><input type="checkbox" value="${id}" name="chbx-proj"/>&nbsp;</td>`;
+            <td><input type="checkbox" value="${id}" name="chbx-proj" id="chbx-proj" />&nbsp;</td>`;
```
ribalba, line 100: Thanks for the great work. I just scanned the PR and noticed that `name` and `id` are note unique. Somehow `name` not being unique [1] must have slipped through our review process. (Probably my fault :) Probably best to solve this with a `class` and then select on `.js-somename`. 

We can either merge this now and then refactor right away (@djesic) or do it in this PR. The latter being my preferred way if you could do this?
 
[1] https://www.w3schools.com/html/html_id.asp#:~:text=The%20id%20attribute%20specifies%20a,element%20with%20the%20specific%20id.
```
The id attribute specifies a unique id for an HTML element. The value of the id attribute must be unique within the HTML document.
```

saiteja13427:
@ribalba So, do you want me to either change the id to something unique or use class instead of id?

Am I right?

frontend/js/index.js
ArneTR: Name is not unique by design. When you have mutliple checkboxes with the same name they will be understood as array. In GET as well as in POST.

ribalba: Totally right. I got confused when looking at the `id` documentation.

ribalba:
Hey, actually I think you can remove the `id` and select on name https://softauthor.com/get-elements-by-name-in-javascript/

saiteja13427:
@djesic @ribalba Fixed it! Do review.

djesic (approved):

PR #297, opened 2023-05-15
Project info for compare runs

djesic:
After a lot of trial and error with `ORJSONResponse`, I got it working.
I took the approach of calling `get_project` for the first project that was passed and then display the data from there.
The `comparison case` will cover the diff (with a nice ' vs. ' message)


ribalba (approved):
looks good

djesic:
M - > L

PR #301, opened 2023-05-19
`stats.html` detailed metrics table headings reorder

djesic:


ribalba:
Hey, I think in the stats.html in the case we are only looking at one run we don't need the StdDev Column as there will never be any data in it.

djesic:
> Hey, I think in the stats.html in the case we are only looking at one run we don't need the StdDev Column as there will never be any data in it.

@ribalba If you look at this function https://github.com/green-coding-berlin/green-metrics-tool/blob/dev/frontend/js/helpers/metric-boxes.js#L166 
there is considerable code for the case of `StdDev != null` although I have not seen single runs that don't have this value as `N/A`.
There is only one call to this function, multicomparisons are handled in `displayDiffMetricBox`.

It's easy enough to remove, I'm just concerned there might be a case where this value is not `N/A` for a single run?

ribalba:
Hey, I am not sure this is ever called in the case when we are not comparing two runs? 
https://github.com/green-coding-berlin/green-metrics-tool/blob/dev/frontend/js/helpers/phase-stats.js#L132

It looks like the table is set up here:
https://github.com/green-coding-berlin/green-metrics-tool/blob/dev/frontend/js/helpers/phase-stats.js#L9

and then filled accordingly. Call me if you need more clarification

ribalba:
@ArneTR Why is the code there?

ArneTR:
> @ArneTR Why is the code there?
> 
> Keeping open till Arne had a chance to comment.

This will be called when the "Repeated Runs" case is called. So no multi-comparison case, which means it will not use the stats.html template but the compare.html.

So asking in the `stats.js` for `determineMultiComparison()` makes no sense. But having the code in the `displaySimpleMetricBox()` is necessary.

I refactored that accordingly here: https://github.com/green-coding-berlin/green-metrics-tool/commit/3c92c5bcc4d2e851ec8af835dd04774bd48bacf0

ArneTR:
Example live: https://metrics.green-coding.berlin/compare.html?ids=4a717cff-18bc-4a91-b9e0-900fcbe47a3b,ff9aa98a-c831-4cea-b792-4a32dd274423,80223820-a0e2-4038-acb6-115e2c1e4099

ArneTR:
This means the code cannot just be removed. 

In order to get rid of the table header a new case must be introduced, so that `setupPhaseTabs()` can produce the correct header and `displaySimpleMetricBox()` can fill it accordingly.

I think this change is best done if I refactor how the value is passed along the functions.

Will take this ticket under my ownership

ArneTR:
Also reminder for me. The "detailed metrics" should be removed in `stats.html`

ArneTR:
I forced push the branch and have chosen a CSS hack to do it in the end. The css label is pretty clear, so that we should recognize in the future that these table columns are dynamically displayed on request.

@djesic Please give it a look and merge when seen

PR #303, opened 2023-05-23
Relative path for usage scenario

djesic:


ArneTR:
Does this also need a refactor to use the `sanitize` method?

What is the merge order for the pull-requests exactly, so I merge them in correctly?

djesic:
@ArneTR , that's true - maybe merge this first, and I'll rebase the `sanitize` PR

ArneTR:
Please rebase this with it using the `sanitize` function, then I will test it

ArneTR:
It works great on Linux, but has an error on macOS.

I will merge it, as it was done nice to to spec and you have no means to test.

But @ribalba can you have a look at the problem.
Following happens: The code cannot detect if the local sub-folder is correct on macos, because a local path mapping is done for the `/tmp` folder to the new alias:

````
ValueError: 1. /private/tmp/green-metrics-tool/repo/drupal/usage_scenario.yml not in /tmp/green-metrics-tool/repo - file was drupal/usage_scenario.yml
````

As you can see realpath results in mapping `/tmp` to `/private/tmp` ... I guess at some point some local symlink resolution should also happen to our comparsion path and not only to the one constructed from the user input ...?

PR #305, opened 2023-05-25
Move gunicorn Docker image dependencies to requirements.txt

djesic:


ArneTR:
Cool!

FYI: I also just checked if docker understands when the requirements.txt has changed and the container will get rebuild: It does.

PR #306, opened 2023-05-25
Refactor how form data is sanitized

djesic:


api/api.py
```diff
@@ -130,8 +129,8 @@ async def get_notes(project_id):
     escaped_data = [
         [
             note[0],
-            escape(note[1], quote=False),
-            escape(note[2], quote=False),
+            sanitize(note[1]),
+            sanitize(note[2]),
             note[3],
         ]
         for note in data
```
ArneTR, line 136: no new fixed length array. Please sanitze the whole note dynamically, so that if it ever becomes one entry longer by changing the SQL query this is automatically handled

api/api.py
```diff
@@ -170,14 +169,14 @@ async def get_projects():
...
+            sanitize(project[3]),
             project[4],
             project[5],
-            safe_escape(project[6]),
-            safe_escape(project[7]),
-            safe_escape(project[8]),
+            sanitize(project[6]),
+            sanitize(project[7]),
+            sanitize(project[8]),
         ]
```
ArneTR, line 180: no new fixed length array. Please sanitze the whole project dynamically, so that if it ever becomes one entry longer by changing the SQL query this is automatically handled

api/api.py
```diff
@@ -369,22 +368,22 @@ async def post_project_add(project: Project):
...
         return ORJSONResponse({'success': False, 'err': 'E-mail is empty'})
 
-    project.email = escape(project.email, quote=False)
+    project.email = sanitize(project.email)
 
     if project.branch.strip() == '':
         project.branch = None
     else:
-        project.branch = escape(project.branch, quote=False)
+        project.branch = sanitize(project.branch)
```
ArneTR, line 386: Please no call for every single variable. Just make one loop at the end that goes over all the accessible attributes of the object or even better: Extend your sanitze function to accept objects

test/api/test_api_helpers.py
```diff
@@ -8,7 +8,7 @@
 import api_helpers
 
 
-def test_escape_dict():
+def test_sanitize_dict():
     messy_dict = {"link": '<a href="http://www.github.com">Click me</a>'}
-    escaped = api_helpers.escape_dict(messy_dict.copy())
-    assert messy_dict['link'] != escaped['link']
+    sanitized = api_helpers.sanitize(messy_dict.copy())
+    assert messy_dict['link'] != sanitized['link']
```
ArneTR, line 14: here please testing for the actual expected output.

please include a test for dicts

please include a test for objects

ArneTR (changes requested):

djesic:
@ArneTR 
> 2. The refactor does not include `quote=False` anymore. Why was it ever introduced and why is it not needed anymore?

It's an argument for `html.escape` https://docs.python.org/3/library/html.html#html.escape
It is `True` by default, but it was set to `False` explicitly in previous commits.
I don't think `'` and `"` should be escaped, it was just very messy with the `quote=False` arguments everywhere.
Now there's only one call to `html.escape` in `sanitize` that will set `quote=False` and that's it.

djesic:
@ArneTR made the requested changes, but I encountered an issue with sanitizing objects. 
Wrote tests that cover sanitizing objects and it handles `pydantic.BaseObject` as expected.

However, I had an issue when I deployed this locally and entered `<>` into the form on `/request.html`.
It's not escaped at all and I failed to debug why - temporary print messages didn't show up in container logs and can't trigger a breakpoint in a running container.
Rebuilt the image, not sure what is the problem.

PR #311, opened 2023-06-01
Refactor API to use proper HTTP status codes

djesic:


api/api.py
```diff
@@ -249,14 +246,14 @@ async def compare_in_repo(ids: str):
...
                 phase_stats_object['common_info']['Branch'] = branch
-                # TODO: diff/same commit?
             case 'Branch':
                 # same repo, same usage scenarios, same machines, diff branch
                 phase_stats_object['common_info']['Repository'] = uri
                 phase_stats_object['common_info']['Usage Scenario'] = usage_scenario
                 phase_stats_object['common_info']['Machine'] = machine
 
     except RuntimeError as err:
+        # FIXME: 204? 500?
```
djesic, line 256: @ArneTR Some of the comparison helper functions raise `RuntimeError` in cases where there is no data or invalid comparison case. I can't distinguish if it should be a HTTP 204 or maybe a 400 or even 500?

api/api.py
```diff
@@ -337,6 +334,7 @@ async def get_badge_single(project_id: str, metric: str = 'ml-estimated'):
     elif metric == 'AC':
         value = 'psu_energy_ac_%'
     else:
+        # FIXME: 400 response?
```
djesic, line 337: @ArneTR No other API functions raise exceptions, should this be refactored?

api/api.py
```diff
@@ -337,6 +334,7 @@ async def get_badge_single(project_id: str, metric: str = 'ml-estimated'):
...
         raise RuntimeError('Unknown metric submitted')
```
djesic, line 338: @ArneTR No other API functions raise exceptions, should this be refactored?

PR #314, opened 2023-06-03
Stdout Logs display in Dashboard

ArneTR:
Hey @dan-mm 

In this feature branch I am introducing logging of the stdout to database for debugging.

I had to change the order of a function: `self.read_and_cleanup_processes()`

Will that conflict with the tests?

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -129,7 +129,6 @@ async def get_notes(project_id):
     escaped_data = [sanitize(note) for note in data]
     return ORJSONResponse({'success': True, 'data': escaped_data})
 
-
 # return a list of all possible registered machines
 @app.get('/v1/machines/')
 async def get_machines():
@@ -397,7 +396,7 @@ async def get_project(project_id: str):
                     LEFT JOIN categories as t on t.id = elements) as categories,
                 start_measurement, end_measurement,
                 measurement_config, machine_specs, machine_id, usage_scenario,
-                last_run, created_at, invalid_project, phases
+                last_run, created_at, invalid_project, phases, logs
             FROM projects
             WHERE id = %s
             """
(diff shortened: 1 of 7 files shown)


ArneTR:
UPdate: I also introduced another function `check_stderr_processes`

dan-mm:
It will a little - in TestFunctions.py, function  run_until: (line 75) -> this needs to match the runner.py flow.

I see that now  there are some nested try-finally's in the flow -> originally it was just runner.cleanup in the finally block, so the tests were instrumented in the same way (see `test_usage_scenario.py::get_env_vars(runner)` for example. I would refactor it a little bit so that instead of the tests calling 

 test_usage_scenario.py::get_env_vars(runner) (line 63)
```
    finally:
        runner.cleanup()
```
I would call:
```
    finally:
        Tests.cleanup(runner)
```
and in test functions add a cleanup function that is the inner block for the runner.py flow:
test_functions:
```
def cleanup(runner):
    try:
        runner.read_stdout_logs()
        runner.read_and_cleanup_processes()
        print(TerminalColors.HEADER, '\nSaving notes: ', TerminalColors.ENDC, self.__notes)
        save_notes(self._project_id, self.__notes)
        runner..save_stdout_logs()
   finally:
        runner.cleanup()  # always run cleanup automatically after each run
```
If you want, I can check out this branch and push these changes myself

ArneTR:
@dan-mm The tests apparently still work without the proposed fixes of yours.

Does your comment mean that the tests only need to be adapted to have more coverage, or they are needed because it would break otherwise?

dan-mm:
Yeah, the tests might still pass, they just won't match the reality of what runner.py does anymore. I can't really predict if it would break or not (though it appears it won't), its more that the tests are wrong now.

Specifically any tests that are instrumented via the run_until function traverse any codepaths that you added in this (so check_stderr_processes), and despite the re-ordering of read_and_cleanup_processes, those tests will be run using the old order. It seems that this doesn't cause an error (though it could have), however it makes the tests a bit stale

PR #319, opened 2023-06-07
First draft of the runtime-sub-phases

ArneTR:
I looked a bit weird to have the extra phases also in the steps navigation bar. Therefore I decided to introduce another level inside of the runtime tab.

Please send me your feedback on how you think it looks and how self-explaining it is.

Screenshot 2023-06-07 at 3 35 04 PM

Steps are generated by having multiple flows. Here the example I used:

```` 
name: Linting tools comparison - Pylint and Ruff
author: Arne Tarara <arne@green-coding.berlin>
version: 1
description: Lint the dev branch of green-metrics-tool using pylint and ruff


services:
  linting:
    image: gcb_lint
    build:
      context: .
      dockerfile: Dockerfile


flow:
  - name: Pylint linting
    container: linting
    commands:
      - type: console
        command: /tmp/linting/venv/bin/python3 -m pylint --rcfile /tmp/src/pylint.toml /tmp/linting/*.py
        note: Linting with Pylint
        ignore-errors: true  # We expect some linting errors to occur

  - name: getting the date
    container: linting
    commands:
      - type: console
        command: date


  - name: Ruff linting
    container: linting
    commands:
      - type: console
        command: /tmp/linting/venv/bin/python3 -m ruff --config /tmp/src/pyproject.toml /tmp/linting/
        note: Linting with Ruff
        ignore-errors: true  # We expect some linting errors to occur


````

The branch should work without any re-measurements on your stuff. It is just changed stuff in the frontend. If you do not have currently data with more than one flow you have to create one :)

### CODE REVIEW

**Comment by ribalba** [comment]
General feedback:
- I don't like the '->' style. I would just leave them away.
- I wouldn't specifically state "May contain sub-phases (see below)"
- Not for this task but would be cool if you could also zoom the Metrics Charts according to the phases and sub-phases.

File: runner.py [hunk 1]
```diff
@@ -915,6 +937,27 @@ def store_phases(self):
...
+                check=True,
+                encoding='UTF-8',
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
+            if log.stdout:
+                self.add_to_log(container_name, log.stdout)
+
+    def save_stdout_logs(self):
+        logs_as_str = '\n\n'.join([f"{k}:{v}" for k,v in self.__stdout_logs.items()])
```
ribalba, line 954: Shouldn't you escape the data here. Otherwise you could log something like `<script> ...`  and this will just be outputted?

ribalba:
I found this PR really hard to read as log stuff is mixed with phase stuff. (Just an FYI)

ArneTR:
Let's discussing the styling tomorrow. 

Why the logging is mixed with the runtime-subphases I don't really understand here on github. I based the runtime-subphases off of the stdout-log branch ... maybe the view somehow got mixed up ...

I def. developed this separately and branched off

PR #324, opened 2023-06-12
Escape single and double quotes

djesic:


ribalba:
The only thing I noticed when looking at the code is that when looking at the Usage Scenario file through the font end we see 
```
author: "Arne Tarara &lt;arne@green-coding.berlin&gt;"
```
because we use the pre. Maybe it makes sense to either:
1) change the `pre` to a styled `div`
2) remove the escaped chars when outputting.

I would merge and we can discuss tomorrow.

ribalba (approved):
👍

PR #328, opened 2023-06-15
Default network creation

ArneTR:
A user recently had the issue that a "working compose file" was not working in the GMT.

Reason being that the containers were not on the same network.

At the moment we require users to explicitely put containers on the same network ... which was a decision by me in the beginning, but we confusingly do not document that anywhere :)

Since the advent of Didis code that we import compose.yml files this has become now problematic.

This PR adresses this issue by creating a default network, similar to docker compose, unlesse a container speficies explicitely otherwise.

Happy for your input.

@ribalba  Put you in as reviewer. Can you just proof read the code, no need to test run

@dan-mm : I made a branch and example application here to showcase what kind of usage_scenario should work: https://github.com/green-coding-berlin/example-applications/tree/default-network-creation/default-network-creation
Can you bring that in a unit test form that lives in the GMT and then remove the branch?

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -124,6 +124,7 @@ class Runner:
         self.__start_measurement = None
         self.__end_measurement = None
         self.__services_to_pause_phase = {}
+        self.__join_default_network = False
 
     def custom_sleep(self, sleep_time):
         if not self._dry_run: time.sleep(sleep_time)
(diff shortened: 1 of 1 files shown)


runner.py
```diff
@@ -124,6 +124,7 @@ def __init__(self,
         self.__start_measurement = None
         self.__end_measurement = None
         self.__services_to_pause_phase = {}
+        self.__join_default_network = False
```
ribalba, line 127: Why are you introducing this new variable? Isn't it enough to just check if `networks` is configured in the `usage_scenario` and if not add the default as you do below.

runner.py
```diff
@@ -682,6 +692,12 @@ def setup_services(self):
                 for network in service['networks']:
```
ribalba, line 692: You could remove the `if`? Why have 2 cases here? Does it matter if the network is specified in the usage or added by us? And you could look over `self.__networks.append(network)` and not the `services?

runner.py
ArneTR: setup_networks() is the function to determine if a network key is in the usage_scenario or not.

I could run the same check again later to determine if that is the case or not, but I opted for setting an internal state, so that later functions to not need to know the whys and wheres when a default network  is needed but can rely on a defined internal state. This because I need the information in two places

runner.py
ArneTR: it is important to only join the default network if the container has no own defined networks as one might want to separate containers.
If I look over `self.__networks.append(network) ` I would append it to any network known. This is most likely not what the user wants.

PR #333, opened 2023-06-21
Hotfix ipmi

ArneTR:


ArneTR:
@ribalba There was some security issue when a command in the sudoers could be extended at the end. Do you remember what I mean?
I do not remember to force the sudoers entry to be an "exact match"

ribalba:
If you specify a parameter it is always an exact match with only that parameter. If you don't specify any parameter in the sudoers file the users can add any except if you add `""` at the end. 

So this code looks good

ribalba (approved):

PR #340, opened 2023-06-22
Network-proxy

ribalba:
Example:

`./runner.py --uri ~/code/GreenCoding/example-applications/network-test --name testwithout  --docker-proxy`


ArneTR:
I was so eager to test it that I directly devoured it :)

Maybe it was not yet Linux tested, so please see my comments as prelimnary ... untested on macOS


1. The exact example does not work for me. The container cannot be build:
````
Building Docker images
Building gcb_alpine_curl
docker run --env HTTP_PROXY=http://host.docker.internal:8889 --env HTTPS_PROXY=http://host.docker.internal:8889 --rm -v /home/gc/example-applications/network-test:/workspace:ro -v /tmp/green-metrics-tool/docker_images:/output gcr.io/kaniko-project/executor:latest --dockerfile=/workspace/./Dockerfile --context dir:///workspace/ --destination=gcb_alpine_curl_gmt_run_tmp --tar-path=/output/gcb_alpine_curl_gmt_run_tmp.tar --no-push
Error: INFO[0000] Retrieving image manifest alpine
INFO[0000] Retrieving image alpine from registry index.docker.io

 error building image: Get "https://index.docker.io/v2/": proxyconnect tcp: dial tcp: lookup host.docker.internal on 10.0.2.3:53: no such host


````
2. I tried another example with `https://github.com/green-coding-berlin/example-applications/tree/main/wordpress-vs-hugo/hugo-apache` 

Here It get: 
````
Imported  0  metrics from  ProxyMetricsProvider
````

3. Also I guess it would be nice if the installation for `tinyproxy` is in the Install script.

**Comment by ArneTR** [comment]
Justed tested on macOS: Here the network example works great!

At the moment if you run a scenario that makes no outbound connections and you enter the Details Page you get "Could not get network intercepts data from API".
I think the 204 should not be counted as an error here.

**Comment by djesic** [comment]
@ribalba I tried running on my Ubuntu.

`--docker-proxy` unrecognized argument for runner.py?

I enabled the provider in my `config.yml` and at first was getting an stderr:
```
RuntimeError: Stderr on ProxyMetricsProvider was NOT empty: b'WARNING: obsolete config item on line 9\nWARNING: obsolete config item on line 10\n'
```
removed those two lines and no more stdout errors, but then I ran into the same docker build errors:

with uncommented `host_ip` in config:

```
docker run --env HTTP_PROXY=http://192.168.1.2:8889 --env HTTPS_PROXY=http://192.168.1.2:8889 --rm -v /home/mr-computer/Workspace/example-applications/network-test/:/workspace:ro -v /tmp/green-metrics-tool/docker_images:/output gcr.io/kaniko-project/executor:latest --dockerfile=/workspace/./Dockerfile --context dir:///workspace/ --destination=gcb_alpine_curl_gmt_run_tmp --tar-path=/output/gcb_alpine_curl_gmt_run_tmp.tar --no-push
Error: INFO[0000] Retrieving image manifest alpine             
INFO[0000] Retrieving image alpine from registry index.docker.io 
 
 error building image: Get "https://index.docker.io/v2/": proxyconnect tcp: dial tcp 192.168.1.2:8889: i/o timeout
 ```
 
 without `host_ip`:
 
 ```
 Building Docker images 
Building gcb_alpine_curl
docker run --env HTTP_PROXY=http://600:8889 --env HTTPS_PROXY=http://600:8889 --rm -v /home/mr-computer/Workspace/example-applications/network-test/:/workspace:ro -v /tmp/green-metrics-tool/docker_images:/output gcr.io/kaniko-project/executor:latest --dockerfile=/workspace/./Dockerfile --context dir:///workspace/ --destination=gcb_alpine_curl_gmt_run_tmp --tar-path=/output/gcb_alpine_curl_gmt_run_tmp.tar --no-push
Error: INFO[0000] Retrieving image manifest alpine             
INFO[0000] Retrieving image alpine from registry index.docker.io 
 
 error building image: Get "https://index.docker.io/v2/": proxyconnect tcp: dial tcp: lookup 600: no such host
 ```
 
 I also tried running the provider directly but get this:
 
 ```
  File "/home/mr-computer/Workspace/green-metrics-tool/metric_providers/network/proxy/proxy_provider.py", line 130, in <module>
    pp.read_metrics()
TypeError: ProxyMetricsProvider.read_metrics() missing 1 required positional argument: 'project_id'
```

File: api/api.py [hunk 1]
```diff
@@ -129,6 +129,25 @@ async def get_notes(project_id):
...
+            SELECT *
+            FROM network_intercepts
+            WHERE project_id = %s
+            ORDER BY time
+            """
+    data = DB().fetch_all(query, (project_id,))
+    if data is None:
+        data == []
+
+    escaped_data = [sanitize(intercept) for intercept in data]
```
djesic, line 147: `escaped_data = sanitize(data)` should be enough here
https://github.com/green-coding-berlin/green-metrics-tool/blob/main/api/api_helpers.py#L256

metric_providers/network/proxy/proxy_conf.conf
```diff
@@ -0,0 +1,10 @@
+User nobody
+Group nogroup
+
+Port 8889
+Timeout 600
+LogLevel Connect
+
+MaxClients 100
+StartServers 2
+MaxSpareServers 5
```
djesic, line 10: Lines 9 and 10 are what caused the warning on stderr on ubuntu

ribalba:
Ahh great feedback. What version of tinyproxy are you using? And could you please run
```bash
ip route list default
```

djesic:
tinyproxy 1.11.0

```
ip route list default
default via 192.168.178.1 dev wlp4s0 proto dhcp metric 600 
```

ArneTR:
I updated the documentation here https://github.com/green-coding-berlin/documentation/pull/35/commits/8685dc51a72438768158defeb9d23e57111fe0b5 with also a wish for renaming of the provider: NetworkConnectionsProxyContainerProvider

This follows our naming scheme.

2. What is the status here? Shall I also test it, or was Danilos input enough to proceed?

ribalba:
I can do the renaming. 

@djesic input was enough for the first round. I need to add a check for the tinyproxy version as the config file has changed from 1.10 to 1.11

Also I need to investigate why the `ip` command is giving a different output on Ubuntu. Or how to find the IP that is accessible from inside the containers. 

On my todo for next week.

ribalba:
Also I need to investigate why there is an error with the `PsuEnergyAcSdiaMachineProvider`

ArneTR:
@dan-mm The tests are failing with the error: "PsuEnergyAcSdiaMachineProvider.init() got an unexpected keyword argument 'CPUChips'"

It is unclear to me why that happens and on my macOS I cannot debug it well. Can you please look into this and provide a feedback to why it happens.

ArneTR:
@dan-mm Correction on this one, just tested on my linux box, this is actually a bug in Didis code

@ribalba can you please test this also on linux with the SDIA provider enabled. The provider has different config params that are not consumed by the constructor. I gues the best fix would be to refactor the SDIA and XGBOOST providers to consume these variables in the constructor and not access the config directly

ribalba:
@ArneTR please test with your ubuntu machine. 
Also @dan-mm I am introducing the first version of the checkers. Can you please add writing some test cases to your backlog

requirements.txt
```diff
@@ -2,4 +2,6 @@ PyYAML==6.0.1
 pandas==2.0.3
 psycopg[binary]==3.1.9
 pyserial==3.5
-schema==0.7.5
\ No newline at end of file
+schema==0.7.5
+psutil==5.9.5
+xgboost==1.7.6
```
ArneTR, line 7: why is this requirement here? XGBoost should be only required if you install the reporter .

The XGBoost reporter needs in any case way more dependencies to work ...?

metric_providers/psu/energy/ac/xgboost/machine/provider.py
```diff
@@ -13,7 +13,7 @@
 from metric_providers.base import BaseMetricProvider
 
 class PsuEnergyAcXgboostMachineProvider(BaseMetricProvider):
-    def __init__(self, resolution):
+    def __init__(self, resolution, **_):
```
ArneTR, line 16: Why did you choose to not implement the same change of signature with actual positional arguments as in the SDIAProvider?

metric_providers/network/proxy/proxy_provider.py
```diff
@@ -0,0 +1,146 @@
...
+import platform
+import subprocess
+from packaging.version import parse
+
+from db import DB
+from global_config import GlobalConfig
+from metric_providers.base import MetricProviderConfigurationError
+CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
+
+class NetworkConnectionsProxyContainerProvider:
```
ArneTR, line 24: The file and path should be named with the same semantic as the class name

metric_providers/network/connections/proxy/container/provider.py

metric_providers/network/proxy/proxy_provider.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def __init__(self, host_ip=None):
+        self._metric_name = 'dockerproxy'
```
ArneTR, line 27: can we rename this to network_connections_proxy_container and then also reflect this in the `config.yml`

metric_providers/network/proxy/proxy_provider.py
```diff
@@ -0,0 +1,146 @@
...
+from packaging.version import parse
+
+from db import DB
+from global_config import GlobalConfig
+from metric_providers.base import MetricProviderConfigurationError
+CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
+
+class NetworkConnectionsProxyContainerProvider:
+
+    def __init__(self, host_ip=None):
```
ArneTR, line 26: Why did you re-implement most of the init method here?

All other providers send the setting through the super.__init__ to the base.py ... I would vouch to keep that uniform

metric_providers/network/proxy/proxy_provider.py
ArneTR: Also: Why is the class not inhering from BaseMetricProvider?

metric_providers/network/proxy/proxy_provider.py
```diff
@@ -0,0 +1,146 @@
...
+    def check_system(self):
+
+        output = subprocess.check_output(["tinyproxy", "-v"], stderr=subprocess.STDOUT, text=True)
+        version_string = output.strip().split()[1].split('-')[0]
+        if parse(version_string) >= parse("1.11"):
+            return True
+
+        raise MetricProviderConfigurationError('Tinyproxy needs to be version 1.11 or greater.')
+
+    def has_started(self):
```
ArneTR, line 47: why are these functions has_started and get_stderr reimplemented with the same code from the base.py here?

I asked the question about inheritance earlier, so I guess this is related

metric_providers/network/proxy/proxy_provider.py
```diff
@@ -0,0 +1,146 @@
...
+                os.killpg(ps_group_id, signal.SIGKILL)
+
+        except ProcessLookupError:
+            print(f"Could not find process-group for {self._ps.pid}",
+                    file=sys.stderr)  # process/-group may have already closed
+
+        self._ps = None
+
+if __name__ == "__main__":
+    pp = ProxyMetricsProvider()
```
ArneTR, line 142: This needs to be updated to the correct name or removed

lib/system_checks.py
```diff
@@ -0,0 +1,112 @@
...
+def check_free_disk_80(_):
+    return check_free_disk(80)
+
+def check_free_disk_90(_):
+    return check_free_disk(90)
+
+def check_free_disk_95(_):
+    return check_free_disk(95)
+
+def check_temperature(_):
```
ArneTR, line 56: This is checking for the critical temp? Does this make sense? To my understanding the PC turns off when that is reached ...?

ArneTR (changes requested):

ArneTR:
Nice work on this branch!
 
I see you mixed a lot of open issues you had and I think they all blend in well with the commit.

I merged the latest changes from my side with the `docker_prune` and left some commits on the new stuff you added.

Triggering a Workflow run now to see if the merge went well

requirements.txt
ribalba: I wanted to benchmark against the xgboost reporter and it didn't work and there was no reference in the documentation [1]. Now looking into this we do the git submodule update in the install script but then don't install the packages as described in [2]. I think we should make this consistent. Either do the submodule init and install the pip packages and give the user a working system or don't init the submodules in the install. 

[1] https://docs.green-coding.berlin/docs/measuring/metric-providers/psu-energy-xgboost-system/
[2] https://docs.green-coding.berlin/docs/installation/installation-linux/#xgboost

metric_providers/psu/energy/ac/xgboost/machine/provider.py
ribalba: Good point :) no easy wins here.

metric_providers/network/proxy/proxy_provider.py
ribalba: Legacy. Good point

lib/system_checks.py
ribalba: It checks for high temp not critical. Critical is another field.

ribalba:
@ArneTR feedback has been added. Can you please test under Ubuntu and then merge if ok.

metric_providers/psu/energy/ac/xgboost/machine/provider.py
```diff
@@ -9,18 +9,25 @@
 
 #pylint: disable=import-error, wrong-import-position
 import model.xgb as mlmodel
-from global_config import GlobalConfig
 from metric_providers.base import BaseMetricProvider
 
 class PsuEnergyAcXgboostMachineProvider(BaseMetricProvider):
-    def __init__(self, resolution, **_):
+    def __init__(self, resolution, HW_CPUFreq, CPUChips, CPUThreads, TDP,
```
ArneTR, line 15: is there the possibility to add these variables not by position, but by keyword? If we every by accident change the order in the config.yml this will bite us in the ass

metric_providers/network/connections/proxy/container/provider.py
```diff
@@ -44,11 +44,6 @@ def check_system(self):
 
         raise MetricProviderConfigurationError('Tinyproxy needs to be version 1.11 or greater.')
 
-    def has_started(self):
-        return self._has_started
-
-    def get_stderr(self):
-        return self._ps.stderr.read()
 
     def get_docker_params(self):
```
ArneTR, line 49: Can you please also refactor the class to use the base.py start_profiling method instead of overloading it.

Supplying extra switches is possible through the `self._extra_switches` variable

ArneTR (changes requested):

ArneTR:
I sadly cannot get the provider to work under linux. The network-test runs fine, but when I run a more complex one like the Nextcloud tests it just timeouts with "waiting for locator("#adminlogin")"

In the logs I find this:

```
<body>
 <h1>Unable to connect</h1>
<p>Tinyproxy was unable to connect to the remote web server.</p>
<p><em>Generated by tinyproxy version 1.11.0.</em></p>
```

This is what I ran: `./runner.py --uri https://github.com/green-coding-berlin/nextcloud-docker --dev-repeat-run --name "Nextcloud Talk (mod_php-MariaDB-Chromium)-Network-Proxy" --filename energy-tests/usage_scenario-mariadb-talk-chromium.yml --skip-unsafe`

I tried on the blockheating box and my local linux macbook. I did not test on mac, as we do not have a mac working playwright container atm. this is still waiting for this: https://github.com/green-coding-berlin/green-metrics-tool/pull/394

On the linux macbook I even tried debugging with `--debug --allow-unsafe`. The nextcloud admin login is actually visible under http://localhost:8080

However the GMT playwright cannot seem to find it from inside the containers by using tinyproxy

ArneTR:
I was today in the situation that I wanted to add some functionality to the checker. Is it maybe possible to have that seperately already included into main? 

At the moment it is in that convoluted branch here only, correct?

ribalba:
Yes currently this is only implemented here as we need the functionality for the network proxy. Let's have a quick chat what to do with this PR

ribalba:
Just for reference a nicer way of doing this probably:
https://www.appsmith.com/blog/test-proxy-support

PR #352, opened 2023-06-29
Removing version requirement

ArneTR:
Currently we have the 'version' in the `usage_scenario.yml`.

I advocate here to remove this as it is just confusing because it conflicts with the `docker-compose.yml` *version* which actually has a meaning.

The idea of having a *version* is that one can versionate the `usage_scenario.yml`. However the way we actually use it we never advance this number and since we now merge with a compose file this is more confusing than helpful and anyway overwritten by the `docker-compose.yml` in the merge process ...

@dan-mm ok? Tests should not be concerned I guess?
@ribalba remarks?

### DIFF
--- a/lib/schema_checker.py
+++ b/lib/schema_checker.py
@@ -73,7 +73,6 @@ class SchemaChecker():
         usage_scenario_schema = Schema({
             "name": str,
             "author": str,
-            "version": Use(int),
             "description": str,
 
             Optional("networks"): {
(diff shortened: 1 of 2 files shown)


ribalba:
All in favour for dropping things.

I could also argue to make `name`, `author` and `description` not required. But I get the idea of making good style required.

ArneTR:
@dan-mm can you please remove the version from all examples, the documentation and also from your tests and then merge this in.

PR #353, opened 2023-06-30
First draft for fix regarding docker context subdir fix

ArneTR:
@ribalba For a client the issue occured that the follow `compose.yml` file lead to an error:

```
---
#name: Stress Container One Core 5 Seconds
services:
  gcb-alpine-stress:
    container_name: gcb-alpine-stress
    build:
      context: ./subdir
      dockerfile: ./Dockerfile
```

This is a modified version of the [Stress example](https://github.com/green-coding-berlin/example-applications/tree/main/stress) where I moved the`Dockerfile` into a subfolder.

The code in the `join_path_and_file` function is only designed to accept files. But in order to check the context, which is triggered here: https://github.com/green-coding-berlin/green-metrics-tool/blob/4753c0e89dca679d7f08abbcb2214e30dfa48844/runner.py#L273
actually a folder is supplied.

This leads to an unwanted error. The error does appear when we use `.` as a context as this is shortcutted earlier: https://github.com/green-coding-berlin/green-metrics-tool/blob/4753c0e89dca679d7f08abbcb2214e30dfa48844/runner.py#L62

The proposed fix removed the filter for `is_file`. My question: Is that safe and also cross-platform safe? Under my linux box the iterator does not include `..` and `.` directories, but only the real contents. Unclear how symlinks might appear hear ... the `Path(path)` call should resolve them if I understand correctly?

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -66,7 +66,7 @@ def join_path_and_file(path, file):
         raise ValueError(f"{file} not in {path}")
 
     # To double check we also check if it is in the files allow list
-    folder_content = [str(item) for item in Path(path).rglob("*") if item.is_file()]
+    folder_content = [str(item) for item in Path(path).rglob("*")]
     if filename not in folder_content:
         raise ValueError(f"{file} not in {path}")
 


ribalba:
This look ok. I can't see any implications in dropping the `is_file`. 

`Path(path)` will not resolve symlinks. You need to call `.resolve()`
https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve

I do this here:
https://github.com/green-coding-berlin/green-metrics-tool/pull/353/files#diff-34cde5e49b54d9b67952c80d3580b6c421ebcf16097996e215aef54e44f78562R74

ArneTR:
I would here also argue that the is_file() is quite nice. But we should also allow for is_dir() and then let the function swap accordingly.

The distinction between the two is quite valueable, as sometimes we want to bind a file and sometimes a directory, sometimes both.

This is TBD. For the moment it works ...

ArneTR:
I have now introduced the functionality to distinguish between dir() and file(). Just waiting for the tests to pass

ArneTR:
@dan-mm The PRs now fail always. Some problem with writing issue. Any clue why?

PR #356, opened 2023-07-01
tmpfs can now be set in install script

ArneTR:
@dan-mm I added a functionality so that the system will be put on a `tmpfs` for the `/tmp` directory in the install script.

This parameter triggers an input dialog.

Could you please adapt tests and pipelines so that the new code will run and then merge it?

### DIFF
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -15,8 +15,9 @@ api_url=''
 metrics_url=''
 no_build=false
 no_hosts=false
+ask_tmpfs=true
 
-while getopts "p:a:m:n:h" o; do
+while getopts "p:a:m:n:h:t" o; do
     case "$o" in
         p)
             db_pw=${OPTARG}
@@ -33,6 +34,10 @@ while getopts "p:a:m:n:h" o; do
         h)
             no_hosts=true
             ;;
+        t)
+            ask_tmpfs=false
+            ;;
+
     esac
 done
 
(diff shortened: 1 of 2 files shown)


install_linux.sh
```diff
@@ -50,6 +55,13 @@ if [[ -z "$db_pw" ]] ; then
     read -sp "Please enter the new password to be set for the PostgreSQL DB: " db_pw
 fi
 
+if [[ $ask_tmpfs == true ]] ; then
+    read -p "We strongly recommend mounting /tmp on a tmpfs. Do you want to do that? (y/N)" tmpfs
+    if [[ "$tmpfs" == "Y" || "$tmpfs" == "y" ]] ; then
+        sudo systemctl enable /usr/share/systemd/tmp.mount
```
ribalba, line 61: We need a check there for fedora. The command under fedora is
```
sudo systemctl unmask --now tmp.mount
```

Just enabling tmp.mount does not work as it is not called by itself but wanted by local-fs.target

ribalba (changes requested):
Should we not also write a little check function to make sure that when installed tmp is tmpfs and then offer to enable it?

ArneTR:
> Should we not also write a little check function to make sure that when installed tmp is tmpfs and then offer to enable it?

I would argue this is part of the guard clauses / sanity checks.

install_linux.sh
ArneTR: cool, ty!

PR #358, opened 2023-07-02
Errors are now all logged and excepted ony by one; Stderr does not fa…

ArneTR:
I have revamped the error handling quite a bit:

1. stderr does not fail a process anymore in the flow. This was very counter intuitive for many people already as stderr does not always indicate an error.
New behaviour is now that only the process returncode is relevant and the stderr is logged in any case

2. when an exception occurs now the metrics that are avaiable are saved. This makes it nice to already see some data and wage if at least some processing / compute did occur.
To enabel this I created a quite ugly indented try-finally block. Open for suggestions


ArneTR:
@dan-mm Could you have a look at this and migrate the tests to my new structure of the runner.py. Also a test seems to be failing, but just because it expects a differnt output

runner.py
```diff
@@ -1123,22 +1136,45 @@ def run(self):
...
+                            self.add_to_log(exc.__class__.__name__, str(exc))
+                            raise exc
+                        finally:
+                            try:
+                                self.stop_metric_providers()
+                            except BaseException as exc:
+                                self.add_to_log(exc.__class__.__name__, str(exc))
+                                raise exc
+                            finally:
+                                self.cleanup()  # always run cleanup automatically after each run
```
ribalba, line 1177: Something like this might be clearer:
```
def execute_and_log_on_exception(method):
    try:
        method()
    except BaseException as exc:
        self.add_to_log(exc.__class__.__name__, str(exc))
        return exc
    return []

try:
    self.read_container_logs()
except BaseException as exc:
    self.add_to_log(exc.__class__.__name__, str(exc))
    raise exc
finally:
    raised = []
    raised += execute_and_log_on_exception(self.read_and_cleanup_processes)
    raised += execute_and_log_on_exception(self.save_notes_runner)
    raised += execute_and_log_on_exception(self.save_stdout_logs)
    raised += execute_and_log_on_exception(self.stop_metric_providers)
    self.cleanup()  # always run cleanup automatically after each run

    if raised:
        for rais in raised:
            print(f"raised: {rais}")
        raise BaseException("Some shit went down")
```

But it is more code and probably more verbose. I am not sure.

ribalba (approved):
Only minor things

dan-mm:
> @dan-mm Could you have a look at this and migrate the tests to my new structure of the runner.py. Also a test seems to be failing, but just because it expects a differnt output

On it - I updated the test runner with the new exception try-block tree, and seeing some more failing tests because they're expecting stderr where there now is none, but I'll sort through that

dan-mm:
I updated the test-runner to reflect the new error structure, and fixed the one failing test whose expected output had changed.

however I'm still seeing failures locally - I think this is an actual bug though, where an exception is being raised in base.py while trying to raise another exception. @ArneTR - I already sent you an email about this

ArneTR:
I saw the email. But the pipeline has worked now? Is that expected then?

dan-mm:
Hm, no, its unexpected. I just ran it again locally and I still get the same error (whether I run it via the tests or the runner.py directly). 

Might be a problem specific to my local system but I can't spot what it is. I double checked my config file that it matches the VM, and reinstalled my pip requirements as well.

If you run the tests locally (to save time, just a single test like test_usage_scenario.py::test_env_variable_no_skip_or_allow, does it pass for you locally?

PR #367, opened 2023-07-07
No default shell

ArneTR:
Until now we are issueing commands by default in a 'sh' process.

Not every container has a shell though. So we make it optional. And also not the default, as this is rather confusing.

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -886,13 +886,38 @@ class Runner:
                     # Since Popen always make the process asynchronous we can leverage this to emulate a detached
                     # behavior
 
-                    #pylint: disable=consider-using-with
-                    ps = subprocess.Popen(
-                        docker_exec_command,
-                        stderr=subprocess.PIPE,
-                        stdout=subprocess.PIPE,
-                        encoding='UTF-8'
-                    )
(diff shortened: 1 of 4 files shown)


runner.py
```diff
@@ -750,7 +750,11 @@ def setup_services(self):
                 continue  # setup commands are optional
             print('Running commands')
             for cmd in service['setup-commands']:
-                d_command = ['docker', 'exec', container_name, 'sh', '-c', cmd] # This must be a list!
+                if shell := service.get('shell', False):
+                    d_command = ['docker', 'exec', container_name, shell, '-c', cmd] # This must be a list!
+                else:
+                    d_command = ['docker', 'exec', container_name, *cmd.split()] # This must be a list!
```
ribalba, line 756: @ArneTR why is `cmd` handled differently here? Shouldn't we either assume a list or always split?

ArneTR: Escaping needs to happen on a one by one level if we want to send it directly to docker, otherwise the CLI does not understand the arguments. in the shell case we want the exact opposite, as we do not want to escape the quoted block

PR #368, opened 2023-07-07
Non blocking buffer reads

ArneTR:
Buffer reads could have a deadlock if there was a lot written to stdout.

This is now migitated by using `subprocess.run()` if possible. If we must use `subprocess.Popen` we use `communicate()` to read the buffer non-blocking. This only works if we do not timeout the process but use that only for detached ones.


ArneTR:
Github has a hickup. Cannot change to merge to dev ... :(

ArneTR:
@dan-mm The tests work for me locally. But they do not on the VM ... if that helps with debugging

PR #369, opened 2023-07-07
Has started rework

ArneTR:
Bugfix so that SDIA and XGBoost provider can still provide metrics although they technically never really start


ArneTR:
Merged branch manually to dev, as github is broken

PR #385, opened 2023-07-17
Sanitize CI

djesic:
Code scanning alerts do have a # but they can't be referenced if they are not issues, so here's what this fixes:

- Alerts 1-6 : Icon replacements. We don't open or visit the URL, and the matching is for the beginning of the URL. I would dismiss these alerts
- Alert 7:  wrap in `encodeURIComponent()`
- Alert 8,9,10,11: wrap in new `sanitize()` js function in `helpers/main.js` that escapes < > & ' "

### DIFF
--- a/frontend/js/ci.js
+++ b/frontend/js/ci.js
@@ -241,10 +241,10 @@ const displayCITable = (runs, url_params) => {
 
         var run_link = ''
         if(source == 'github') {
-            run_link = `https://github.com/${url_params.get('repo')}/actions/runs/${run_id}`;
+            run_link = `https://github.com/${escapeString(url_params.get('repo'))}/actions/runs/${run_id}`;
         }
         else if (source == 'gitlab') {
-            run_link = `https://gitlab.com/${url_params.get('repo')}/-/pipelines/${run_id}`
+            run_link = `https://gitlab.com/${escapeString(url_params.get('repo'))}/-/pipelines/${run_id}`
         }
 
         const run_link_node = `<a href="${run_link}" target="_blank">${run_id}</a>`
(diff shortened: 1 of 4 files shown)


djesic:
@dan-mm all tests were passing, but there was an issue:
https://github.com/green-coding-berlin/green-metrics-tool/actions/runs/5574522161/jobs/10183235301?pr=385#step:5:63

dan-mm:
> @dan-mm all tests were passing, but there was an issue: https://github.com/green-coding-berlin/green-metrics-tool/actions/runs/5574522161/jobs/10183235301?pr=385#step:5:63

Aye, its a bug I introduced in v2 of the eco-ci, looking at it now

frontend/js/index.js
```diff
@@ -5,7 +5,7 @@ const compareButton = () => {
     checkedBoxes.forEach(checkbox => {
         link = `${link}${checkbox.value},`;
     });
-    window.location = link.substr(0,link.length-1);
+    window.location = encodeURIComponent(link.substr(0, link.length - 1));
```
ArneTR, line 8: What is this supposed to do? Which part of the checkboxes is supposed to be malicious?

ArneTR (changes requested):

frontend/js/index.js
djesic: I don't honestly understand the vulnerability here, but this was the recommendation on how to resolve it.

frontend/js/ci.js
```diff
@@ -318,16 +318,16 @@ $(document).ready((e) => {
...
         if(badges_data.data[0][8] == 'github') {
-            repo_link = `https://github.com/${url_params.get('repo')}`;
+            repo_link = `https://github.com/${sanitize(url_params.get('repo'))}`;
         }
         else if(badges_data.data[0][8] == 'gitlab') {
-            repo_link = `https://gitlab.com/${url_params.get('repo')}`;
+            repo_link = `https://gitlab.com/${sanitize(url_params.get('repo'))}`;
         }
         //${repo_link}
         const repo_link_node = `<a href="${repo_link}" target="_blank">${url_params.get('repo')}</a>`
```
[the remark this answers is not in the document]
djesic: Missed. Fixed in last commit

djesic:
> 1. Are all the security vulnerabilities from https://github.com/green-coding-berlin/green-metrics-tool/security/code-scanning handled here or is this supposed to be in a different PR? It looks to me like quite some are missing?
> 2. Also I two comments in the review
> 3. I think we should rather use the word escape than sanitze, as we are not really removing anything. @dan-mm Can you confirm on the wording?

1. See my opening comment.
2. Replied
3. Renamed to `escapeString` (escape was reserved)

frontend/js/index.js
ribalba: Do you have a link?

ribalba:
Work has been ported to https://github.com/green-coding-berlin/green-metrics-tool/pull/411

PR #388, opened 2023-07-18
Fix pr workflow

dan-mm:


install_linux.sh
```diff
@@ -129,6 +129,9 @@ PYTHON_PATH=$(which python3)
 PWD=$(pwd)
 echo "ALL ALL=(ALL) NOPASSWD:$PYTHON_PATH $PWD/lib/hardware_info_root.py" | sudo tee /etc/sudoers.d/green_coding_hardware_info
 
+print_message "Installing IPMI tools"
+sudo apt-get install freeipmi-tools ipmitool
```
ribalba, line 133: @ArneTR @dan-mm This will not work under Fedora

dan-mm: This PR was a mistake - I meant to merge my branch into dev (but main was selected by default). That's why this PR is already closed.

However, this PR is effectively what dev->main would look like, so @ribalba's concern would still be relevant there.

ArneTR: > @ArneTR @dan-mm This will not work under Fedora

If this does not work under fedora: Can you then push a guard clause / if statement that makes it work?

ArneTR: > > @ArneTR @dan-mm This will not work under Fedora
> 
> If this does not work under fedora: Can you then push a guard clause / if statement that makes it work?

@ribalba

PR #389, opened 2023-07-18
eco-ci start measurement now also referencing v2

dan-mm:


ArneTR:
@dan-mm does this need a merge to main?

PR #392, opened 2023-07-20
Print name of service that had an issues with creating volumes

djesic:


ArneTR:
Looks ok to me, but tests are failing. @djesic Did you run the tests locally?

It seems this is a string matching error and since you changed the error string it is now failing ...

Can you also fix the matching in the tests?

ArneTR:
That looks quite confusing, the changes to the strings you made are unrelated to the code you have initially commited?

How did the tests ever work before? Please pull Dan in if you don't know.

djesic:
They are directly related, I've updated the part of the string that gets matched.
The tests made sense, the only exception string that is not covered by these tests is: "Volumes must be a list but is".
I don't see a reason why the tests would not work, what makes you say that?

Pinging @dan-mm just to keep him in the loop.

ArneTR:
Thanks for the clarification.

This was actually an oversight from me. I did not see that the word "trying" was actually in the string you touched in code.

Resolved.

PR #393, opened 2023-07-20
Fixes the metric output file becoming corrupted

ribalba:


metric_providers/powermetrics/provider.py
```diff
@@ -33,13 +35,35 @@ def __init__(self, resolution):
...
+            print('Killed powermetrics process with killall!')
+
+        # As killall returns right after sending the SIGKILL we need to wait and make sure that the process
+        # had time to flush everything to disk
+        count = 0
+        while self.is_powermetrics_running():
+            time.sleep(1)
+            count += 1
+            if count >= 5:
+                raise RuntimeError('powermetrics is not stopping. Please kill with `kill -9 ...`.')
```
ArneTR, line 63: If it needs to be anyway killed with kill -9, why don't we do it?

This would then not only leave the output file corrupted, but also the state of the GMT, as you might have two instances of powermetrics running. correct?

ArneTR (changes requested):

metric_providers/powermetrics/provider.py
ArneTR: Has been stale for a week now. If you do not have any further input I would vouch for changing the code to be less manual and more automatic.

ribalba: Totally. Is on my todo.

ribalba: I now remember why I did it like this. `kill -9` needs to be called as root and I was hesitant to add another line to sudoers. No real reason why to be honest. Added it.

PR #394, opened 2023-07-23
added workflow for building and pushing containers from docker/auxili…

dan-mm:
…ary-containers subfolders


dan-mm:
This will need a merge to main as well for the workflow to show up in the Actions tab

ArneTR:
1. Also here the question: Why is the script only `set -e` and not `set -euo pipefail` ?

2. I tried https://github.com/dan-mm/test-repo-a/pkgs/container/test-repo-a%2Fdan-mm%2Ftest-puppeteer and the container works fine!

What do you need, other than a merge to main, to have these available on Github? I think we should ditch docker hub entirely ...

In what interval is the workflow running?

ArneTR:
@dan-mm Can you please double check this: https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages

It speaks of a data storage and data transfer limit. Does this apply to us?

If so: Your image is already 2 GB. How can that even work then?

dan-mm:
> 1. Also here the question: Why is the script only `set -e` and not `set -euo pipefail` ?
No good reason my end, pushed it with euo pipefail

> 2. I tried https://github.com/dan-mm/test-repo-a/pkgs/container/test-repo-a%2Fdan-mm%2Ftest-puppeteer and the container works fine!
That's really good info to know! The only difference the two versions is in how its built - the one in the link above was built using the buildx github action, what's checked in is a refactor that calls buildx from a shell script. I thought it would be identical but it appears to not be the case...

I'll work on a version of this that uses the buildx github action then, but still searches through the subdirectories as I intended. Please hold on merging this until I do so.

> What do you need, other than a merge to main, to have these available on Github? I think we should ditch docker hub entirely ...
Should just be the merge to main-  but hold off on that until I make the changes described above. Our gmt repo should already be setup to push containers because we already use it for our cached containers.

> In what interval is the workflow running?
Its on manual run only at the moment, that's what made the most sense to me. I figured we'd only want to update these containers intentionally

dan-mm:
> @dan-mm Can you please double check this: https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages
> 
> It speaks of a data storage and data transfer limit. Does this apply to us?
> 
> If so: Your image is already 2 GB. How can that even work then?

hm, not sure. I'll double check this.

dan-mm:
> @dan-mm Can you please double check this: https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages
> 
> It speaks of a data storage and data transfer limit. Does this apply to us?
> 
> If so: Your image is already 2 GB. How can that even work then?

"GitHub Packages usage is free for public packages. For private packages, each account on GitHub.com receives a certain amount of free storage and data transfer, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits."

So we should be fine as long as we keep these packages public.

ArneTR:
Can you please rework this to automatically trigger when the Dockerfile has been updated (which will happen through Dependabot)

Also please let know if the token from Docker Hub Works. Containers should be on Docker Hub

dan-mm:
@ArneTR I made the requested changes, but need two things before this is done.

1- we need to set two repository secrets - DOCKERHUB_TOKEN and DOCKERHUB_USERNAME. I don't have access to do it myself, but instructions are here: https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository

2 - while this should theoretically work, I haven't been able to test it on dockerhub yet. I tried to test this on my private repo, but the token doesn't have permissions to pull/push from that repo. So in order to test this I need one of two things:
    a - login password for dockerhub so I can check/edit the token permissions
    b - this merged in to main + secrets set so I can test it on this repository (though if it doesn't work, I'll still likely need the login to debug). It likely works as is though, its pretty straightforward.

dan-mm:
I updated the tag, however both the points from the last comment are still true. We need the secrets set for dockerhub, and testing for this is not possible without either the dockerhub login or secrets set

ArneTR:
I don't quite get the comment?
- Is the container now pushed via your private workflow to Docker Hub? If so: I do not see it here: https://hub.docker.com/repository/docker/greencoding/gcb_playwright/tags?page=1&ordering=last_updated
- The repository secrets are set. I did this yesterday during our call. You cannot read them?
Screenshot 2023-09-01 at 8 52 58 AM

dan-mm:
@ArneTR that comment was written before our call, so it was out of date.

I just ran my private repo workflow again, and it passed. I see the containers on dockerhub too now: https://hub.docker.com/r/greencoding/gcb_playwright/tags

All the changes should now be in this branch, and I think we can merge this in

ArneTR:
can you please configure a workflow then in this repo that we can try which uses the repository secrets.

merging in now, so that you can try out

dan-mm:
This workflow already does that - I'll do a manual run of it

dan-mm:
Actually I didn't have to manually run it- since this was a PR that touched one of the dockerfiles and its now been merged in, it ran automatically as intended:

https://github.com/green-coding-berlin/green-metrics-tool/actions/runs/6048000694

PR #395, opened 2023-07-23
Schema checker improvements

dan-mm:
- fixes: image is now mandatory in the case that there is no build key
- network key parsing should now accept keys and list as formats
- added new log-stdout, log-stderr, shell keys


ArneTR (changes requested):
Can you please add tests for that. Especially with the network key we have run into issues before and a test is in dire need.

dan-mm:
> Can you please add tests for that. Especially with the network key we have run into issues before and a test is in dire need.

good call, tests added

ArneTR (changes requested):
These yaml files are not valid.

A key cannot be a duplicate. You cannot have the networks key twice, it will just be removed by the loader.

Please split the cases of having a list and a hash-map out into two tests.

To reproduce what I said::
```
>>> a = yaml.safe_load(open('/tmp/compose.yml'))
>>> a
{'name': 'Test', 'author': 'Dan Mateas', 'description': 'test', 'networks': ['network-a', 'network-b'], 'services': {'test-container': {'type': 'container', 'image': 'gcb_stress', 'build': '.'}}, 'flow': [{'name': 'Stress', 'container': 'test-container', 'commands': [{'type': 'console', 'command': 'stress-ng -c 1 -t 1 -q'}]}]}
```
As you see only one network key gets read.

dan-mm:
@ArneTR good catch, thank you. Updated the tests. This also uncovered a case which I thought was working but was mistaken, where multiple network names as keys would not validate (it'd work only if there was one key). This has now been fixed in my last commit

test/data/usage_scenarios/schema_checker_valid.yml
```diff
@@ -16,6 +21,9 @@ flow:
...
     commands:
-      - type: console
-        command: stress-ng -c 1 -t 1 -q
-        note: Starting Stress
\ No newline at end of file
+      - type: console 
+        command: stress-ng -c 1 -t 1 -q 
+        note: Starting Stress
+        shell: bash 
+        log-stdout: yes
```
ArneTR: Why is this valid? log-stdout is a boolean ...?

ArneTR (changes requested):

test/data/usage_scenarios/schema_checker_valid.yml
dan-mm: I've changed this to a more standard 'true'

dan-mm:
changes made, unless there's anything else needed I think this can be merged in

PR #409, opened 2023-08-03
Enables empty services in the usage_scenario

ribalba:
This enables you to have a compose file in which you specify all services that might be used at some stage. You can disabale a service in your usage_scenario by setting it to None. So let's say you have two databases defined in your compose.yml `db1` and `db2`.  You can only use one by adding to your usage file but still include the compose.
```
services:
    db2:
```

### DIFF
--- a/lib/schema_checker.py
+++ b/lib/schema_checker.py
@@ -80,7 +80,7 @@ class SchemaChecker():
             },
 
             Optional("services"): {
-                Use(self.contains_no_invalid_chars): Or (None, {
+                Use(self.contains_no_invalid_chars): {
                     Optional("type"): Use(self.valid_service_types),
                     Optional("image"): str,
                     Optional("networks"): self.single_or_list(Use(self.contains_no_invalid_chars)),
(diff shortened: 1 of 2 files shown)


lib/schema_checker.py
```diff
@@ -80,7 +80,7 @@ def check_usage_scenario(self, usage_scenario):
             },
 
             Optional("services"): {
-                Use(self.contains_no_invalid_chars): {
+                Use(self.contains_no_invalid_chars): Or (None, {
```
ArneTR, line 83: I understand that you want to remove services in the resulting usage_scenario from the initial compose, but the schema checker runs after the merge.

I would argue if after the merge the services key is empty we should still fail ...?

ArneTR (changes requested):

lib/schema_checker.py
ribalba: A very very good catch. I initially allowed None as a value and handled it all places in the runner but then decided to do the "easier" thing and just remove them.

ribalba:
@ArneTR changed and ready to squash and merge

PR #420, opened 2023-08-11
Timeline comparison

ArneTR:
This commit is a bit bigger, as I merged in old stale branches where I did some previous work. This contained also some refactorings.

- Removed Danilos Code where internally API functions with JSON output are called
- Refactored "sanitize" to "html_escape_multi"
- Added /timeline and /badge/timeline API endpoints
- Added /timeline.html view
- Small bugfixes when calling fetch_one with aggregate queries
- METRIC_MAPPINGS removed from Python code and moved to JS
- Charts now display nicer titles with info icon
Screenshot 2023-08-11 at 5 53 52 PM


ArneTR:
@ribalba Can you give the code a quick look and also check if the display is ok everywhere.

I have checked out a version only already and sent out for display. So no rush on this task

ribalba:
I gave the code a quick read and it all looks ok. We should maybe add some data to the sample data so tests are easier.

PR #428, opened 2023-08-15
added backwards compatability for cpu_util_avg stat, if its missing

dan-mm:
- if cpu_util is missing, it will display as '--%' in the front end in the stats table, all measurements table, and chart tooltip
- if there is partial cpu_util data, the average will be calculated over the number of cpu_utils that we have data for, not all runs
- tooltips with info on front-end in case people see '--%' and have questions

incidentally, in my tests I see that the math.mean does work and give you a correct value if you have only 1 measurement, which is why I changed the line from if (cpuUtilMeasurements.length > 1) -> if (cpuUtilMeasurements.length > 0) in calculateStats


ArneTR:
I see that the code mitigates the issue with the cpuUtil being null, but It still feels very hacky.

Why are you not appending elements to an array and then later on just run mean() on the array? This solves the problem that you have to keep track of count, etc. and makes it very obvious what you are doing.

Also you could do this for all the values. Although I see no reason why for instance energy and time should have different counts, but in your code this would also lead to a problem where when they get out of sync their value is mal-attributed to zero.

ArneTR:
If I counted correctly you are in total, with all filters and maps, iterating 12 times over the data structure ... I have to say this is a bit much.

Can you not rather do this in one loop and create the bins there that you need? 

Also passing the raw array to `calculateStats` and then mapping it in there feels a bit like this could be way better solved with having 5 arguments to the function and passing it the energy, time, cpu etc. bins directly which have been constructed beforehand

dan-mm:
I agree that 12 iterations is a bit much for this - I will simplify.

> Why are you not appending elements to an array and then later on just run mean() on the array? 
I am a bit confused by this, because that's pretty much exactly what I'm doing. That's what calculateStats does (albeit in too many steps). getFullRunStats & getStatsofLabel are just two filtering functions that take all your raw measurements, and filters it down to only the ones you want stats for. getFullRunStats additionally needing to aggregate some of the values together before passing them to calculateStats.

But I take it then that the best way to do this is when doing this initial filtering, to at the same time also build up the energy/time/cpu arrays and pass them into calculateStats as arguments. Ok yeah that makes sense, I'll do it that way. In my head I just went ( step 1 - filter results, step 2- calculate the stats based on the filters).

PR #430, opened 2023-08-16
Added python and system packages listing

ArneTR:
@ribalba Can you please check if this works on Fedora.

I was not sure if I should do `dnf list installed` or `rpm -qa`.

Also I hacked it into a bash query, which I think is nicer than doing it in python. Or did you plan already for a mechanism to make that if/else smoother in Python?

Demo Run with data: https://metrics.green-coding.berlin/stats.html?id=7b635c8e-5dbc-4bc2-b2c5-99d373cce558

### DIFF
--- a/lib/hardware_info.py
+++ b/lib/hardware_info.py
@@ -78,7 +78,7 @@ linux_info_list = [
     [rpwr, 'Hardware Model', '/usr/bin/hostnamectl', r'Hardware Model:\s*(?P<o>.*)'],
     [rpwr, 'Docker Info', 'docker info', r'(?P<o>.*)', re.IGNORECASE | re.DOTALL],
     [rpwr, 'Docker Version', 'docker version', r'(?P<o>.*)', re.IGNORECASE | re.DOTALL],
-    [rpwr, 'Docker containers', 'docker ps -a', r'(?P<o>.*)'],
+    [rpwr, 'Docker Containers', 'docker ps -a', r'(?P<o>.*)'],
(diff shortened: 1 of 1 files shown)


lib/hardware_info.py
```diff
@@ -77,6 +78,9 @@ def read_directory_recursive(directory):
     [rpwr, 'Hardware Model', '/usr/bin/hostnamectl', r'Hardware Model:\s*(?P<o>.*)'],
     [rpwr, 'Docker Info', 'docker info', r'(?P<o>.*)', re.IGNORECASE | re.DOTALL],
     [rpwr, 'Docker Version', 'docker version', r'(?P<o>.*)', re.IGNORECASE | re.DOTALL],
+    [rpwr, 'Docker Containers', 'docker ps -a', r'(?P<o>.*)'],
+    [rpwr, 'Installed System Packages', 'if [ -f /etc/lsb-release ]; then dpkg -l | cat ; elif [ -f /etc/redhat-release ]; then dnf list installed | cat; fi', r'(?P<o>.*)', re.IGNORECASE | re.DOTALL],
```
ribalba, line 82: Why are you using `| cat` here?

ribalba (approved):
Works on fedora. The only thing is why are you using the | cat ?

lib/hardware_info.py
ArneTR: happy to take recommendations here. `dpkg -l ` gives me an `ncurses` window when I call it unpiped. This is why I feed it to cat just to get an output.

This test was only on the command line actually. I did not test if the implied pipe to Python is sufficient. Will post an update

ribalba: I didn't check the ubuntu one only the fedora command. Here you don't need the cat

PR #444, opened 2023-08-31
Uses /env to run python file

ribalba:


ArneTR:
LGTM. Two remarks:
- Can you please move to escaped call in subprocess again. It should work like this: subprocess.run(['sudo', '/usr/bin/env', 'python3', python_file])
- Please a comment for the good idea that the file should be non writeable by the user after making it sudo available

ArneTR:
- Please also add the functionality that the hardware_info with the root rights is not modified directly (as this would hinder a git pull), but rather a git-ignored copy is made that will be cleaned on install / update by the script with root rights

ArneTR:
The so far made fixes should not be necessary now that we force a venv and Python is always running in venv mode.

The `sys.executable` and `which pythton3` should be identical. However the locking down of the hardware_info is still needed. Can you schedule this in at some point in the next weeks?

ribalba:
This is still open as I need to move the hardware_info file to another file to make it owned by root

ribalba:
Ready to merge IMHO

ArneTR:
The version commited works "by accident".

Actually the code suggests that actually the Python in the venv is used. That is not the case though.

WIth our new venv setup with correct include paths actually the `/usr/bin/env` should be removed and the actualy venv python binary location used.

To check that not the correct venv is used:
```
$ sudo /usr/bin/env python3 -c 'import sys; print(sys.executable)'
/usr/bin/python3
```

The expected line should be the same as when the venv is active:
```
$ /usr/bin/env python3 -c 'import sys; print(sys.executable)'
/home/gc/green-metrics-tool/venv/bin/python3
```


2. I would argue for removing `hardware_info_root.py` from the tree now, as it shows up as modfied file afterwards.

3. Main was not merged in before correctly. This leads to the fact that an addition to the hardware_info (`lshw`) got lost. Brought it back in.

4. hardware_info_root_original.py was not venv ready.

I commited a fix to this. When using venvs though one could actually notice that it does not understand the /usr/bin/venv logic:
```
(venv) gc@palit:~/green-metrics-tool$ sudo /usr/bin/env python3 /home/gc/green-metrics-tool/lib/hardware_info_root.py
  File "/home/gc/green-metrics-tool/lib/hardware_info_root.py", line 10, in <module>
    from lib.hardware_info import rdr, rpwr, get_values
ModuleNotFoundError: No module named 'lib'
```

ArneTR:
@ribalba Please review. Put you in

@dan-mm This is in dire need of a test. We need to check that actually some data is returned that is root readable only. Does Github VM have the directory `/sys/kernel/debug/sched` ?

PR #447, opened 2023-09-03
Frontend refactor

ArneTR:
## Changelog:
- Introduced Last 50 Runs as default home page
- Introduced View by repository view on separate page
- Frontend table creation moved to DataTables internal ajax calls for performance
- Renaming project_* => run_* everywhere
- Deletion of project_id column in CI table for now
- Removed all redefined-outer-name disablements. Having these is way more valuable than overwriting a variable
    + Talk to Didi: Overwriting time
- Removed last_run from the runs (old projects) table. Now a run only appears there if started and created_at is used as date marker.
- jobs.py will not pick up a job anymore that has a NULL machine_id value. Only directly assigned jobs will be scheduled.
- Emails are now only stored until the run is done or failed (14 days)
- Jobs.py is now a class to better reflect the state machine nature
- Timeline-Projects introduced, which allows to recurringly monitor software projects from the Dashboard
- Temporary config options notify_admin_for_own_software_add and notify_admin_for_own_run_ready removed as since the Timeline-Projects is now an integrated functionality they are not needed anymore. Admins never get mails for timeline projects unless an error occured.
- updated_at columns introduced for all tables that set current timestamp on update
- Runner.py does now insert it's own run and does not need to have it pre-inserted
Screenshot 2023-09-03 at 4 55 42 PM


ArneTR:
The tests will most likely be broken due to many refactorings. @dan-mm we have to discuss this during the week how to adapt the tests.

dan-mm:
Tests are passing for this branch again.

The one thing I would say can be done still is another pass at the test_jobs.py module. There's been a significant change in the behaviour of the jobs.py class, and the tests are more reflective still of the old behaviour - I can think of a few more tests that would make sense to be added to the test_jobs.py for better coverage.

I wanted to finish the other stuff on my todo list though, so just focused this round of checkins on bringing failing tests back online.

PR #452, opened 2023-09-05
Powermetrics is now checked if it runs already on the system

ArneTR:
@ribalba I removed a comment from you that I think was outdated in the NetworkProxy. Correct?

Can you quickly check if it is also working on your mac and then merge it in?

I refactored your `pgrep` call a little. I think it is cleaner this way ...


ribalba (approved):
lgtm merge

PR #457, opened 2023-09-08
Runner now adds arguments to run

ArneTR:
The runner now adds dynamically all passed arguments to the constructor to the database.

These values may indicate why an energy is higher or lower and will be diffed later.

This PR has to wait for *frontend-refactor* to be merged in


ArneTR:
Hey @dan-mm 

the tests do not work. My assumption: Your code seems to work only sometimes given a certain ordering of the tests.

As you can see in the Actions runners logs it executes this line: `Command '['git', 'clone', '--depth', '1', '-b', 'Test Branch', '--single-branch', '--recurse-submodules', '--shallow-submodules', 'Test URL', '/tmp/green-metrics-tool/repo']' returned non-zero exit status 128.`

Which should never be executed as the fixtures should delete it. But apparently it stays in the DB and then gets picket up.

However if I re-run the tests it works! So there is something off.

2. You are inseting a machine into the db here: 
`machine = Machine(machine_id=0, description='test-machine')`

Later on though you are referencing it as machine_id 1. 
Example: `Job.insert(name, url,  'Test Email', 'main', filename, 1)`

How did that ever work?

I am getting `psycopg.errors.ForeignKeyViolation: insert or update on table "jobs" violates foreign key constraint "jobs_machine_id_fkey"
E           DETAIL:  Key (machine_id)=(1) is not present in table "machines".`

dan-mm:
Took a look at this, I can answer #2 at least:

I checked what was going on, the test db does have a machine with ID 1 in it, but that was indeed added from another test ( test_runner.py, which uses the two_psu_providers.yml, which has this machine defined). Hence why it was passing. 

I've tried to make sure tests cleanup after themselves as I see it, but I didn't write all the tests myself (and in this particular case I didn't even know that it wrote anything to the DB). I should do a pass through all the tests though to look for any places like this where there isn't a proper cleanup happening.

As for your first point - I just took a second look and I see what you mean. Looking through the fixtures I think I see the issue - the cleanup fixtures in test_jobs have scope="module" so they run once per module, not per test. Hence why test_simple_run_job is trying to process the job from test_insert_job.

ArneTR:
Here another example from the main branch directly. One run works the other one fails, although they run on the same commit:

- https://github.com/green-coding-berlin/green-metrics-tool/actions/runs/6134251459
- https://github.com/green-coding-berlin/green-metrics-tool/actions/runs/6130539478

PR #467, opened 2023-09-19
446 ci page UI ux overhauls

dan-mm:
Corresponding Eco-CI PR here: https://github.com/green-coding-berlin/eco-ci-energy-estimation/pull/41

Though there shouldn't be breaking API changes, so this can be merged in first.

This has the workflow_name now as a field in the DB/API, and displayed in the eco-ci front end.

It also has delimiters for long numbers on the front end, with a helper function in converters.js (should this be in main.js instead?)

ci-index.html/js has also been rewritten to condense the rows by repo, as I think this will make it much easier to find what you're looking for. That list was getting too big already with even the small amount of users we have.


test/api/test_api.py
```diff
@@ -71,7 +72,10 @@ def test_ci_measurement_add():
...
     assert data is not None
     for key in measurement.model_dump().keys():
-        assert data[key] == measurement.model_dump()[key], Tests.assertion_info(f"{key}: {data[key]}", measurement.model_dump()[key])
+        if key == 'workflow':
+            assert data['workflow_id'] == measurement.model_dump()[key], Tests.assertion_info(f"workflow_id: {data['workflow_id']}", measurement.model_dump()[key])
+        else:
+            assert data[key] == measurement.model_dump()[key], Tests.assertion_info(f"{key}: {data[key]}", measurement.model_dump()[key])
 
 
 def todo_test_get_runs():
```
ArneTR, line 81: Althoug unrelated to the PR: Why is there a todo function here?

frontend/js/helpers/converters.js
```diff
@@ -44,4 +44,9 @@ const rescaleCO2Value = (value,unit) => {
     else if(value > 1_000_000) return [(value/(10**6)).toFixed(2), 'g'];
     else if(value > 1_000) return [(value/(10**3)).toFixed(2), 'mg'];
     return [value.toFixed(2) , unit];
-}
\ No newline at end of file
+}
+
+// Function to format 1000000 to 1,000,000
+const formatLongValue = (value, delimiter=",") => {
```
ArneTR, line 50: What is this function? Why write such a complex regex if there are functions in the stdlib?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

Also this function is buggy. It does what it should for the input but cannot be reused with proper integers / rationals.

Example: `a = 10000000.00; a.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",")`

frontend/js/ci.js
```diff
@@ -418,7 +424,8 @@ $(document).ready((e) => {
         const ci_data_node = document.querySelector('#ci-data')
         ci_data_node.insertAdjacentHTML('afterbegin', `<tr><td><strong>Repository:</strong></td><td>${repo_link_node}</td></tr>`)
         ci_data_node.insertAdjacentHTML('afterbegin', `<tr><td><strong>Branch:</strong></td><td>${escapeString(url_params.get('branch'))}</td></tr>`)
-        ci_data_node.insertAdjacentHTML('afterbegin', `<tr><td><strong>Workflow:</strong></td><td>${escapeString(url_params.get('workflow'))}</td></tr>`)
+        ci_data_node.insertAdjacentHTML('afterbegin', `<tr><td><strong>Workflow ID:</strong></td><td>${workflow_id}</td></tr>`)
+        ci_data_node.insertAdjacentHTML('afterbegin', `<tr><td><strong>Workflow:</strong></td><td>${workflow_name}</td></tr>`)
```
ArneTR, line 428: workflow_name needs no sanitation?

frontend/js/ci.js
```diff
@@ -343,7 +343,7 @@ const displayCITable = (measurements, url_params) => {
                             <td class="td-index">${run_link_node}</td>\
                             <td class="td-index">${escapeString(label)}</td>\
                             <td class="td-index"><span title="${escapeString(created_at)}">${dateToYMD(new Date(created_at))}</span></td>\
-                            <td class="td-index">${escapeString(value)}</td>\
+                            <td class="td-index">${formatLongValue(value)}</td>\
```
ArneTR, line 346: Why no escaping anymore? The custom function does nothing if a malicious string is passed in terms of replacement.

frontend/js/ci.js
```diff
@@ -310,7 +310,7 @@ const displayCITable = (measurements, url_params) => {
     measurements.forEach(el => {
         const li_node = document.createElement("tr");
 
-        [energy_value, energy_unit] = convertValue(el[0], el[1])
+        const [energy_value, energy_unit] = convertValue(el[0], el[1])
         const value = `${energy_value} ${energy_unit}`;
```
ArneTR, line 314: nice :)

frontend/js/ci-index.js
```diff
@@ -1,78 +1,95 @@
...
+                <td>
                     <div class="ui accordion" style="width: 100%;">
-                      <div class="title">
-                        <i class="dropdown icon"></i> ${uri_link}
-                      </div>
-                      <div class="content">
-                      </div>
+                        <div class="title">
+                            <i class="dropdown icon"></i>
+                            ${getRepoLink(repo, source)}
```
ArneTR, line 34: you are often reusing values in modfied format where I see no escaping before. What happens if source is already corrupted? will it not just display here?

ArneTR (changes requested):

test/api/test_api.py
dan-mm: this is from the front-end refactor, when we changed "projects" into "runs"

previously this test was "get_projects", and used to be a simple call as a quick effective smoketest for the api. it inserted a  project directly into the DB called the endpoint to see if we got the same back.

after the refactor the concept of projects changed into timeline_projects, whose endpoint relied on a more complicated logic that also relied on the machines table, and this test felt like it needed a more extensive re-write.

At the same time there was also a lot of other work which was needed to get the tests back up running generally, which I prioritized first and ended up being time consuming and over time-budget. As we had other tests in the same file that served the purpose of smoketesting the API, this remained as a todo.

Looking at this with fresh eyes though its not hard to get a working test for either the timeline_projects or runs endpoint, and our API  is lacking good test coverage, so its worth doing, but outside the scope of this PR

frontend/js/helpers/converters.js
dan-mm: I wasn't aware of this function in the stdlib - that's definitely the better solution. Refactored to remove my custom function and used that instead.

frontend/js/ci-index.js
dan-mm: I was out of the loop as to where (and more importantly, why) values needed sanitizing - I thought it was only for strings in user-set variables, and mostly for display reasons. I didn't know of it in terms of a security risk for any maliciously set variable (including ones that users can set via direct API calls, instead of the automated ones by eco-ci)

Looking at it through that lens, I did see quite a few places where we weren't sanitizing values in ci.js and ci-index.js, and I've updated those now.

frontend/js/ci.js
dan-mm: added sanitation

frontend/js/ci.js
dan-mm: added sanitation

frontend/js/ci.js
dan-mm: word

PR #469, opened 2023-09-20
Adding email blocking for requests to non-configured domain

ArneTR:
Currently the API does not always alert the admin in case of errors.

With a hotfix I have already integrated some more HTTP exceptions: https://github.com/green-coding-berlin/green-metrics-tool/commit/37a56fb84aea526fbe638062e502562c86864d4f

The result is atm that we are getting a lot of email noise from non-relevant malformed crawler requests.

This PR blocks all emails to requests like IP, reverse DNS PTR etc.

What is left to discuss is how we handle HTTP 400 errors, which are currently not triggered in the middleware, but the application code itself.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -8,7 +8,7 @@ import sys
 import os
 
 from xml.sax.saxutils import escape as xml_escape
-from fastapi import FastAPI, Request, Response, status
+from fastapi import FastAPI, Request, Response
 from fastapi.responses import ORJSONResponse
 from fastapi.encoders import jsonable_encoder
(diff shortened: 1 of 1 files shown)


ribalba:
Is there an easy way in which we can pipe all these "errors" into a logstash or similar and then categorise them? I used to to that with Django and then had certain rules to drop crawler messages for example

ArneTR:
As discussed internally the logging feature to a industry standard drain will be postponed for now in favor of a remote solution in the future.

The feature stays email only for now

PR #476, opened 2023-09-22
Import refactoring

ArneTR:
First version that uses `.pth` files in the `venv` to set a default include path

Downside: Now venvs are mandatory and must be activated for the GMT to work


ArneTR:
The tests are failing, but the code itself should work.

@ribalba Can you test if the new install script works in Fedora?

@dan-mm Can you look at the tests please. Major change here is that the tests folder has been renamed from "test" to "tests", because now loading by folder Python can get confused with modules and apparently there is a "test" module, but no "tests" module in the standard-lib

I am unsure if that is due to the tests itself or with the instrumentation on Github that the renaming leads to errors

ribalba:
What a refactor. Works on Fedora and giving the commit a quick read it all looks A OK.

ribalba (approved):

ribalba:
@ArneTR I did a tiny refactor the the pylint checks and pushed it to this branch.

metric_providers/psu/energy/ac/xgboost/machine/provider.py
```diff
@@ -1,13 +1,7 @@
...
 import os
 from io import StringIO
 import pandas
 
-CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
-sys.path.append(f"{CURRENT_DIR}/../../../../../../lib")
-sys.path.append(CURRENT_DIR)
-
-#pylint: disable=import-error, wrong-import-position
 import model.xgb as mlmodel
```
dan-mm, line 5: I get an error here when I try to run the runner with the XGBoost model enabled:

```
>   import model.xgb as mlmodel
E   ModuleNotFoundError: No module named 'model'

../metric_providers/psu/energy/ac/xgboost/machine/provider.py:5: ModuleNotFoundError
```
The model referenced in this branch is on the same commit as the one in main: https://github.com/green-coding-berlin/spec-power-model/tree/95f13ad95ced02350c75841616306115a1402e93

Does the spec-power-model need an update for this too? I see this error both in tests, and using the runner.py manually.

dan-mm:
Updated everything test related to reflect the new path. Just a few small changes here and there.

If I turn off the xgboost model in my test-config, everything else passes locally. The workflow should also be good too, though its currently also failing because of the xgboost model (and there's always a chance this hides a failure further down the pipeline.... but I don't expect that to be the case here)

ArneTR:
so what is the plan? I do not want to merge it if it fails with the XGBoost. That model should work.

Can you post the error here?

dan-mm (changes requested):

dan-mm (approved):

ArneTR:
Status from today: It is unclear why the `lib.db` import is not working. This will be fixed tomorrow morning by Dan

PR #487, opened 2023-09-28
Removing CpuFrequencySysfsCoreProvider from default providers

ArneTR:


ArneTR:
@dan-mm Can you also please give this a look? The problem here seems to be that the Network Proxy Provider gets activated ... I remember you had a piece of code for the VM that just activates all providers first and then deactivates them (or was it the other way round?)

The provider gets activated however, but it should not. This is why the docker pulls fail

dan-mm:
Should be all good now

PR #507, opened 2023-10-13
Refactors the system checkers

ribalba:


ArneTR:
LGTM. Tested on Linux box and works fine

ArneTR (approved):
LGTM. Tested on Linux box and works fine

PR #514, opened 2023-10-24
Added lshw to hardware_info_root

ArneTR:
Added lshw output, as we had no info about PCI, GPU, etc. info before.

The command delivers output without out, but is more comprehensive as root


ribalba (approved):
Added the packages to the install script. Maybe you want to test under Ubuntu but otherwise looks good to merge

ArneTR:
Package `lshw` is default on Ubuntu installed. Does not hurt to add to install though. ty

PR #516, opened 2023-10-26
Calibrate

ribalba:
Adds a little script that calibrates the temperature cool down period


ArneTR:
Please assign me as reviewer  if this is final and needs review. Have removed me atm

ribalba:
@ArneTR Ready for the next round of testing

ArneTR:
has main been merged in here? Also I see that a merge conflict with main is still lingering ... ?

ribalba:
no, main has not been merged

ribalba:
@ArneTR merged

ArneTR:
I did quite some reworking and also found some bugs. Still some questions are open that we have to address:

- Provider's overhead is now calculated with actual configured resolution. Variable name was thus changed to provider_interval_override
- On fail now more logging is outputted, even without setting -v or -d flag
- Methods all at top and more directly graspable flow in main()
- Introduced minimum length for STD and MEAN calculations. Calculating a StdDev on less than 3 results makes no sense.
- added global metric_providers    global timings where applicable notes
- docker_sleeper was not killed, because global was not set
- moved initial logger warning below actual instantiation of logging format
- added formatting to logger with color
- renamed first idle to "baseline" which is what it actually is according to Blue Angel naming schema


**Points that I coud not address**:
- Outliers check is not sufficient. It would not calibrate correctly if you suddenly ramp up energy during 50% of the calibration and the StdDev becomes extremely large ... as the outliers threshold thus also become larger
  - Here a check for relative a std_dev must be done. I added 3% for now. 1% should be a target for our machines at least
  - Currently I am getting 25% deviance due to the nature of how the metrics are reported. They are energy and not time normalized. Shall we normalize them or just use power?
 
- Energy overhead must be normalized per second. If you run the script longer the overhead will be longer. Either we normalize per second, or we use power, or we make a relative ratio.

- Could not refactor two very similar method for making the first idle and checks, and then the later idle with included temperature checks. Can it be done? Or did you specifically opt for so much duplicate code as it was not feasible to split?

ArneTR:
@ribalba Added some functionalities for calculating overhead.

It still does not work with the machine energy provider as the nature of the switched power supply provides to much jitter and cannot get the STDDEV low enough.

However the script is very nice to get the provider overhead. Have documented it all here: https://docs.green-coding.io/docs/measuring/metric-providers/overhead-of-measurement-providers/

PR #518, opened 2023-10-26
Tag dockerhub containers

dan-mm:
- build containers workflow will now get the list of images to build based the changed files in the PR:
    - It looks for Dockerfiles that were touched under `docker/auxiliary-containers/**/Dockerfile' and identifies those subfolders as the ones that need to be build
    
 - build and push containers script will now check dockerhub for the latest tag
     - if the response is null, it sets the tag to v1
     - if the latest tag is in the format "v##", it increments the ##
     - else, it sets it as "latest"


ArneTR:
Have not checked the bash commands manually but logic looks ok to me

PR #523, opened 2023-10-31
Update Dockerfile - Removed playwright version pinning

ArneTR:


ArneTR:
The Container updated workflow did not trigger here. For me this is unexpected  ... the files touched this directory? Trying the manual route now

ArneTR:
Update: it did seem to run, but did not show here ... quite unexpected still. Should it show here?

dan-mm:
I don't think so, and I wouldn't necessarily expect it since it is only triggered on PR closed (successfully). I believe the PR's that show up underneath the discussion here are just for on PR push/open triggers (aka, "is this PR safe to merge?", not "do this once this PR has been merged") 

That being said I'm not seeing any documentation one way or another from a quick search.

PR #527, opened 2023-11-06
test PR for eco-ci

dan-mm:


ArneTR:
@dan-mm Formatting is broken in the old post and a stray HTML tag is flying around.

Can you give this a look after making a first protype for the codespaces thing please. I think the version is not critically broken, but more like "room for visual improvement". It kinda does what it should :)

PR #530, opened 2023-11-06
Makes error optional except if there is data we need

ribalba:


ArneTR:
Implementation and structure looks good to me. I cannot really check the validation logic due to missing domain knowledge.

I was so bold and did a ChatGPT by just saying: "This code has a bug, can you fix it?".

This is the reply:


> You have a bug where you try to call exc.errors() on the ValueError exception. ValueError does not have an errors() method. To fix this, you should just print the exception message directly.

I think it is right. When you actually call .errors() on the second exception object it will crash.

api/main.py
```diff
@@ -592,6 +634,13 @@ async def hog_add(measurements: List[HogMeasurement]):
...
         except RequestValidationError as exc:
+            print(f"Caught Exception {exc}")
+            print(f"Errors are: {exc.errors()}")
+            email_helpers.send_admin_email('Hog parsing error', str(exc.errors()))
+
+        try:
+            validate_measurement_data(measurement_data)
+        except ValueError as exc:
             print(f"Caught Exception {exc}")
             print(f"Errors are: {exc.errors()}")
```
ArneTR, line 645: Please remove this line

ribalba:
@ArneTR do you want to do a new deploy?

PR #537, opened 2023-11-13
Adds the citation file

ribalba:
Closes https://github.com/green-coding-berlin/green-metrics-tool/issues/533

For now I have only added it to the GMT. We should remember to add a similar file to the other repos when they become "big"

Also I have opted for only including the "core" developers as authors as otherwise we would habe to modify the file every time someone contributes


ArneTR:
perfect

PR #541, opened 2023-11-16
Allow digits in env var keys

### LINKED ISSUE #540: Allow digits in environment variable keys
Currently, digits in keys of env vars as part of the compose file are not allowed. However, digits are quite common in Linux env vars and are allowed by the [Open Group Base Specification](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html) (A.K.A. POSIX Regulation):

> Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in [Portable Character Set](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap06.html#tagtcjh_3) and do not begin with a digit.

davidkopp:
Closes #540

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -751,12 +751,12 @@ class Runner:
                     else:
                         raise RuntimeError('Environment variable needs to be a string with = or dict and non-empty. We do not allow the feature of forwarding variables from the host OS!')
 
-                    if not self._allow_unsafe and re.search(r'^[A-Z_]+$', env_key) is None:
+                    if not self._allow_unsafe and re.search(r'^[A-Z_]+[A-Z0-9_]*$', env_key) is None:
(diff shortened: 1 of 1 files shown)


ArneTR:
@dan-mm The test pipeline cannot execute this external PR. Can you check why.

ArneTR:
The test pipeline failed on this one due a configuration error. We fixed it but cannot make the tests run for this PR. (https://github.com/green-coding-berlin/green-metrics-tool/pull/546)

I created a copy PR with the fix to run the pipelines. Everything successful!

So this will get merged.

ty @davidkopp <3

Issue #545, opened 2023-11-18
Unclear Unit Test


ArneTR:
Hey @dan-mm 

I do not understand this test: https://github.com/green-coding-berlin/green-metrics-tool/blob/c13d73c79cb363f32cb55566401f377f4405f1c6/tests/test_usage_scenario.py#L73

It references the wrong values in this file: https://github.com/green-coding-berlin/green-metrics-tool/blob/main/tests/data/usage_scenarios/env_vars_stress_unallowed.yml

How I understand the test it will only check if *any* of the variables will trigger a fail, but not everyone individually. Which is what we want, not?

ArneTR:
Als I do not understand the next two tests in the file:

```
def test_env_variable_skip_unsafe_true():
    runner = Tests.setup_runner(usage_scenario='env_vars_stress.yml', skip_unsafe=True)
    env_var_output = get_env_vars(runner)
    assert env_var_output == 'alpha-num123_\n', Tests.assertion_info('alpha-num123_', env_var_output)

def test_env_variable_allow_unsafe_true():
    runner = Tests.setup_runner(usage_scenario='env_vars_stress.yml', allow_unsafe=True)
    env_var_output = get_env_vars(runner)
    assert env_var_output == 'alpha-num123_\n', Tests.assertion_info('alpha-num123_', env_var_output)
```

Why are you testing here with skip_unsafe if the variables are anyway safe to use? Shouldn't you test on the https://github.com/green-coding-berlin/green-metrics-tool/blob/main/tests/data/usage_scenarios/env_vars_stress_unallowed.yml file rather than the https://github.com/green-coding-berlin/green-metrics-tool/blob/main/tests/data/usage_scenarios/env_vars_stress.yml ?

dan-mm:
for the latter two tests (test_env_variable_skip_unsafe_true && test_env_variable_allow_unsafe_true) you're right. I've gone through and tried to understand why I did it that way, but don't see the logic. I think it was just a mistake on my part. I see you've made the changes for these two tests in https://github.com/green-coding-berlin/green-metrics-tool/pull/546 - I've checked them out and it looks good.

[...]

PR #546, opened 2023-11-18
Env vars numbers tests

ArneTR:
Needed to make another PR as the test pipeline currently has a hickup.

This includes the fix from @davidkopp here: https://github.com/green-coding-berlin/green-metrics-tool/pull/541

Also I added tests and changed the logic of the env vars checking.

### DIFF
--- a/tests/data/usage_scenarios/env_vars_stress.yml
+++ b/tests/data/usage_scenarios/env_vars_stress.yml
@@ -11,6 +11,7 @@ services:
     build: .
     environment:
       TESTALLOWED: 'alpha-num123_'
+      TEST1_ALLOWED: 'alpha-key-num123_'
 flow:
   - name: Stress
     container: test-container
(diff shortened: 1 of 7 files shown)


ArneTR:
@dan-mm Can you please give this a look. You did not check the env vars in the skip case if not present, but just checked for exceptions.

Also you did not check if in allow_unsafe they were actually set.

please give it a look if I did not introduce an error.

dan-mm:
it looks good to me, and I pulled and ran the tests locally - they all still pass.

based on the feedback from #545 I also extended the test_env_variable_no_skip_or_allow test into 3 tests, which check that each "unallowed" env var triggers a fail. This adds about +30 seconds to the test run though and I'm not convinced it provides enough added value, though it is better coverage. Should I check it in, or what are your thoughts?

davidkopp:
I think it makes sense to use `dry_run=True` in the runner for the execution of these tests. The avoidance of the sleep saves a lot of time and the sleep is not relevant for testing if the env vars are correct or not.

dan-mm:
That's a really good point actually - I had forgotten we had that flag. I just checked it with these specific tests and it cuts the time down to about a third. We could probably save a lot of time over our whole test suite by using this as the default - I'll make sure to check that. For now I just added it to these tests.

dan-mm (approved):

PR #553, opened 2023-11-21
Breaking the workflow

ArneTR:
Because we had issues with external contributors making a submission and the workflow file was not working.

We could not alter the workflow file anymore and the PR had to be used as is.

This PR shall check if an external PR can actually modfiy the workflow file. And if not: How do we then updated workflows? Can it only happen AFTER a merge has been done to main?


ArneTR:
closed. was just for test

PR #566, opened 2023-11-29
Adds skip_checks to providers

ribalba:


metric_providers/powermetrics/provider.py
```diff
@@ -55,17 +57,21 @@ def stop_profiling(self):
...
-            print(f"Waiting for powermetrics to shut down (try {count}/60). Please do not abort ...")
-            time.sleep(1)
-            count += 1
-            if count >= 60:
-                subprocess.check_output('sudo /usr/bin/killall -9 powermetrics', shell=True)
-                raise RuntimeError('powermetrics had to be killed with kill -9. Values can not be trusted!')
-
-        # We need to give the OS a second to flush
-        time.sleep(1)
+        if self._skip_check:
```
ArneTR, line 60: If you run into a Permission error in the block before this (the except block) you would still kill the process, which is likely not what you want with that code. I think that should also be guard-claused.

ArneTR (commented):
I left a comment for a IMHO opionin missing guard clause.

Also I do see the need for having the GMT work when PowerHog is installed, but would like to brainstorm a bit if there cannot be an alternative solution other than having this waiting loop.

In the end we have the case that the Hog or another process might need the powermetrics support. Even if we instruct the GMT to kill it by force the daemon restarts it so quickly, that there is no way of getting it properly shut down.

So we run into an unsolvable conflict: If Hog is present the measurement with the GMT will be less useful. If we force it the GMT measurement will be less helpful, but at least the HOG data is uninterrupted (if fix proposed for the except block is applied).
So I see the argument for having this mode. It keeps at least one of the programs data good.

However it would be nice if the user can also opt for the case of favoring the GMT run over the HOG data. This is at the moment done by having the flag not set and getting a warning.

So the case in order to have "both", aka the HOG keeps running and the GMT will only get some overhead, but will not run into conflict, we could try the route to supply "stop_profiling" with a longer waiting amount. At the moment this waits for 5 secods for the forked process. Maybe if we use 60+ seconds here, and not in the loop after killall we get better results and can run both side by side?

metric_providers/powermetrics/provider.py
```diff
@@ -35,43 +41,44 @@ def __init__(self, resolution, skip_check=False):
...
+                                                   'Please close it before running the Green Metrics Tool.\n'
+                                                   'You can also override this with --skip-system-checks\n')
 
-    def is_powermetrics_running(self):
-        ps = subprocess.run(['pgrep', '-qx', 'powermetrics'], check=False)
-        if ps.returncode == 1:
-            return False
-        return True
+    def is_our_powermetrics_running(self):
+        pw_processes = subprocess.check_output(['pgrep', '-ix', 'powermetrics'], text=True)
```
ArneTR, line 52: Why using text=True here instead of encoding="UTF-8" ?

I think both do the same but I would opt for being consistent throughout the codebase.

metric_providers/powermetrics/provider.py
```diff
@@ -11,6 +11,12 @@
 
 class PowermetricsProvider(BaseMetricProvider):
     def __init__(self, resolution, skip_check=False):
+        # We get this value on init as we want to have to for check_system to work in the normal case
+        pw_processes = subprocess.check_output(['pgrep', '-ix', 'powermetrics'], text=True)
+        self._pm_process_count = len(pw_processes.strip().split('\n')) if pw_processes else 0
+        print(f"NUMBER BEFORE: {self._pm_process_count}")
```
ArneTR, line 17: This is very unspecifc debug message. Suggestion: "Currently already running powermetrics on the system before starting GMT powermetrics process"

metric_providers/powermetrics/provider.py
```diff
@@ -35,43 +41,44 @@ def __init__(self, resolution, skip_check=False):
...
+        return  minus_startup >= 1
 
 
     def stop_profiling(self):
         try:
             # We try calling the parent method but if this doesn't work we use the more hardcore approach
             super().stop_profiling()
         except PermissionError:
-            #This isn't the nicest way of doing this but there isn't really any other way that is nicer
+            # This isn't the nicest way of doing this but there isn't really any other way that is nicer that doesn't
```
ArneTR, line 63: I would like to have proper explanatory comment here, because we will forget why we implemented it that way.

Also the comment in the "try" block before is very misleading.

Suggestion:
```
try:
            # We try calling the parent method but since we do not have sudo rights the process running with root rights will ignore the signal. We still call the method because it might contain tear-down functionality in the future that we want to trigger
            super().stop_profiling()
        except PermissionError:
            # We will land here in any case as stated before (root permissions missing). When we trigger *killall* now the process will be terminated. We opted for this implementation as other processes on the system, like for instance the power hog (https://github.com/green-coding-berlin/hog) should not be affected too much. They restart the process anyway when it gets killed. However manual processes that the user might have started will also be killed, so we issue a notice.
# There is really no better way of doing this as of now. Keeping the process id for instance in a hash and killing only that would also fail du to root permissions missing. If we add an /etc/sudoers entry with a wildcard for a PID we open up a security hole. Happy to take suggestions on this one!

```

metric_providers/powermetrics/provider.py
```diff
@@ -35,43 +41,44 @@ def __init__(self, resolution, skip_check=False):
...
+            # open up a huge security hole
             subprocess.check_output('sudo /usr/bin/killall powermetrics', shell=True)
             print('Killed powermetrics process with killall!')
+            if self._pm_process_count > 0:
+                print('----------------------------------------------------------------------------------------------')
+                print('This means we will have also killed the already running powermetrics process. Please restart!')
```
ArneTR, line 69: Suggestion as I think it is confusing what exactly to restart and also referencing "the process" ... that implies too much knowlege: 
"This means we will have also killed any other already running powermetrics process. Please restart them if needed!

metric_providers/powermetrics/provider.py
```diff
@@ -35,43 +41,44 @@ def __init__(self, resolution, skip_check=False):
...
-                    subprocess.check_output('sudo /usr/bin/killall -9 powermetrics', shell=True)
-                    raise RuntimeError('powermetrics had to be killed with kill -9. Values can not be trusted!')
+        count = 0
+        while self.is_our_powermetrics_running():
+            print(f"Waiting for powermetrics to shut down (try {count}/60). Please do not abort ...")
+            time.sleep(1)
+            count += 1
+            if count >= 60:
+                subprocess.check_output('sudo /usr/bin/killall -9 powermetrics', shell=True)
+                raise RuntimeError('powermetrics had to be killed with kill -9. Values can not be trusted!')
```
ArneTR, line 81: Actually what exactly happens in this case? If powermetrics is killed and then a Runtime error is raised the measurement is broken in any case, as no processing is done anymore, right?
Wouldn't it be better to just continue parsing the values and issue a warning that data might be missing as no flusing has occured?

ArneTR (changes requested):

metric_providers/powermetrics/provider.py
ribalba: As far as I can see "text=True" uses the systems encoding and is preferred to forcing the encoding. In most cases UTF is the system encoding so it doesn't really matter. Maybe we should think about changing the UTF to text at some stage.

metric_providers/powermetrics/provider.py
ribalba: This was actually debugging for me and I forgot it. I think we should remove it as there isn't really much value in this message for the average user.

ArneTR: ok with the removal., but would also be ok with the long text statement I proposed

metric_providers/powermetrics/provider.py
ArneTR: I did a similar research when waging if going for one or the other switch. Actually what I found is that users generally prefer a UTF-8 encoding, even if the have non-classic charsets like GB* (chinese).

I would strongly prefer being forceful and explicit, rather than using system encoding ... at least until we have a breaking case.

metric_providers/powermetrics/provider.py
```diff
@@ -33,40 +41,45 @@ def __init__(self, resolution):
...
+        pw_processes = subprocess.check_output(['pgrep', '-ix', 'powermetrics'], text=True)
+        total_count = len(pw_processes.strip().split('\n')) if pw_processes else 0
+        minus_startup = total_count -  self._pm_process_count
+        return  minus_startup >= 1
 
 
     def stop_profiling(self):
         try:
             # We try calling the parent method but if this doesn't work we use the more hardcore approach
             super().stop_profiling()
```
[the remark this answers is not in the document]
ribalba: In the olden days we didn't have a shell that enables this behavior. We are leaving it in just to be sure.

metric_providers/powermetrics/provider.py
ribalba: I also thought about this. The problem is that we don't really know how much data has actually been flushed. If the process hasn't died after 60 seconds something bad has happened and I would argue the system was in an inconsistent state. So I wanted to fail big and not have any values available at all.

ribalba:
@ArneTR feedback incorporated and added the utf check

ArneTR:
k, looks good. re-running the tests again as the networkProxyProvider is having some hickups on Github lately ... this is afaik a transient error

ArneTR:
github shows an empty merge force-push for your last commit. Was that indended?

ribalba:
For some reason GitHub is not showing the merge commit files. In my local git they are showing up. Chose a graphical representation to make it easier to see right away. I also didn't force push. I simply did
`git merge main && git push`

Screenshot 2023-12-21 at 08 45 55

ArneTR:
Weird that git shows the merge incorrectly. But I can see it locally just fine.

However I do not understand your fix. The error is not actually empty. Tinyproxy seems to fail.

Did you see that line? https://github.com/green-coding-berlin/green-metrics-tool/actions/runs/7282733628/job/19851217512#step:5:4931

`RuntimeError: Stderr on NetworkConnectionsProxyContainerProvider was NOT empty: b'/usr/bin/tinyproxy: Could not create listening sockets.\n'`

Let's phone tomorrow about it.

ribalba:
This pr has blown out of proportion for what it actually set out to do. Sorry for the chaos. So now it also fixes testing under MacOS and fixes quite a few things with power metrics like the wrong return codes when there was no power metrics process running. Also it fixes a few smaller errors like that powermetrics was called with `-o` and piped into a file which in some cases cases the file to be corrupted. Also it fixes std output being a byte stream.

ribalba:
@ArneTR ready for your comments. 

I have decided not to implement file support/ or non redirection as it will just blow up the base.py even more. I have opted for catching the case in which the file is empty and returning. Sending a SIGIO does nothing but I am keeping it in as it is good practice.

metric_providers/powermetrics/provider.py
```diff
@@ -181,7 +217,14 @@ def read_metrics(self, run_id, containers=None):
     def get_stderr(self):
         stderr = super().get_stderr()
 
-        if stderr is not None and str(stderr).find('proc_pid') != -1:
+        if stderr is not None and str(stderr).find('proc_pid') != -1 :
```
ArneTR, line 220: Will that really do what we want? How I understand it you are returning here if the string "proc_pid" is in stderr.

However actually we want to return if ONLY this string is in stderr, not?

So we would have to iterate over all the lines and match?

metric_providers/powermetrics/provider.py
```diff
@@ -181,7 +217,14 @@ def read_metrics(self, run_id, containers=None):
...
+            return None
+
+        # This has been showing up and we don't really understand why. Google has no results and looking at the
+        # strings of powermetrics doesn't show anything. There also seems to be no correlation with the interval.
+        # A shame we can't look into the code and figure this one out. For now we just ignore it as we don't really
+        # have any other chance to debug.
+        if stderr is not None and str(stderr).find('Second underflow occured') != -1 :
```
ArneTR, line 227: Will that really do what we want? How I understand it you are returning here if the string "Second underflow" is in stderr.

However actually we want to return if ONLY this string is in stderr, not?

So we would have to iterate over all the lines and match?

ArneTR (changes requested):
I am now done with all my other PRs and now ready to tackle this.

It will require quite some merging as soon as the other PRs are aready as process killing and UTF-8 stderr are now processed differently.

I will take this on, but need a feedback on the comments I have put in here in this new review.

metric_providers/powermetrics/provider.py
ribalba: @ArneTR good point. Fixed please review

Issue #423, opened 2023-08-14
Way to check if containers have booted


ribalba:
Flows are currently failing because containers are taking longer to boot than we currently wait in the `runner.py`. Currently I modify the runner to have a custom wait. 
I see the following options to fix this:
- Add a variable to the usage_scenario where we can set a custom wait time if we know that containers might take longer to boot
- Add a variable to the config.yml to wait in the boot phase. Not a big fan of this as development with the GMT is already quite time consuming/ annoying based on all the sleeps. And this would be for all runs even the ones we don't need the sleep as the containers don't need the time to boot.
- Offer some checker to see if a container has booted with a timeout[1][2]. I imagine something like:
```
...
services:
  a:
    container_name: a
    image: a
    healthcheck:
      test: ["CMD-SHELL", "true"]
      interval: 10s
      timeout: 5s
      retries: 5
```
As soon as we see this in the services we start a routine to check that all containers have started before continuing.

[1] https://docs.docker.com/engine/reference/builder/#healthcheck
[2] https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck

ArneTR:
Nice find with the healthcheck option, was not aware that existed.

To my understanding when a container is not healthy an event is triggered, so we could leverage that directly and not roll on logic on top.
I would really like to understand how that is actually used in the wild then. Is the signal only know to the host, or does the container also get the signal? If so, does the container typically respond with a rountine? Or if only the host gets the signal, is there a default behaviour .... like for instance restarting the container?

However I think the most tricky part is how to integrate it into our measurement model without introducing randomness or overhead. Technically it can also be integrated between the measurement steps to not create randomness in the data. However the question is then, is the lazy startup time not actually part of the boot process?

I think this needs quite some further debate ...

ArneTR:
Referencing: https://github.com/green-coding-berlin/green-metrics-tool/issues/567

ArneTR:
Healthcheck has now been implemented https://github.com/green-coding-berlin/green-metrics-tool/pull/637

PR #568, opened 2023-11-30
Proposal: Use depends_on for container startup order

### LINKED ISSUE #567: Compose flag `depends_on` gets ignored
GMT currently doesn't handle service dependencies defined by the Docker Compose key [depends_on](https://docs.docker.com/compose/compose-file/05-services/#depends_on). That is a problem, because in usage scenarios there could be dependencies between services that have to be respected during startup. If not, the startup of services could fail.

Known workaround:
Change the order of services in the compose file.

davidkopp:
This PR tries to resolve #567.

It uses `depends_on` to determine a better startup order, and it checks if the dependent containers are running before starting the respective container.

There are some open questions that have to be discussed. Therefore, this PR is currently more like a proposal.

- Would this PR be problematic for the current measurement model of the boot phase?
- What is the best way to check if the dependent containers are ready? The check if their state is "running" is actually not sufficient. A health check as proposed in #423 would be better
- What should be configurable regarding the waiting time for dependent containers?
- The [Compose Specification of depends_on](https://docs.docker.com/compose/compose-file/05-services/#depends_on) states, that it also removes the services in dependency order. I think this is not relevant for GMT. Or is it relevant to have a proper measurement of the removal phase?


ArneTR:
That PR looks excellent!

The logic looks sound and from just looking at it should do exactly what `depends_on` advertises.

Having the healthcheck logic also integrated would be a dream, but I would argue it is more reasonable to merge this first and then we keep the other issue open for people interesting to open a PR on this.

I will give it a more detailed look tomorrow and make a proper review.

runner.py
```diff
@@ -637,18 +638,39 @@ def setup_networks(self):
             self.__networks.append(network)
             self.__join_default_network = True
 
+    # Order service names based on 'depends_on'
+    def order_service_names(self, order_array, service_name):
```
ArneTR, line 642: since you are using a recursin approach this has to be guarded for a cycle.

Example:
```docker
---
services:
  test-container-1:
    image: alpine
    depends_on:
      - test-container-2
      - test-container-3
  test-container-2:
    image: alpine
  test-container-3:
    image: alpine
    depends_on:
      - test-container-4
  test-container-4:
    image: alpine
    depends_on:
      - test-container-2
      - test-container-4
```

If you feed this file into `docker-compose` it will fail with: 
```log
cycle found: test-container-4 -> test-container-4
```

The python code will fail with:
```log
  File "/private/tmp/test.py", line 27, in <module>
    order_of_service_names = order_service_names(order_of_service_names, service_name)
  File "/private/tmp/test.py", line 14, in order_service_names
    order_array = order_service_names(order_array, dep)
  File "/private/tmp/test.py", line 14, in order_service_names
    order_array = order_service_names(order_array, dep)
  File "/private/tmp/test.py", line 14, in order_service_names
    order_array = order_service_names(order_array, dep)
  [Previous line repeated 995 more times]
  File "/private/tmp/test.py", line 13, in order_service_names
    if dep not in order_array:
RecursionError: maximum recursion depth exceeded in comparison
```

This needs a cycle protection

runner.py
```diff
@@ -787,10 +809,38 @@ def setup_services(self):
             if 'cmd' in service:  # must come last
                 docker_run_string.append(service['cmd'])
 
+            # before starting the container, check if the dependent containers are "ready"
+            # if not, wait for them
+            if 'depends_on' in service:
+                for dependent_container in service['depends_on']:
```
ArneTR, line 815: It would be nice to have a condition here that checks if the long form or the short form is used.

The short form is how this PR currently behaves. However the long form would also have features like conditions, health_checks etc.

I imagine a condition check that then throws a RuntimeError if the long form is used.

However this can be "overridden" by supplying an argument to the runner like `--skip-unsupported`.

Just not using the variables silently might introduce quite some confusion. What do you think?

Actually we could use this flag then in an extended form and throw this warning for every key in the `compose.yml` that we do not support 🤔

ArneTR (changes requested):
I gave the PR a closer look. The approach is great!

I put some comments and remarks. Please give it a look and share your thoughts.

runner.py
```diff
@@ -787,10 +809,38 @@ def setup_services(self):
...
+                    # TODO: Make some options configurable through config.yml
+                    max_waiting_time = 20
+                    sleep_time_per_iteration = 1
+                    time_waited = 0
+                    state = ""
+                    while time_waited < max_waiting_time:
```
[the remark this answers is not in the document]
davidkopp: I created the waiting loop for the case in which a container needs a long time to boot, and I want to wait for it until it is "ready". With the current code this is not the case, because it just checks if the container is "running" and that is the case almost immediately after it was created by Docker. To actually check if the container is ready, we need the health check.
So for now this waiting loop is most likely useless. I will remove it for this PR.
However, if the health check is implemented, something like a waiting loop is needed, I guess.

Regarding your last question: I don't think so. Therefore, I think it makes sense to throw an error immediately if a dependent container was not started/is not running.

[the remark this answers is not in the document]
ArneTR: I would argue to keep the waiting loop in, even if it should abort in the first loop run. Maybe a comment helps that this is pre-work for having the health_check coming.

2. You are right about my sample with the alpine container. That was not representative.

But then think of a container that overwrites the entrypoint and sets CMD ["sleep", "1"]

This would exit after for example even if a terminal is attached. So our method of starting does not protect from this behaviour fully.

Thus I would argue even more to keep the loop in as it would nicely catch a directly dropping container.

ArneTR:
The key in the config is aptly named. Can you put it under "measurement" though. I think this makes more sense, as we already have config vars there that influence measurements.

So hierarchy would be measurement -> boot -> wait_time_dependencies

2. Check for long syntax looks good.

3. I believe the cycle check is not doing what it should. It will find if a resource is lisitng itself, but it does not find a real cycle.
Imagine a setup of 6 containers that always list the next one, and the last then listing the first again. This will lead to a cycle also and is currently not checked.

Let me know if you need a sample yml.

davidkopp:
The config looks now as follows:

```yaml
measurement:
   idle-time-start: 10
   idle-time-end: 5
   flow-process-runtime: 3800
   phase-transition-time: 1
   boot:
     wait_time_dependencies: 20
   metric-providers:
```

Is it the way you had in mind?

The faulty implementation of the cycle detection is a little embarrassing. Should be fixed now.

runner.py
davidkopp: Created a new issue for this:
https://github.com/green-coding-berlin/green-metrics-tool/issues/580

ArneTR:
The position of the config option is exactly how I had it in mind, thanks for moving that!

@ribalba can you have a look at the cycle protection implementation and how David implemented it. Am a bit short in time today and tomorrow.
@ribalba Please also create a more complex test case with a bigger hierarchy of the containers with more than one link between containers just be on the safe side.

runner.py
```diff
@@ -637,18 +638,51 @@ def setup_networks(self):
...
         # This use case is when you have running containers on your host and want to benchmark some code running in them
-        for service_name in self._usage_scenario.get('services', []):
-            print(TerminalColors.HEADER, '\nSetting up containers', TerminalColors.ENDC)
+        services = self._usage_scenario.get('services', {})
+
+        order_of_service_names = []
+        # Check if there are dependencies defined with 'depends_on'.
+        # If so, change the order of the services accordingly.
+        for service_name in services.keys():
+            order_of_service_names = self.order_service_names(order_of_service_names, service_name)
```
ribalba, line 674: Hey, thank you for this great PR again. Arne and me just talked about the code and while this code is fully functional we see that it might be hard to maintain in the future as you carry the whole list `order_of_service_names` or `order_array` into every sub iteration. While this is fine for now we see that this might be the source of errors in a few years time as it is harder to comprehend. Also this pattern can be the source of errors when doing more complex things. When calling recursive functions it might be better to only pass in the context of the recursion and then handle the building of the list in the parent. Not something we need to fix right now but just as a remark.

ribalba (approved):

ribalba:
As soon as this is merged I will add a more complex test. As I can't write to @davidkopp repo.

runner.py
davidkopp: Based on this feedback I created an alternative PR with some refactorings:
https://github.com/green-coding-berlin/green-metrics-tool/pull/593

ribalba:
This was merged here https://github.com/green-coding-berlin/green-metrics-tool/pull/593#issuecomment-1862944828

PR #574, opened 2023-12-01
Allow characters . / : in env var values

### LINKED ISSUE #569: Allow more characters in environment variable values
Currently, only values in the environment section of a compose file are allowed that match the regex `^[a-zA-Z0-9_]+[a-zA-Z0-9_-]*$`.

I have an application with a Docker Compose file that uses the following values:
- `http://backend:8080`
- `org.postgresql.Driver`
- `jdbc:postgresql://postgres/db`

These values are not allowed, because of the characters `:`, `/` and `.`
It is possible for me to change the application, so it doesn't need these environment variables anymore. Nevertheless, I want to ask if maybe some more characters could be allowed or the strict regex is required due to security concerns?

davidkopp:
Fixes #569

Also checks if the value is not longer than 1024.

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -759,11 +759,11 @@ class Runner:
                         raise RuntimeError(f"Docker container setup environment var key had wrong format. Only ^[A-Z_]+[A-Z0-9_]*$ allowed: {env_key} - Maybe consider using --allow-unsafe or --skip-unsafe")
 
                     if not self._allow_unsafe and \
-                        re.search(r'^[a-zA-Z0-9_]+[a-zA-Z0-9_-]*$', env_value) is None:
+                        re.search(r'^[a-zA-Z0-9_]{1}[a-zA-Z0-9_.:/-]{0,1023}$', env_value) is None:
                         if self._skip_unsafe:
-                            print(TerminalColors.WARNING, arrows(f"Found environment var value with wrong format. Only ^[A-Z_]+[a-zA-Z0-9_]*$ allowed: {env_value} - Skipping"), TerminalColors.ENDC)
+                            print(TerminalColors.WARNING, arrows("Found environment var value with wrong format. Only ^[a-zA-Z0-9_]{1}[a-zA-Z0-9_.:/-]{0,1023}$ allowed: " + env_value + " - Skipping"), TerminalColors.ENDC)
                             continue
-                        raise RuntimeError(f"Docker container setup environment var value had wrong format. Only ^[A-Z_]+[a-zA-Z0-9_]*$ allowed: {env_value} - Maybe consider using --allow-unsafe --skip-unsafe")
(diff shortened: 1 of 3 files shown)


ribalba:
Thank you so much for the PR. I think there are files missing.

runner.py
```diff
@@ -759,11 +759,11 @@ def setup_services(self):
                         raise RuntimeError(f"Docker container setup environment var key had wrong format. Only ^[A-Z_]+[A-Z0-9_]*$ allowed: {env_key} - Maybe consider using --allow-unsafe or --skip-unsafe")
 
                     if not self._allow_unsafe and \
-                        re.search(r'^[a-zA-Z0-9_]+[a-zA-Z0-9_-]*$', env_value) is None:
+                        re.search(r'^[a-zA-Z0-9_]{1}[a-zA-Z0-9_.:/-]{0,1023}$', env_value) is None:
                         if self._skip_unsafe:
-                            print(TerminalColors.WARNING, arrows(f"Found environment var value with wrong format. Only ^[A-Z_]+[a-zA-Z0-9_]*$ allowed: {env_value} - Skipping"), TerminalColors.ENDC)
+                            print(TerminalColors.WARNING, arrows("Found environment var value with wrong format. Only ^[a-zA-Z0-9_]{1}[a-zA-Z0-9_.:/-]{0,1023}$ allowed: " + env_value + " - Skipping"), TerminalColors.ENDC)
```
ribalba, line 764: Why are you changing this from `f`-strings

runner.py
```diff
@@ -759,11 +759,11 @@ def setup_services(self):
                         raise RuntimeError(f"Docker container setup environment var key had wrong format. Only ^[A-Z_]+[A-Z0-9_]*$ allowed: {env_key} - Maybe consider using --allow-unsafe or --skip-unsafe")
 
                     if not self._allow_unsafe and \
-                        re.search(r'^[a-zA-Z0-9_]+[a-zA-Z0-9_-]*$', env_value) is None:
+                        re.search(r'^[a-zA-Z0-9_]{1}[a-zA-Z0-9_.:/-]{0,1023}$', env_value) is None:
```
ribalba, line 762: Maybe we should change this to a block list instead of allow list, so we don't need to revisit this in the near future?

runner.py
davidkopp: Because the regex contains the characters `{` and `}`

runner.py
davidkopp: I have created an alternative PR that changes it to a blacklist approach: #577

ArneTR:
I am also in favor of Didis comment and rather go to the blacklist approach. Reason being that since @davidkopp mentionend that our Regex was to strict from the start I would really like to base it on stronger assumptions why we only allow certain characters.

ribalba:
@davidkopp @ArneTR I think we can close this PR as the other solution has been merged

PR #576, opened 2023-12-01
fix for issue where tests on external PR's will fail when attempting to write to cache

dan-mm:
We have an issue where some external PR's will fail because they attempt to write to the gunicorn cache and recieve a 403 Forbidden error, failing the entire pipeline.

I think we want to keep the security restriction, but the pipeline should continue. So the buildx command itself will now always return a success. However the pipeline should fail if the buildx command fails for any other reason, so I capture the (real) exit code and build output and will still fail on anything that is not a 403 Forbidden.

Unfortunately I don't see a quick/easy way to limit this to just external PR's, and our tests failing on cache issues isn't necessarily desired behaviour anyways, so this is now the default behaviour on all test runs. This does mean if we do have actual cache permissions errors they will be harder to spot.

I could restrict this to test runs triggered by PR's only (external or otherwise), and use the basic build command on all other triggers. Thoughts?

### DIFF
--- a/.github/actions/gmt-pytest/action.yml
+++ b/.github/actions/gmt-pytest/action.yml
@@ -60,9 +60,21 @@ runs:
         password: ${{ inputs.github-token }}
 
     - name: Build docker-compose
+      id: build-docker-compose
       shell: bash
       working-directory: ${{ inputs.gmt-directory }}/docker
-      run: docker buildx bake --file test-compose.yml --file docker-compose-cache.json
+      run: |
+        { DOCKER_OUTPUT=$(docker buildx bake --file test-compose.yml --file docker-compose-cache.json 2>&1); DOCKER_EXIT_CODE=$?; } || true
+        if [ "$DOCKER_EXIT_CODE" -ne 0 ]; then
+          echo "Docker build failed with exit code $DOCKER_EXIT_CODE"
+          echo "buildx output:"
+          echo $DOCKER_OUTPUT
+          if echo "$DOCKER_OUTPUT" | grep -q "403 Forbidden"; then
+            echo "Docker build failed due to permissions issue. Continuing..."
+          else
+              exit 1
+          fi
+        fi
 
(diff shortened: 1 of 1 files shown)


ArneTR:
Are there any means of sending the repository owner a ping / message / email etc. via the API?

I see this as the best possibility to alert on soft errors?

dan-mm:
That's a good call actually - I think that'd be the best solution. I don't know off the top of my head to do it in the github actions way but I'm sure it must be possible. I'll look into it and update this PR.

dan-mm:
After looking into it, unfortunately there's no simple way to give a custom/specific notification to the user. There's no internal github mechanism in workflows nor an API endpoint to call which can achieve this. It appears the only notification github workflows are meant to do are the overall pass/fail/cancelled, without any granularity.

There are some third party actions which can do this however (such as https://github.com/dawidd6/action-send-mail), we just have to configure them with an smtp server. do we want to go this route?

ArneTR:
The fix looks nice, although very unspecific. 403 can occur for multiple reasons, even when the container could not be pulled in the first place (as the compose also has external container downloads), but I guess it does the trick as the pipeline will then fail after.

However would it not be a better fix to check if the PR is external and then not write to the cache? Can we not block that?

2. Regarding the message: I now understand that this is actually not the correct place to discuss this. We wanted the functionality for our Eco-CI to alert users that the pipeline has soft-fails.
Integrating an email into this is not a good idea I think.
@ribalba pulling you in. Do you have any good idea how we could alert users of our Eco-CI plugin for Github that for instance they have a soft fail? Examples would be: They should upgrade to a newer version. Or: A measurement did not succeed (we do not fail in this case) or similar. So anything that is considered unexpected and to be optimized, but did not actually fail the pipeline.

ArneTR:
Merging this now as we have problems of getting external PRs through the pipeline. However putting the non-cache option as new issue

PR #577, opened 2023-12-01
Refactor env var checks and tests

### LINKED ISSUE #569: Allow more characters in environment variable values

davidkopp:
Alternative to PR #574

If merged, this PR fixes #569 and #545

Refactors how the env vars are tested.
Changes the check if env var values are valid to a blacklist approach.
Adds a check if env var values are longer than 1024 characters.


ArneTR:
Thanks again for this great PR. The code quality is really good and I like your approaches of also proposing alternative routes ❤️

Also, similar like in the #574 I would like to raise the argument here to even broaden the blacklist to only highly likely problems, like for instance the character limit.

I read for instance through the discussion here: https://askubuntu.com/questions/755033/dollar-sign-in-environment-variables-value

Although my security muscle is a bit twitching when you allow users to submit a dollar sign to a process fork this is actually a safe operation. Since it never sees a shell and can only be read in the end inside of the container, where the user anyway has free roam, I would argue we should not restrict the characters here.

Apparently some applications have issues of reading shell/env-vars that have dollar signs in them. But the GMT has not and we should thus not provide a restriction artificially.

What are your thoughts? Did I maybe overlook something?

davidkopp:
I added my security tests to the following branch:
https://github.com/davidkopp/green-metrics-tool/tree/env-var-security-tests

Don't think it is worth it to add these tests to upstream.

The only warning I get with these tests is:

```log
/home/david/green-metrics-tool/tests/test_env_var_security_problems.py:119: DeprecationWarning: invalid escape sequence '\$'
    assert '\$\$PWD' in env_var_output, Tests.assertion_info('\$\$PWD in env var output', env_var_output)
```

However, it is just a warning. The env var actually gets added (`env`):
```env
TEST_DOLLAR4=\$\$PWD
```

Maybe we can consider forbidding the backslash character.

ArneTR:
Thanks for the analysis!

Forbidding the backslash character makes sense then in that case due to the Python deprecation. 

If you alter it to only block on the backslash then I think this is a candidate to merge. 

Still waiting for @ribalba reply on this early next week.

davidkopp:
The warning about the backslash character was not caused by the value of the environment variable. It was caused by a mistake in my test code 🙈
So I think it is ok to allow every character in env var values. I changed the code accordingly.

ribalba (approved):
After thinking about this for quite some time. I have come to the conclusion that this should be ok. Although I am somehow uneasy with allowing everything I can't really find out why and can't think of an exploit. If we ever set `shell=True` in the run command all hell will break loose though.

ArneTR:
❤️ @davidkopp . Merging in

PR #584, opened 2023-12-10
Fix frontend flow menu to wrap automatically

davidkopp:
If there are many flows some of them are not visible and clickable in the frontend.
[Example measurement](https://metrics.green-coding.berlin/stats.html?id=b5478c99-c8b4-4f65-a25b-99180f5ced2f):

image

This PR adds the `wrapping` attribute to the menu ([→Fomantic docs](https://fomantic-ui.com/collections/menu.html#wrapping)).

Result:

image

### DIFF
--- a/frontend/compare.html
+++ b/frontend/compare.html
@@ -187,7 +187,7 @@
                 Runtime can contain multiple flows.
                 By default all runtime flows are aggregated. Please select a separate flow if needed.
             </div>
-            <div id="runtime-sub-phases" class="ui top attached tabular menu">
+            <div id="runtime-sub-phases" class="ui top attached tabular wrapping menu">
                 <a class="active item" data-tab="[[RUNTIME]]">All Flows</a>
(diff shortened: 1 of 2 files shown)


ArneTR:
Again, David thank you very much ❤️

The attribute makes perfect sense. Can you send me a link to a run though where you have generated that data, so I can quickly look how it behaves in mobile and different browsers?

davidkopp:
Here a run with 25 flows:
https://metrics.green-coding.berlin/stats.html?id=b5478c99-c8b4-4f65-a25b-99180f5ced2f

ribalba:
Just tested it and it works. Thx @davidkopp

Isn't the prettiest way to show this if we have loads of flow steps. The question is how often does this occur? If this is a very typical case I would change it to something that looks a little better. If this is an edge case good to merge IMHO

ArneTR:
lgtm ❤️

PR #588, opened 2023-12-11
581 metric reporters checks

dan-mm:


dan-mm:
@ArneTR - also tagging you for the sample IPMI data, when you have a moment

ArneTR:
@dan-mm Here is IPMI data from one of our machines with a broken BMC:

```
gc@m3-silent:~$ sudo /usr/sbin/ipmi-dcmi --get-system-power-statistics
Current Power                        : 0 Watts
Minimum Power over sampling duration : 0 watts
Maximum Power over sampling duration : 0 watts
Average Power over sampling duration : 0 watts
Time Stamp                           : 12/15/2023 - 12:41:53
Statistics reporting time period     : 5000 milliseconds
Power Measurement                    : Active
```

metric_providers/psu/energy/ac/sdia/machine/provider.py
```diff
@@ -25,6 +26,22 @@ def get_stderr(self):
     def start_profiling(self, containers=None):
         self._has_started = True
 
+
+    #TODO: not a fan of using the full key name here. any way to avoid this?
+    # keeping original checks in read_metrics for now
+    def check_system(self):
+        config = GlobalConfig().config
+        provider_config = config['measurement']['metric-providers']['common']['psu.energy.ac.sdia.machine.provider.PsuEnergyAcSdiaMachineProvider']
```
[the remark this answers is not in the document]
ArneTR: You can extract the string information for the key from the current filename of the provider.py, just replacing / with ., and the classname you have readily available. Please use that.

dan-mm:
> @dan-mm Here is IPMI data from one of our machines with a broken BMC:
> 
> ```
> gc@m3-silent:~$ sudo /usr/sbin/ipmi-dcmi --get-system-power-statistics
> Current Power                        : 0 Watts
> Minimum Power over sampling duration : 0 watts
> Maximum Power over sampling duration : 0 watts
> Average Power over sampling duration : 0 watts
> Time Stamp                           : 12/15/2023 - 12:41:53
> Statistics reporting time period     : 5000 milliseconds
> Power Measurement                    : Active
> ```

Updated with non-zero checks on current power ^^

metric_providers/psu/energy/ac/sdia/machine/provider.py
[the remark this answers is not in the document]
ArneTR: jap, that needs to be hardcoded. But I think that is fair, as it is quite a restrictive requirement in the first place.

metric_providers/psu/energy/ac/powerspy2/machine/provider.py
```diff
@@ -12,3 +12,8 @@ def __init__(self, resolution):
             metric_provider_executable="metric-provider.py",
         )
         self._extra_switches = ['-u','mJ']
+
+    def check_system(self):
+        file_path = "/dev/rfcomm0"
```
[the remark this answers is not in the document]
dan-mm: You're right, there should be a read here. IIRC, I think the reason I didn't add it is because I had no way of checking it, as I don't have this device, so this was the only way I could verify it myself. Its straightfoward code for the read, so I'll add it

metric_providers/psu/energy/ac/mcp/machine/provider.py
```diff
@@ -12,3 +12,8 @@ def __init__(self, resolution):
             unit="mJ",
             current_dir=os.path.dirname(os.path.abspath(__file__)),
         )
+
+    def check_system(self):
+        file_path = "/dev/ttyACM0"
+        if not os.path.exists(file_path):
```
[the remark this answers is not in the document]
dan-mm: Same note as below - would make sense to read, however I don't have the device and couldn't verify it myself.

When porting over to C though, it seems it already has a pretty good check near the beginning that does a check on this device + some:

```
    fd = f511_init("/dev/ttyACM0");
    if(fd < 0) {
        intf(stderr, "Error. Connection could not be opened\n");
        return -1;
    }
```

I would just make the code exit after this if the check_system_flag is passed. The only difference between this and how we do the check in other files is return -1 vs exit(127). Is there a relevant difference between exiting these two ways? I don't think there is.

metric_providers/psu/energy/ac/ipmi/machine/provider.py
```diff
@@ -43,3 +44,19 @@ def read_metrics(self, run_id, containers=None):
...
 
         return df
+
+    def check_system(self):
+        # Run 'sensors' command and capture the output
+        ps = subprocess.run(['sudo', '/usr/sbin/ipmi-dcmi', '--get-system-power-statistics'], capture_output=True, text=True, check=False)
+        if ps.returncode != 0:
+            raise MetricProviderConfigurationError(f"\n\n{self._metric_name} provider could not be started.\nCannot run the 'sudo /usr/sbin/ipmi-dcmi --get-system-power-statistics' command.\n\nAre you running in a VM / cloud / shared hosting?\nIf so please disable the {self._metric_name} provider in the config.yml")
+
+        current_power_line = [line for line in ps.stdout.split('\n') if line.startswith('Current Power')]
```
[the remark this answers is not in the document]
dan-mm: While your implementation is definitely faster, I wouldn't go so far to say that mine is crazy slow, especially if we're comparing with regexes. This code is only run once per GMT run on a small text snippet, not 1_000_000 times.

Though in the interest of our work here, hopefully this code will be run 1_000_000 times over its "lifetime" across many machines - so ok yes in that context I can see the point. Still, I don't think running it a million times in sequence like this gives us an accurate representation...

I re-ran the benchmark (adding your string function as well), but with num_iterations = 10

```
/tmp via 🐍 v3.10.12 
❯ py3 time.py 
Regex method time: 0.000164 seconds
String method (dan) time: 0.000023 seconds
String method (arne) time: 0.000008 seconds

/tmp via 🐍 v3.10.12 
❯ py3 time.py 
Regex method time: 0.000243 seconds
String method (dan) time: 0.000032 seconds
String method (arne) time: 0.000009 seconds

/tmp via 🐍 v3.10.12 
❯ py3 time.py 
Regex method time: 0.000163 seconds
String method (dan) time: 0.000024 seconds
String method (arne) time: 0.000009 seconds

/tmp via 🐍 v3.10.12 
❯ py3 time.py 
Regex method time: 0.000172 seconds
String method (dan) time: 0.000021 seconds
String method (arne) time: 0.000008 seconds
```
Running this many times, averaging, and then scaling up the result I think would be a much more fair comparision. Its actually interesting to me that the string methods scale that much worse than regexe's when batched together, that's not the result I would have expected! I originally did this because I was curious what the "real world" time cost of one run was, I didn't expect to see this sort of flipped scaling...

Regardless, your version is definitely much more efficent, its 2-3 times as fast.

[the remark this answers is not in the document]
dan-mm: But now that I've rambled about this too, yes I agree it makes sense to put it in the script directly

metric_providers/base.py
```diff
@@ -47,8 +47,17 @@ def __init__(
...
+    # by default we expect the executable to have a -c switch to test functionality
     def check_system(self):
+        call_string = self._metric_provider_executable
+        if self._metric_provider_executable[0] != '/':
+            call_string = f"{self._current_dir}/{call_string}"
+
+        ps = subprocess.run([f"{call_string}", '-c'], capture_output=True, encoding='UTF-8', check=False)
+        if ps.returncode != 0:
+            raise MetricProviderConfigurationError(f"{self._metric_name} provider could not be started.\nError: {ps.stderr}\nAre you running in a VM / cloud / shared hosting?\nIf so please disable the {self._metric_name} provider in the config.yml")
+
```
dan-mm, line 60: This implementation doesn't quite work, as the call_string is not always the binary. For NetworkConnectionsProxyContainerProvider for example, the   callstring is "/usr/bin/tinyproxy".

Additionally, wouldn't this require all providers with a binary to include this -c flag? Is that something we want to enforce moving forward?

I suggest we move this functionality in its own function (CheckMetricProviderBinary) , and explictely call it in whichever providers its applicable to (which admittedly is most of them now)

metric_providers/psu/energy/ac/ipmi/machine/provider.py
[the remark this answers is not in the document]
dan-mm: Done  - 
as a note, because of set -euo pipefail, if ipmi is missing or the sudo /usr/sbin/ipmi-dcmi --get-system-power-statistics command fails for any reason, it will already exit and spit into stderr (which is what I see on my machine), so that's why there's no explicit check for that.

Also this check would be good to have a look at on a machine that gives back real data. I tested the check function with:

```
if [ "$check_system" = True ]; then
    test="Current Power                        : 0 Watts
Minimum Power over sampling duration : 10 watts
Maximum Power over sampling duration : 0 watts
Average Power over sampling duration : 0 watts
Time Stamp                           : 12/15/2023 - 12:41:53
Statistics reporting time period     : 5000 milliseconds
Power Measurement                    : Active"

    #first_line=$(sudo /usr/sbin/ipmi-dcmi --get-system-power-statistics | head -1)
    first_line=$(echo $test | head -1)
    if ! [[ "$first_line" =~ ^"Current Power" ]]; then
        echo "Unable to find 'Current Power' in the output of 'sudo /usr/sbin/ipmi-dcmi --get-system-power-statistics' command. Found $first_line instead">&2
        exit 1
    fi

    current_power_watts=$(echo $first_line |  awk '{print $4}')
    if [[ $current_power_watts -le 0 ]]; then
        echo "Current Power Watts is $current_power_watts, which is unexpected">&2
        exit 1
    fi
    exit 0
fi
```
and various other test strings, but that's not a standin for real data from the ipmi invocation. Shell scripting can be finicky with these things

metric_providers/base.py
dan-mm: For now, I fixed the NetworkConnectionsProxyContainerProvider by moving the simulatenous provider  check into its own function (that's still called by the base check_system function), and changed that specific provider to call the new function instead of super.check_system().

ArneTR:
if ready to look at please trigger a re-review or ping me

dan-mm:
@ArneTR - ready for review again

metric_providers/psu/energy/ac/xgboost/machine/provider.py
```diff
@@ -36,6 +37,15 @@ def get_stderr(self):
...
         self._has_started = True
 
+    #TODO: not a fan of using the full key name here. any way to avoid this?
+    # keeping original checks in read_metrics for now
+    def check_system(self):
+        config = GlobalConfig().config
+        if 'cpu.utilization.procfs.system.provider.CpuUtilizationProcfsSystemProvider' not in config['measurement']['metric-providers']['linux']:
+            raise MetricProviderConfigurationError(f"{self._metric_name} provider could not be started.\nPlease activate the CpuUtilizationProcfsSystemProvider in the config.yml\n \
+                This is required to run PsuEnergyAcSdiaMachineProvider")
+        self.check_parallel_provider_running()
```
ArneTR, line 47: I do see that you cannot call the base super function here. But what "parallel provider" are you checking here for?

There is no binary or script to execute. Actually this is even a logical bug in the code already. Neither the base check_system nor the check_parallel_provider_running may be called. Also a comment should clarify that we do that on purpose

metric_providers/psu/energy/ac/xgboost/machine/README.md
```diff
@@ -0,0 +1,3 @@
+# Documentation
+
+Please see https://docs.green-coding.berlin/docs/measuring/metric-providers/psu-energy-xgboost-system/ for details
```
ArneTR, line 3: io

metric_providers/psu/energy/ac/sdia/machine/provider.py
```diff
@@ -26,6 +27,25 @@ def get_stderr(self):
...
+
+        if not provider_config['CPUChips']:
+            raise MetricProviderConfigurationError(f"{self._metric_name} provider could not be started.\nPlease set the CPUChips config option for PsuEnergyAcSdiaMachineProvider in the config.yml")
+        if not provider_config['TDP']:
+            raise MetricProviderConfigurationError(f"{self._metric_name} provider could not be started.\nPlease set the TDP config option for PsuEnergyAcSdiaMachineProvider in the config.yml")
+
+        if 'cpu.utilization.procfs.system.provider.CpuUtilizationProcfsSystemProvider' not in config['measurement']['metric-providers']['linux']:
+            raise MetricProviderConfigurationError(f"{self._metric_name} provider could not be started.\nPlease activate the CpuUtilizationProcfsSystemProvider in the config.yml\n \
+                This is required to run PsuEnergyAcSdiaMachineProvider")
+        self.check_parallel_provider_running()
```
ArneTR, line 47: I do see that you cannot call the base super function here. But what "parallel provider" are you checking here for?

There is no binary or script to execute. Actually this is even a logical bug in the code already. Neither the base check_system nor the check_parallel_provider_running may be called. Also a comment should clarify that we do that on purpose

metric_providers/psu/energy/ac/sdia/machine/README.md
```diff
@@ -0,0 +1,4 @@
+# Documentation
+
+Please see https://docs.green-coding.berlin/docs/measuring/metric-providers/psu-energy-sdia-system/ for details
```
ArneTR, line 3: io

metric_providers/psu/energy/ac/powerspy2/machine/provider.py
```diff
@@ -13,3 +13,16 @@ def __init__(self, resolution, skip_check=False):
...
+        if os.path.exists(file_path):
+            try:
+                with open(file_path, 'r', encoding='utf-8') as file:
+                    file.read()
+            except PermissionError as exc:
+                raise MetricProviderConfigurationError(f"{self._metric_name} provider could not be started.\nCannot read device at {file_path}.\n\nAre you running in a VM / cloud / shared hosting?\nIf so please disable the {self._metric_name} provider in the config.yml") from exc
+        else:
+            raise MetricProviderConfigurationError(f"{self._metric_name} provider could not be started.\nCould not find device at {file_path}.\n\nAre you running in a VM / cloud / shared hosting? \nIf so please disable the {self._metric_name} provider in the config.yml")
+
+        self.check_parallel_provider_running()
```
ArneTR, line 28: why not the base function?

metric_providers/psu/energy/ac/powerspy2/machine/README.md
```diff
@@ -6,6 +6,9 @@ under https://invent.kde.org/vkrause/powerspy2-tools. Some influence from https:
 The protocol documentation can be found under:
 https://www.alciom.com/wp-content/uploads/2018/04/LG0838-003-powerspy-protocol-specification-1G.pdf
 
+Full documentation of this provider can be found on our main docs page:
```
ArneTR, line 9: io

metric_providers/psu/energy/ac/mcp/machine/source.c
```diff
@@ -241,11 +244,16 @@ int main(int argc, char **argv) {
     setvbuf(stdout, NULL, _IONBF, 0);
 
     fd = f511_init("/dev/ttyACM0");
+
     if(fd < 0) {
         fprintf(stderr, "Error. Connection could not be opened\n");
         return -1;
     }
 
+    if(check_system_flag && fd >= 0) {
```
ArneTR, line 253: This check for fd >= 0 is redundant and I feel rather confusing. Make it rather an if / elseif clause with the one above.

metric_providers/psu/energy/ac/mcp/machine/README.md
```diff
@@ -1,3 +1,3 @@
-# Information
+# Documentation
 
-TODO
\ No newline at end of file
+Please see https://docs.green-coding.berlin/docs/measuring/metric-providers/psu-energy-ac-mcp-machine/ for details
```
ArneTR, line 3: io

metric_providers/psu/energy/ac/ipmi/machine/ipmi-get-machine-energy-stat.sh
```diff
@@ -2,15 +2,35 @@
 set -euo pipefail
 
 i=''
-
-while getopts "i:" o; do
+check_system=False
```
ArneTR, line 5: Captilalied in Bash? isn't it false ?

metric_providers/psu/energy/ac/ipmi/machine/README.md
```diff
@@ -1,5 +1,7 @@
 # Information
 
+See https://docs.green-coding.berlin/docs/measuring/metric-providers/psu-energy-ac-ipmi-machine/ for details.
```
ArneTR, line 3: io

metric_providers/network/io/cgroup/container/source.c
```diff
@@ -152,9 +152,47 @@ static int parse_containers(container_t** containers, char* containers_string, i
     return length;
 }
 
+static int check_system(int rootless_mode) {
+    char file_path_cgroup_procs[BUFSIZ];
+    char file_path_proc_net_dev[BUFSIZ];
+    int found_error = 0;
+
+    if(rootless_mode) {
+        sprintf(file_path_cgroup_procs, "/sys/fs/cgroup/user.slice/cgroup.procs");
```
ArneTR, line 161: const

metric_providers/memory/energy/rapl/msr/component/README.md
```diff
@@ -18,4 +18,4 @@ Example:
 
 # Documentation
 
-For details and output format please look at https://docs.green-coding.org/docs/measuring/metric-providers/memory-energy-rapl-msr-component/
\ No newline at end of file
+For details and output format please look at https://docs.green-coding.berlin/docs/measuring/metric-providers/memory-energy-rapl-msr-system/
```
ArneTR, line 21: io

metric_providers/base.py
ArneTR: I am not a fan of the additional function that technically belongs to the check_system logically.

The reason for having this function seems that it can be overriden with a different call to the binary for checking.

If that is the reason, why don't we make rather a variable that specifies how the default call is as a python list ala `binary_check=["metric_provider_binary", "-c"]`

This way it is way cleaner and there is no risk of not calling the functions correctly. Flow should be that you cann the super.check_system() and then add logic on top, or override the function alltogether.
Having a mix of not calling the super function, adding own logic and then calling another super / object method (currently check_parallel...) seems hard to follow

ArneTR (changes requested):

metric_providers/network/connections/proxy/container/provider.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, *, host_ip=None, skip_check=False):
         self._host_ip = host_ip
 
     def check_system(self):
-        super().check_system()
+        self.check_parallel_provider_running()
```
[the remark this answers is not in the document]
dan-mm: you mean super().check_parallel_provider_running() ?

If so, I can - but as far as I know, these two lines are functionally identical. I saw a mixture of both and just standardized it to self.check_parallel_provider() because I didn't really see a difference.

Though thinking about it "out loud" now - super() makes it explicit to the reader that the function being called is in the super class, so stylistically I can see why that'd be preferred. Is there any other reason I'm missing?

metric_providers/psu/energy/ac/ipmi/machine/ipmi-get-machine-energy-stat.sh
dan-mm: that's true... it was working before, but in a string comparision way. Rewrote both of the shell scripts which use this to use bash's true/false checks

metric_providers/network/connections/proxy/container/provider.py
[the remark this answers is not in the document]
ArneTR: no, I mean super.check_system()

Why can you not call the complete super function? Why only the check_parallel_provider_running?

metric_providers/base.py
ArneTR: I unresolved this as the question is not answered and I saw no commit on chaning this yet

metric_providers/network/connections/proxy/container/provider.py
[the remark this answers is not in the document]
dan-mm: I think maybe this is an issue in how Github PR's show up - if you saw the whole class it would be clear I think...

This function isn't just calling check_parallel_provider - I am overwriting the check_system with its own logic (in this provider's case, checking the output of the "tinyproxy -v" command. This is different from the base.py's check_system function, which tries to call the binary with -c

The super().check_parallel_provider_running is an additional check on top of that to see if there are multiple instances of this metric provider running. This is shared amongst almost all the providers, which is why its a function from base.by which I call from here (in addition to being part of the default implementation of check_system)

metric_providers/base.py
dan-mm: For the record on this - I had marked this as resolved after my last comment, before you had asked any additional questions - I thought it was ok as I was the one who opened this issue, answered my own question, and then resolved it.

To answer your question - first, the check_parallel_provider is part of and called by the check_system. It was split out for the cases where we want to do a different kind of system check, but still also want to check parallel providers.

I did think of adding adding parameters for the check_system with what the binary call would be - but this doesn't cover all of our bases. I can think of at least one provider where the check requires overwriting this function, but we still want to check the parallel providers. 

I see that the flow of overriding the function, but still calling another part of the function can be confusing though. What we could do instead is remove it from check_system, but instead earlier in the base.py do this:

```
        if not self._skip_check:
            self.check_system()
            self.check_parallel_provider_running()
```

that way a metric provider that wants to have a different kind of system check, but still keep the parallel provider check wouldn't have to remember to call the super().check_parallel_provider_running(). I don't see this as really resolving anything though, as then the providers which *don't* need to run the check_parallel_provider_running (xgboost/sdia, as you pointed out before), then need to remember to pass these as well, so its just moving the mental load over to a different place. Though - in the case its forgotten, its better to accidentally have an extra check than accidentally have one less check.

metric_providers/gpu/energy/nvidia/smi/component/provider.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, resolution, skip_check=False):
 
 
     def check_system(self):
-        super().check_system()
+        self.check_parallel_provider_running()
```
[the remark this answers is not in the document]
dan-mm: Because I'm overwriting the check_system with a seperate check - (which nvidia-smi etc...)

As that check is a one-liner, we can rewrite the base.py system_check to take a variable, but that won't cover all of our bases anyways, so I don't see the benefit

metric_providers/psu/energy/ac/powerspy2/machine/provider.py
dan-mm: Because I'm overwriting the check_system with a separate check - (checking the /dev/rfcomm0 filepath)

I'm not sure here if I'm missing something obvious, or if this file was just confusing the read fully on the github pr?

dan-mm:
@ArneTR - made the requested changes, ready for review/merge

ArneTR:
lgtm!

PR #593, opened 2023-12-12
Use depends_on for container startup order (refactored)

### LINKED ISSUE #567: Compose flag `depends_on` gets ignored

davidkopp:
Alternative to PR #568 with some refactorings.
If merged, resolves #567


ArneTR:
great rework @davidkopp !

lgtm

PR #598, opened 2023-12-15
Make loading of detailed measurements optional

ArneTR:
Loading of a complex detailed measurements page can take enormous amounts of time du to the large amount of data transferred and the long query in the database.

Since this value cannot really be brought down, it is already compressed and DB indices are set, this PR shall enable the option to make the data transfer optional.

Especially beginner users do not need the timeline charts and rather use the phase_stats.

See the timing from the developer console attached. The PR brings this down to 10%

Screenshot 2023-12-15 at 10 33 16 AM

### DIFF
--- a/frontend/js/stats.js
+++ b/frontend/js/stats.js
@@ -404,7 +404,8 @@ $(document).ready( (e) => {
         renderBadges(url_params);
 
         fillRunData(run_data);
-        console.log(phase_stats_data);
+
+        displayNetworkIntercepts(network_data);
 
         if(phase_stats_data != null) {
             displayComparisonMetrics(phase_stats_data)
@@ -413,8 +414,6 @@ $(document).ready( (e) => {
         if (localStorage.getItem('fetch_detailed_measurements') === 'true') {
             getDetailedMeasurements(url_params);
         }
-
-        //displayNetworkIntercepts(network_data);
 
         // after all charts instances have been placed
         // the flexboxes might have rearranged. We need to trigger resize
--- a/frontend/js/stats.js
+++ b/frontend/js/stats.js
@@ -282,39 +282,32 @@ const displayTimelineCharts = (metrics, notes) => {
 
 
 
-async function makeAPICalls(url_params) {
+async function makeBaseAPICalls(url_params) {
+
+    let run_data = null;
+    let phase_stats_data = null;
+    let network_data = null;
 
     try {
-        var run_data = await makeAPICall('/v1/run/' + url_params.get('id'))
+        run_data = await makeAPICall('/v1/run/' + url_params.get('id'))
     } catch (err) {
         showNotification('Could not get run data from API', err);
     }
 
     try {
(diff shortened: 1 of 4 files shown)


ArneTR:
Forgot to mention: This behaviour can also be changed in settings. By default loading data is off.

In every run this button will be shown:

Screenshot 2023-12-15 at 10 48 39 AM

frontend/js/stats.js
```diff
@@ -355,36 +348,73 @@ const getURLParams = () => {
...
 
-        let [run_data, measurements_data, notes_data, phase_stats_data, network_data] = await makeAPICalls(url_params);
+        let [run_data, phase_stats_data, network_data] = await makeBaseAPICalls(url_params);
 
         if (run_data == undefined) return;
 
         renderBadges(url_params);
 
         fillRunData(run_data);
+        console.log(phase_stats_data);
```
ribalba, line 407: Should the log stay in there?

frontend/js/stats.js
```diff
@@ -355,36 +348,73 @@ const getURLParams = () => {
...
-        if (measurements_data == undefined) return;
-        const metrics = getTimelineMetrics(measurements_data);
-
-        if (notes_data == undefined) return;
-        displayTimelineCharts(metrics, notes_data);
+        if (localStorage.getItem('fetch_detailed_measurements') === 'true') {
+            getDetailedMeasurements(url_params);
+        }
 
-        displayNetworkIntercepts(network_data);
```
ribalba, line 387: And this?

ribalba:
I really like the idea. But wouldn't it be easier just to load the data if someone clicks on the "Click here for detailed metrics ..." button. You could show a little loading spinner. Like this people don't really need to know the details and it would just be in the flow of looking at things?

Screenshot 2023-12-15 at 10 51 27

ArneTR:
Thanks for the spot on the outcommented code. A good reminder to always to a git diff.

2. The button "Click here for detailed metrics" contains a table and in comparison mode the comparison charts.

I see this as different data, although the naming is pretty close atm.

I am personally not a fan of clicking on a button and then data appears in two different places, so I would not like to integrate it that way.
Especially since the behaviour will be different between stats.html and compare.html. In the latter no detailed time series will appear at the bottom.

I would keep it very local. If you click on a button that talks about data, it should appear right below it.

Having said that: Tackling the loading issue for the compare.html is a separate task. I have planned that out already, but since it is not tied to fetch events, but phase calculations internally this is a bigger one

ribalba:
Sorry I wasn't clear. I think it is confusing to have the same button twice. I wouldn't know what the difference is between the two and it is confusing that they are pretty much named the same. I would have one button for detailed metrics and not one which opens a table and one which loads graphs and then a setting that only sets the one. We would have to restructure the way the page is laid out though in this case.

ArneTR:
as said, this is iteratively. The latter will come as a seprate functionality.

I agree on the naming though. Have pushed to main and call it "Detailed Time Series"

PR #599, opened 2023-12-16
Status has now a waiting time

ArneTR:
This PR introduces a new table on the "status" page with current job status of the machines as well as an approximate waiting time.

Waiting time is calculated as `JOB_AMOUNT * 15 Minutes + JOB_AMOUNT * COOLDOWN_TIME`

All config data is updated through the `client.py` script on cluster announce
Screenshot 2023-12-16 at 10 04 52 PM


ArneTR:
@ribalba Any additional feature you would like to see on the status page?

frontend/js/status.js
```diff
@@ -16,5 +16,28 @@ $(document).ready(function () {
...
+                { data: 2, title: 'Available'},
+                { data: 3, title: 'Status'},
+                { data: 4, title: 'Status date'},
+                { data: 5, title: 'Cooldown after Job', render: function(el) {
+                    return `${el/60} Minutes`
+                }},
+                { data: 6, title: 'Waiting Jobs'},
+                { data: 5, title: 'Estimated waiting time', render: function(el, type, row) {
+                    // 900 Seconds is current average job time. WE add this to the amount of waiting time
+                    return `${( (900+row[5]) * row[6]) / 60} Minutes`
```
ribalba, line 34: Is there a way to make the 900 calculated dynamically. As this might change depending on the workloads. Maybe take the AVG time of the last 10 runs? Couldn't this be done with an SQL subquery in the SELECT?

ribalba (approved):
Otherwise looks great

frontend/js/status.js
ArneTR: good one. will do

ArneTR:
Sometimes the time is unknown as the machine has not yet come up to info about it's cooldown time or we have no runs in the history.

In both cases I just issue "Unknown".
Screenshot 2023-12-22 at 4 41 29 PM

Awaiting tests to run, then will put it online

PR #600, opened 2023-12-17
Validate.py script to check stddev of energy measurements for cluster

ArneTR:
This script `tools/validate.py` can run a validation workload and compare the STDDEV over a defined window.

Atm we are doing this manually here: https://metrics.green-coding.berlin/timeline.html?uri=https://github.com/green-coding-berlin/measurement-control-workload&filename=usage_scenario.yml&branch=&machine_id=7

Whenever the workload STDDEV goes out of the thresshold the script throws an error and sends an error message.

Optionally, when everything is fine, the script will also debug-report this via email.

This script is to be used in unattended cluster setups. We have it as a systemd oneoff service to be run when the machine powers on.

```
[Unit]
Description=The Green Metrics Measurement Control Service
After=network.target

[Service]
Type=oneoff
User=gc
Group=gc
WorkingDirectory=/home/gc/green-metrics-tool/
ExecStart=/home/gc/green-metrics-tool/venv/bin/python3 /home/gc/green-metrics-tool/tools/validate.py
StandardOutput=append:/var/log/green-metrics-measurement-control-service.log

[Install]
WantedBy=multi-user.target
```

Then the cluster script must depend on this service and will only start if this successfully returns


tools/validate.py
```diff
@@ -0,0 +1,138 @@
...
+        warning = False
+        info_string_acc = ''
+        for el in data:
+            info_string = f"{el[0]} {el[1]}: {el[4]} +/- {el[5]} {el[6]*100} %"
+            print(info_string)
+            info_string_acc = f"{info_string_acc}\n{info_string}"
+            if el[6] > config['machine']['control_workload']['threshhold']:
+                print(TerminalColors.FAIL, 'Warning. Threshhold exceeded!', TerminalColors.ENDC)
+                warning = True
+        if warning:
```
ribalba, line 121: Why are you working with a warning here if you overwrite the `info_string_acc`? Couldn't you just raise right away? Or I would append to the`info_string_ac` so you see where the threshold was exceeded

ribalba (changes requested):

ribalba:
Currently this is only run on boot. Don't we want to run this every day to see if something has changed?

ArneTR:
I think it comes down to a measurement philiosophy.

The idea is that once the cluster is up and no updates are installed it shall be in a stable state.

Everytime you install an update the cluster shall be restarted and thus re-validated.

If we run a daily workload one could also make the argument: Why not after every workload. 

Effectively this is the same question as with backups. The most safe would be to backup after every new file to the file system. A typical routine is however every month, week, day or similar.

So: What is the most sane for us? 

Having given this lenghty explanation just to order my thoughts I think I come to the same conclusion as you. Every day + After every reboot. Will add as fix

tools/validate.py
ArneTR: Maybe you missed the line. I am appending: info_string_acc = f"{info_string_acc}\n{info_string}"

ribalba: Totally missed that. Isn't that really inefficient? 
```
didi@marge /tmp % cat a.py
x = ''
for i in range(200_000):
    x = f"{x}{i}"
print(len(x))
didi@marge /tmp % cat b.py
x = []
for i in range(200_000):
    x.append(str(i))
print(len(''.join(x)))
didi@marge /tmp % time python3 a.py
1088890
python3 a.py  1.92s user 0.01s system 99% cpu 1.946 total
didi@marge /tmp % time python3 b.py
1088890
python3 b.py  0.05s user 0.02s system 82% cpu 0.083 total
```

ArneTR: wow, I am convinced.

I did this because I saw it as recommendation in https://wiki.python.org/moin/PythonSpeed/PerformanceTips#String_Concatenation . The first one is actally your recommendation, but the second one is an inline replacement. Somehow I thought this is similar.

=> Changed

ArneTR:
@ribalba All changes incorporated. I however stopped the latest run, as it was just a typo. Please re-review

ArneTR:

ribalba (approved):

PR #601, opened 2023-12-17
Run ID is now accessible even after fail and thus can be sent via ema…

ArneTR:
This PR fixes the confusing email behaviour that sometimes this would be show:

```
Unfortunately, your Green Metrics analysis has run into some issues and could not be completed.

Name: Nextcloud Docs (mod_php-MariaDB-Firefox)
Run Id: None
Machine: Fujitsu Esprimo P956 [NOP Linux] (Blue Angel compatible)
Link: https://metrics.green-coding.berlin/stats.html?id=None
```

Since the run_id could not be seen when a run was failed it would show up as None in the email.

This lead to the further complication that some email programs want to generate a link preview and access this URL. This lead to further error emails for 404 access and so on ...

Now run_id is always transported and if really not present then no link is shown


ribalba (approved):
lgtm. Not a big fan of adding more global vars but as we use this anyway this seems the easiest solution

ArneTR:
> lgtm. Not a big fan of adding more global vars but as we use this anyway this seems the easiest solution

global var? What do you mean?

ArneTR:
> > lgtm. Not a big fan of adding more global vars but as we use this anyway this seems the easiest solution
> 
> global var? What do you mean?

@ribalba

ribalba:
I'll explain on the phone. Nothing that stops the merge.

PR #602, opened 2023-12-17
Branch and filename are now always not null

ArneTR:
This PR removes the possibility of having NULL values for branch or filename in the database.

When operating with these values we always need to set defaults and assume *main*  / *master*, which is superflous.

The information can be readily provided when providing the directory and set to the git default and the GMT default


ribalba (approved):
great idea. Makes the code a lot cleaner

PR #613, opened 2023-12-21
Fix check if stderr is empty

davidkopp:
Commit https://github.com/green-coding-berlin/green-metrics-tool/commit/270463c3663b7f5689209f4d934d596bace45d70 by @ribalba lead to the following warnings:

```log
/home/david/green-metrics-tool/runner.py:964: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if stderr_read is not None and str(stderr_read) is not '':
/home/david/green-metrics-tool/runner.py:1098: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if stderr_read is not None and str(stderr_read) is not '':
```

This PR fixes it.

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -961,7 +961,7 @@ class Runner:
 
             stderr_read = metric_provider.get_stderr()
             print(f"Stderr check on {metric_provider.__class__.__name__}")
-            if stderr_read is not None and str(stderr_read) is not '':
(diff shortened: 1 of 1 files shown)


ribalba:
Great thank you. I accidentally pushed to the wrong branch while on a train and then the internet was gone to fix it. Fix will be in another pr.

PR #616, opened 2023-12-22
Tests run_until must be guard-claused with cleanup routine

ArneTR:
The tests functions use the function `run_until`

This does no automatic cleanup. We had a regression that the tests were not running anymore because the call was not guard claused.


ArneTR:
@ribalba This fixes the NetworkConnectionsProxyContainerProvider not empty stderr we are seeing. Assuming the tests will run ...

PR #619, opened 2023-12-22
System check providers running

ArneTR:
Metric providers can be left behind on the system in case of an unclean shutdown.

This PR introduces a system check to look for running providers in different GMT instance and also adds a test case to test the guard clause to trigger.

### DIFF
--- a/metric_providers/base.py
+++ b/metric_providers/base.py
@@ -47,7 +47,17 @@ class BaseMetricProvider:
 
     # this is the default function that will be overridden in the children
     def check_system(self):
-        pass
+        cmd = ['pgrep', '-f', self._metric_provider_executable]
+        result = subprocess.run(cmd,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE,
+                            check=False, encoding='UTF-8')
+        # returncode 1 sadly indicates program failure or no found process, so we need to go deeper
+        if result.returncode not in [0, 1]:
+            raise subprocess.CalledProcessError(result.stderr, cmd)
+        if result.stdout != '':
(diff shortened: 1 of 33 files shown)


metric_providers/base.py
```diff
@@ -47,7 +47,17 @@ def __init__(
 
     # this is the default function that will be overridden in the children
     def check_system(self):
-        pass
+        cmd = ['pgrep', '-f', self._metric_provider_executable]
+        result = subprocess.run(cmd,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE,
+                            check=False, encoding='UTF-8')
+        # returncode 1 sadly indicates program failure or no found process, so we need to go deeper
```
ribalba, line 55: The comment here could be a little clearer. 
```
EXIT STATUS
     The pgrep and pkill utilities return one of the following values upon exit:

     0       One or more processes were matched.

     1       No processes were matched.

     2       Invalid options were specified on the command line.

     3       An internal error occurred.
```
The code looks ok though

metric_providers/base.py
```diff
@@ -47,7 +47,17 @@ def __init__(
...
+        if result.returncode not in [0, 1]:
+            raise subprocess.CalledProcessError(result.stderr, cmd)
+        if result.stdout != '':
```
ribalba, line 58: Couldn't you work with return codes here? If you are already looking at the code beforehand? The Empty string check is also ok though IMHO

ribalba (approved):

metric_providers/base.py
ArneTR: I restructured it to make it clearer when reading the code

metric_providers/base.py
ArneTR: see comment above. The restructuring makes the comment obsolte I think

PR #620, opened 2023-12-22
Euo pipefail

ArneTR:
This PR adds bash guards as `set -euo pipefail`

Just to be safe:
@dan-mm Is there a problem adding these to the test-containers scripts that I might be missing?
@ribalba Is there a problem adding these to the cluster cleanup bash script that I might be missing?


ribalba:
I can't think of any.

PR #623, opened 2023-12-23
Disable tinyproxy systemd service

ArneTR:
Tinyproxy was started as a service. We do not want that on our machines.


install_linux.sh
```diff
@@ -117,9 +117,13 @@ git submodule update --init
 print_message "Installing needed binaries for building ..."
 if lsb_release -is | grep -q "Fedora"; then
     sudo dnf -y install lm_sensors lm_sensors-devel glib2 glib2-devel tinyproxy lshw
+    sudo systemctl stop tinyproxy
```
ribalba, line 120: @ArneTR why do you issue the same command in Fedora and else. Couldn't you only have it once outside of the if?

ArneTR: true, this could be simplified. I changed it on main now.

PR #624, opened 2023-12-23
Stderr is now by default UTF-8

ArneTR:


ArneTR:

metric_providers/base.py
```diff
@@ -51,8 +51,18 @@ def check_system(self):
...
+    # However this function ALWAYS waits for the process to terminate and it does not allow reading from processes
+    # in chunks while they are running. Thus we we cannot set encoding='UTF-8' in Popen and must decode here.
     def get_stderr(self):
-        return self._ps.stderr.read()
+        stderr_read = ''
+        if self._ps.stderr is not None:
+            stderr_read = self._ps.stderr.read()
+            if isinstance(stderr_read, bytes):
+                stderr_read = stderr_read.decode('utf-8')
+        return stderr_read
```
ribalba, line 65: If we always return the empty string we can remove quite a few checks for stderr being None

ribalba (approved):
Looks good. I would also remove all the checks for `stderr is None` in this PR or we can also do it while refactoring the code later one

PR #625, opened 2023-12-23
Refactored kill/killpg mechanism to be unified and actually fail on n…

ArneTR:
…ot process not found

### DIFF
--- a/lib/process_helpers.py
+++ b/lib/process_helpers.py
@@ -2,7 +2,8 @@ import signal
 import os
 import subprocess
 
-def kill_pg(ps, pgid, cmd):
+def kill_pg(ps, cmd):
+    pgid = os.getpgid(ps.pid)
     print(f"Trying to kill {cmd} with PGID: {pgid}")
 
     os.killpg(pgid, signal.SIGTERM)
@@ -13,7 +14,8 @@ def kill_pg(ps, pgid, cmd):
         os.killpg(pgid, signal.SIGKILL)
         raise RuntimeError(f"Killed the process {cmd} with SIGKILL. This could lead to corrupted data!") from exc
 
-def kill_ps(ps, pid, cmd):
+def kill_ps(ps, cmd):
+    pid = ps.pid
     print(f"Trying to kill {cmd} with PID: {pid}")
 
     os.kill(pid, signal.SIGTERM)
--- a/lib/process_helpers.py
+++ b/lib/process_helpers.py
@@ -2,36 +2,26 @@ import signal
 import os
 import subprocess
 
-def kill_ps(ps_to_kill):
-    print('Killing processes')
-    for ps_info in ps_to_kill:
-        pid = ps_info['ps'].pid
-        print(f"Trying to kill {ps_info['cmd']} with PID: {pid}")
-        try:
-            if ps_info['ps_group'] is True:
-                try:
-                    ps_group_id = os.getpgid(pid)
-                    print(f" with process group {ps_group_id}")
-                    os.killpg(os.getpgid(pid), signal.SIGTERM)
-                    try:
-                        ps_info['ps'].wait(timeout=5)
-                    except subprocess.TimeoutExpired:
-                        # If the process hasn't gracefully exited after 5 seconds we kill it
-                        os.killpg(os.getpgid(pid), signal.SIGKILL)
-                except ProcessLookupError:
-                    # process may be not have been in a process group
-                    print(f"Could not find process-group for {pid}")
-            # always, just in case the calling process (typically the shell) did not die
-            os.kill(pid, signal.SIGTERM)
-            try:
-                ps_info['ps'].wait(timeout=5)
-            except subprocess.TimeoutExpired:
-                # If the process hasn't gracefully exited after 5 seconds we kill it
-                os.kill(pid, signal.SIGKILL)
+def kill_pg(ps, pgid, cmd):
+    print(f"Trying to kill {cmd} with PGID: {pgid}")
 
-        except ProcessLookupError:
-            # process may already have ended or been killed in the process group
-            print(f"Could not find process {pid}")
+    os.killpg(pgid, signal.SIGTERM)
+    try:
+        ps.wait(timeout=10)
+    except subprocess.TimeoutExpired as exc:
+        # If the process hasn't gracefully exited after 5 seconds we kill it
+        os.killpg(pgid, signal.SIGKILL)
+        raise RuntimeError(f"Killed the process {cmd} with SIGKILL. This could lead to corrupted data!") from exc
+
+def kill_ps(ps, pid, cmd):
+    print(f"Trying to kill {cmd} with PID: {pid}")
+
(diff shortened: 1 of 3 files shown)


ArneTR:

runner.py
```diff
@@ -1099,7 +1099,11 @@ def stop_metric_providers(self):
             if stderr_read is not None and str(stderr_read):
                 errors.append(f"Stderr on {metric_provider.__class__.__name__} was NOT empty: {stderr_read}")
 
-            metric_provider.stop_profiling()
+            # pylint: disable=broad-exception-caught
+            try:
+                metric_provider.stop_profiling()
+            except Exception as exc:
```
ribalba, line 1105: Why are you not catching more specific here? Couldn't you catch `ProcessLookupError` and `RuntimeError`?

lib/process_helpers.py
```diff
@@ -2,36 +2,26 @@
...
-            print(f"Could not find process {pid}")
+def kill_ps(ps, pid, cmd):
+    print(f"Trying to kill {cmd} with PID: {pid}")
+
+    os.kill(pid, signal.SIGTERM)
+    try:
+        ps.wait(timeout=10)
+    except subprocess.TimeoutExpired as exc:
+        os.kill(pid, signal.SIGKILL)
+        raise RuntimeError(f"Killed the process {cmd} with SIGKILL. This could lead to corrupted data!") from exc
```
ribalba, line 24: Couldn't you merge most of the code in `kill_pg` and `kill_ps`. The only thing that is really different is the `os.kill..`

runner.py
```diff
@@ -1120,7 +1124,17 @@ def stop_metric_providers(self):
 
     def read_and_cleanup_processes(self):
         print(TerminalColors.HEADER, '\nReading process stdout/stderr (if selected) and cleaning them up', TerminalColors.ENDC)
-        process_helpers.kill_ps(self.__ps_to_kill)
+
+        ps_errors = []
+        for ps in self.__ps_to_kill:
+            try:
+                process_helpers.kill_ps(ps['ps'], ps['ps'].pid, ps['cmd'])
+            except ProcessLookupError as exc:
```
ribalba, line 1132: Can't this also be a RunTimeError?

runner.py
```diff
@@ -1236,7 +1250,12 @@ def cleanup(self):
 
         print('Stopping metric providers')
         for metric_provider in self.__metric_providers:
-            metric_provider.stop_profiling()
+            # pylint: disable=broad-exception-caught
+            try:
+                metric_provider.stop_profiling()
+            except Exception as exc:
```
ribalba, line 1256: Isn't a more specific catch possible?

runner.py
```diff
@@ -1253,7 +1272,16 @@ def cleanup(self):
 
         self.remove_docker_images()
 
-        process_helpers.kill_ps(self.__ps_to_kill)
+        ps_errors = []
+        for ps in self.__ps_to_kill:
+            try:
+                process_helpers.kill_ps(ps['ps'], ps['ps'].pid, ps['cmd'])
+            except ProcessLookupError as exc:
```
ribalba, line 1279: see above

ribalba (commented):
A few little things otherwise good to go

metric_providers/base.py
```diff
@@ -146,20 +145,6 @@ def start_profiling(self, containers=None):
...
-            except subprocess.TimeoutExpired:
-                # If the process hasn't gracefully exited after 5 seconds we kill it
-                os.killpg(ps_group_id, signal.SIGKILL)
-                print("Killed the process with SIGKILL. This could lead to corrupted metric log files!")
-
-        except ProcessLookupError:
-            print(f"Could not find process-group for {self._ps.pid}",
-                    file=sys.stderr)  # process/-group may have already closed
 
+        process_helpers.kill_pg(self._ps, os.getpgid(self._ps.pid), self._metric_provider_executable)
```
ribalba, line 149: couldn't you do the `getpgid` in the `kill_pg`?

lib/process_helpers.py
```diff
@@ -2,36 +2,26 @@
...
+        ps.wait(timeout=10)
+    except subprocess.TimeoutExpired as exc:
+        # If the process hasn't gracefully exited after 5 seconds we kill it
+        os.killpg(pgid, signal.SIGKILL)
+        raise RuntimeError(f"Killed the process {cmd} with SIGKILL. This could lead to corrupted data!") from exc
 
-        except ProcessLookupError:
-            # process may already have ended or been killed in the process group
-            print(f"Could not find process {pid}")
+def kill_ps(ps, pid, cmd):
```
ribalba, line 16: Do you need ps and pid? Can't you get the pid from the ps?

lib/process_helpers.py
```diff
@@ -2,36 +2,26 @@
...
-                    # process may be not have been in a process group
-                    print(f"Could not find process-group for {pid}")
-            # always, just in case the calling process (typically the shell) did not die
-            os.kill(pid, signal.SIGTERM)
-            try:
-                ps_info['ps'].wait(timeout=5)
-            except subprocess.TimeoutExpired:
-                # If the process hasn't gracefully exited after 5 seconds we kill it
-                os.kill(pid, signal.SIGKILL)
+def kill_pg(ps, pgid, cmd):
```
ribalba, line 5: Can't you get the pgid from the ps. See comment below on putting the getpid in here

ArneTR: true. will changea

ArneTR: true. will change

lib/process_helpers.py
ArneTR: yes, you can. But for a ten line function that would need two switch statements to merge I thought it is cleaner to have two separate functions in the end

metric_providers/base.py
ArneTR: resolved

runner.py
ArneTR: the reason for my choice here is that this is a sensitive function path and in any case you want to call the stop_profiling on every of the providers and not fail early due to an uncaught exception.

runner.py
ArneTR: The idea here is to let the RuntimeError bubble up and fail the path. Having a process not available anymore is very often expected and metrics are fine. But having to SigKill is an error and metrics are most likely corrupted.

To let the user see this I choose for a harder error. Clairfied this with a comment

runner.py
ArneTR: Same rationale here: Sensitive function path and we want to trigger in any case.

runner.py
ArneTR: resolved in previous commend

ArneTR:
@ribalba Thanks for the many comments and relevant questioning of the implementation. I have changed the relevant fields or put a comment where I think the implementation was justified. Will merge in if tests pass as only stylistic implementation questions reman.

If you have further comments please put in this PR. I will fix in main

PR #629, opened 2023-12-26
Phase Stats Providers

ArneTR:
This PR refactors the phase_stats to use a plug-in mechanism.

Concept:

- A new class is created in config.yml: Phase Stats Provider
- This can then define an input that it swallows
- and an output that it emits.
- The provider will then take the input and aggregate it and create an average value
- It needs a name as output
- And the provider can also multiply by a fixed value

- The system gets one or more inputs
    - It can then aggregate these itself, or spit out something per value again

~~Translated with DeepL.com (free version)~~

Note: This PR is a draft and might not come into action in this form. It is public only to track and save progress on Github

### DIFF
--- a/phase_stat_providers/carbon/intensity/entsoe/provider.py
+++ b/phase_stat_providers/carbon/intensity/entsoe/provider.py
@@ -1,0 +1,43 @@
+import os
+
+# from phase_stat_providers.base import BasePhaseStatProvider
+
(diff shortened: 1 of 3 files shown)


ribalba:
```
Translated with DeepL.com (free version)
```
?

ArneTR:
> ```
> Translated with DeepL.com (free version)
> ```
> 
> ?

If you hate deepl.com let me know :)

PR #630, opened 2023-12-26
Monitor mode

ArneTR:
This PR adds the monitor mode which was originally brought in as idea from @davidkopp  in this PR https://github.com/green-coding-berlin/green-metrics-tool/pull/556

The monitor mode sets the GMT in an endless loop where the providers continue to collect data.

Main change is that providers that do not collect system metrics but rather transient data like container metrics need an auto discovery mode.

This has in this PR draft so far been implemented for the `CpuUtilizationCgroupContainerProvider`.

Currently the other phases show as blank, but they will be removed in a future versions: https://metrics.green-coding.berlin/stats.html?id=b3527cc8-c17d-4284-95c4-aa4f353682fa

Caveats / TODOs:
- Currently the monitor mode is ended through a SIGINT (CTRL+C). If used in production it should not react to SIGHUP and also properly terminate on SIGTERM
- The GMT gets confused when monitor runs are compared. This has to be guard-claused.
- Untested how the tool behaves when left logging for 24 hours. This was not the actual design case of the monitor mode but people might use it in that form. TBD
- The implementation for the container discovery is done with a native linux cgroup implementation. 
  - Advantages: Extremely performant
  - Disadvantages: 
    - Since docker API is not used the names for the containers are unknown
    - We cannot scan containers under macOS
   
For docker name problem one could argue that a hashmap can be used where the already known containers are kept in so that lookups are kept to a minimum. 
However bringing a hashmap into the C code is quite some work and will create some overhead.

For the macOS problem: This is also already the case with the GMT. However what we could actually do is enter the VM and inject a script. 
Entering the VM via a tty is possible like so:
`stty -echo -icanon && nc -U ~/Library/Containers/com.docker.docker/Data/debug-shell.sock && stty sane`

Leaving this as a point of discussion ...


ArneTR:
Here for discussion my test implementation with glib hashmap:

```
container_t* docker1 = malloc(sizeof(container_t));
    sprintf((*docker1).path,"asdasd");

    container_t* docker2 = malloc(sizeof(container_t));
    sprintf((*docker2).path,"xxxx");


    // Create a new hash table with custom string hash and equal functions
    GHashTable *myHashTable = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);

    // Insert some key-value pairs
    g_hash_table_insert(myHashTable, g_strdup("docker1"), docker1);
    g_hash_table_insert(myHashTable, g_strdup("docker2"), docker2);

    gpointer lookup_key = NULL;
    lookup_key = g_hash_table_lookup(myHashTable, "docker1");
    if(lookup_key != NULL) {
        printf("Found 'docker1': %s\n", ((container_t*)lookup_key)->path);
    }

    lookup_key = g_hash_table_lookup(myHashTable, "docker2");
    if(lookup_key != NULL) {
        printf("Found 'docker2': %s\n", ((container_t*)lookup_key)->path);
    }

    lookup_key = g_hash_table_lookup(myHashTable, "docker3");
    if(lookup_key != NULL) {
        printf("Found 'docker3': %s\n", ((container_t*)lookup_key)->path);
    } else {
        printf("Could not find docker3\n");
    }


    printf("Ready to free\n");
    g_hash_table_destroy(myHashTable);
```

ribalba:
Hey, I really like how clean this is. Not quite how I would have implemented it though. I am not quite sure what the result of the monitor mode is. As far as I can see there is nothing that orchestrates the containers anymore and such it is more like a tool that just runs on the system and logs the resource usage of all containers. I think this is a really useful functionality and could be something we could really use. Also is the frontend built to handle data like this? Wouldn't we need to add more details on when containers are started and killed?
I understood the monitor mode to still adhere to some sort of orchestration. So let's say you want to see how the scaling of containers impacts energy usage you would still have some sort of flow but just monitor all containers on the system and not the predefined. I would also keep the option to have services just with a service also being able to start other containers that are then monitored. Not that easy to describe.

ArneTR:
Keeping the services as optionals is a good idea.

Also a "trigger" will be implemented that can start an action and then see how the system behaves.

ribalba:
Shall I take another look at this?

ArneTR:
No, i removed you as a reviewer. I have kept this open since I sent it to an external party to review and found it much clearer to leave it open that just merge it in for now.

PR #634, opened 2023-12-27
GMT color via own commit hash

ArneTR:
The GMT can now color in the Timeline view not only by the hash of the tested software, but also by the hash of the GMT itself.

This functionality is used to visually debug energy anomalies when trying to discern wether a change in the software or a change in the tool itself is causing an energy anomaly.

Also the hashes are made clickable in the note. A problem however is that the note itself cannot be clicked in the current form.
@ribalba Do you know of a way to make the note hovering persistent in eCharts?

Sorting GMT:
Screenshot 2023-12-27 at 12 31 41 PM

Sorting Software:

Screenshot 2023-12-27 at 12 31 47 PM

### DIFF
--- a/api/api_helpers.py
+++ b/api/api_helpers.py
@@ -163,7 +163,7 @@ def get_timeline_query(uri, filename, machine_id, branch, metrics, phase, start_
     query = f"""
             SELECT
                 r.id, r.name, r.created_at, p.metric, p.detail_name, p.phase,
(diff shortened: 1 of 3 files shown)


ArneTR:
Here a screenshot of the hovering problem
Screenshot 2023-12-27 at 12 37 38 PM

The blue links cannot be reached

ArneTR:
@ribalba The menu clicking problem has been solved!

PR #635, opened 2023-12-27
MacOS tests on GitHub CI/CD

### LINKED ISSUE #621: Enable testing on Mac
Currently we only test on ubuntu with the GitHub ci. I would also want to run the Mac tests once in a while. GitHub offers support for 'macos-latest' [0] so adding it shouldn't be that hard. 

[0] https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners

ribalba:
We have tried to get our tests running on the GitHub Mac CI/CD runners. Unfortunately it turns out that this is not that easy to do in a reliable matter. While it is possible to actually get the tests to run there is loads of boiler plate code you need to wrap around the original tests to get things to work sort of defeating the purpose of the tests. A few things that needed fixing are:
- As docker on uses a VM you can't use localhost as database path. You need to set the IP of the Qemu VM
- You can't use the official docker image as there is no Windowing system
- Kanico can't write to `/tmp` for some reason so you need to write all the temp data to `~`
- Docker does not run on the M1 series of chips so we need to pin the VM to Intel Macs ]

We are pausing this for now as we don't really officially support Mac for runs. We would love some help though. Feel free if this is something you would want to help us doing.


ArneTR:
As a note : In the dev mode branch I implemented a time reduction for the tests. Maybe this helps for this PR?

If you review that we can merge that in, or you rebase this PR ontop of the `dev-mode` branch. This contains also the macos  test fixes: https://github.com/green-coding-berlin/green-metrics-tool/pull/637

Alternatively you can try our private repository that contains the SSH access to Github runners to start tests and runners interactively with a shell and see how the tests behave

ribalba:
This is me just playing around. Thank you for the feedback. Next step would be to ssh into the Mac runner and see why things are failing. Will discuss with @dan-mm when he is back.

dan-mm:
I'm back, and I agree with @ribalba. ATM it seems to fail just trying to run the install_mac.sh script, so next step is to just ssh into the runner try it manually. 

Would have been quite nice if it just worked out of the box, worth the try! :-D

ArneTR:
Note to you both also in the PR: Adding SSH to the PR while it being public in our organization is not secure. Everyone can connect to that session while it is open.

Either do this in a forked private repository or lock it down with your public key. Please do not run this in our organizations with also potentially secrets exposed.

dan-mm:
That makes sense. As a side note though, the ssh upterm action fails specifically in the case for macos + limited users (https://github.com/fawazahmed0/action-debug).

And github doesn't allow you to make private forks of public repos (for security reasons)

its not a blocker, there's other ways around this (I'll just start with a blank private repo and clone gmt after ssh), but I found it funny that the upterm action fails for this specific use case

ribalba:
closing this as this has changed so much in the meantime we would start new

PR #637, opened 2023-12-28
Dev mode

ArneTR:
This PR introduces parts of the dev-mode and the healthcheck as mentioned in PR https://github.com/green-coding-berlin/green-metrics-tool/pull/556 with @davidkopp 

Following new flags have been introduced:
`--dev-no-sleeps` - Skips all sleeps.
`--dev-no-build` - Skips building of the docker containers. This replaces and extends `--dry-run`
`--dev-no-metrics` - Does not import and run the metric providers

Skipping sleeps can lead to containers not being fully booted. This is why the [healthcheck](https://docs.docker.com/compose/compose-file/05-services/#healthcheck) was implemented along with the long syntax of [depends_on](https://docs.docker.com/compose/compose-file/05-services/#long-syntax-1) as discussed in https://github.com/green-coding-berlin/green-metrics-tool/issues/423

All of the flags have also been employed in the test cases. Expecting a decrease from ~12 minutes to ~3 minutes

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -341,7 +341,7 @@ class Runner:
             # creating benchmarking scripts and you want to have all options in the compose but not in each benchmark.
             # The cleaner way would be to handle an empty service key throughout the code but would make it quite messy
             # so we chose to remove it right at the start.
-            for key in [sname for sname, content in yml_obj['services'].items() if content is None]:
+            for key in [sname for sname, content in yml_obj.get('services', {}).items() if content is None]:
                 del yml_obj['services'][key]
 
             self._usage_scenario = yml_obj
@@ -376,7 +376,7 @@ class Runner:
                                 check=True, encoding='UTF-8')
         for line in result.stdout.splitlines():
             for running_container in line.split(','): # if docker container has multiple tags, they will be split by comma, so we only want to
-                for service_name in self._usage_scenario.get('services', []):
+                for service_name in self._usage_scenario.get('services', {}):
                     if 'container_name' in self._usage_scenario['services'][service_name]:
                         container_name = self._usage_scenario['services'][service_name]['container_name']
                     else:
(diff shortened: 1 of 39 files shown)


ArneTR:
Sadly not quite the improvement as seen locally :(. The tests now take 6 minutes instead of 12. But at least 2x :)

runner.py
```diff
@@ -1416,8 +1483,9 @@ def run(self):
     parser.add_argument('--verbose-provider-boot', action='store_true', help='Boot metric providers gradually')
     parser.add_argument('--full-docker-prune', action='store_true', help='Stop and remove all containers, build caches, volumes and images on the system')
     parser.add_argument('--docker-prune', action='store_true', help='Prune all unassociated build caches, networks volumes and stopped containers on the system')
-    parser.add_argument('--dry-run', action='store_true', help='Removes all sleeps. Resulting measurement data will be skewed.')
-    parser.add_argument('--dev-repeat-run', action='store_true', help='Checks if a docker image is already in the local cache and will then not build it. Also doesn\'t clear the images after a run')
+    parser.add_argument('--dev-no-metrics', action='store_true', help='Skips loading the metric providers. Runs will be faster, but you will have no metric')
+    parser.add_argument('--dev-no-sleeps', action='store_true', help='Removes all sleeps. Resulting measurement data will be skewed.')
```
ribalba, line 1487: Maybe this could also take in a sleep value so you can set it to 1 second and such mitigate some of the errors. It defaults to 0

ribalba (approved):
I didn't go through all the tests but next to my little comment looks good to me

runner.py
ArneTR: Thats a nice idea, but the sleeps quickly balloon up even if you only set a 1 to this.

Having no-sleeps and using the healthcheck is actually a breeze and I am fixing the Nextcloud containers atm and it works very well. 

Let's reconsider this once we run into issues with the 0-sleep implementation

ribalba:
@ArneTR is there also a PR to update the documentation?

PR #638, opened 2023-12-29
Dev mode timetravel

ArneTR:
This PR introduces the timetravel mode for developing usage scenarios.

Everytime an error occurs in a flow the developer can re-run a certain part of the flow or even reload the services inline.

This saves enormous time when patching behaviour of the triggered files from the usage scenario.

Example screenshot when run with `python3 runner.py --uri ~/Sites/green-coding/nextcloud-docker --filename energy-tests/usage_scenario-sqlite-event-chromium.yml  --name Nextcloud-Test --dev-no-metrics --dev-no-sleeps --dev-no-build --dev-flow-timetravel --skip-unsafe` :

Screenshot 2023-12-29 at 12 05 21 PM


ArneTR:
@ribalba Please test once with a usage_scenario that will fail due to timeout or other errors. Looking for a UX feedback

runner.py
```diff
@@ -1084,87 +1076,130 @@ def end_phase(self, phase):
...
         # run the flows
-        for el in self._usage_scenario['flow']:
-            print(TerminalColors.HEADER, '\nRunning flow: ', el['name'], TerminalColors.ENDC)
+        ps_to_kill_tmp = []
+        ps_to_read_tmp = []
+        flow_id = 0
+        flows_len = len(self._usage_scenario['flow'])
+        while flow_id < flows_len:
+            flow = self._usage_scenario['flow'][flow_id]
+            ps_to_kill_tmp = []
```
ribalba, line 1085: I would use the `.clear()` here as it was confusing as it looked like you were redefining from line 1079

runner.py
```diff
@@ -1326,33 +1361,38 @@ def save_stdout_logs(self):
                 """, params=(logs_as_str, self._run_id))
 
 
-    def cleanup(self):
+    def cleanup(self, inline=False):
```
ribalba, line 1364: I would rename `inline` to `keep_state` or something like that as inline is not really clear what it does

ribalba (approved):
Few minor things. The main point I would make that I would want to time travel even if there is no error.

runner.py
ArneTR: renamed to continue_measurement

ArneTR:
> Few minor things. The main point I would make that I would want to time travel even if there is no error.

I like it. Thanks for the design idea.

PR #642, opened 2024-01-04
Forbidding script to be called with any arguments

ArneTR:
Although not critical I think we should be explicit. 

The script should only be called without any arguments


ArneTR:
This was a bit rushed, as I fixed this elsewhere. The command in the install script is actually not at risk.

Fix was done in the docs though for the cluster: https://github.com/green-coding-berlin/documentation/commit/4f6835403df9d3af615fc0d21bef16ebc6bda9ca

PR #643, opened 2024-01-05
Temperature Baseline cooldown in client.py

ArneTR:
This PR removes the static sleeping feature of the Cluster and adds a a dynamic cooldown temperature

### DIFF
--- a/metric_providers/lm_sensors/source.c
+++ b/metric_providers/lm_sensors/source.c
@@ -176,7 +176,8 @@ static void output_value(int value, char *container_id) {
 
 int main(int argc, char *argv[]) {
     int c, err;
-    int one_off = 0;
+    int measurement_amount = 0;
+    int i = 0;
     const char *config_file_name = NULL;
 
     // These are the lists that we pass in through the command line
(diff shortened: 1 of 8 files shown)


metric_providers/lm_sensors/source.c
```diff
@@ -202,7 +203,7 @@ int main(int argc, char *argv[]) {
     setlocale(LC_CTYPE, "");
 
     while (1) {
-        c = getopt_long(argc, argv, "c:f:hts:i:", long_opts, NULL);
+        c = getopt_long(argc, argv, "c:f:hts:i:1", long_opts, NULL);
```
ribalba, line 206: I understand the 1 for one record. But this sort of implies that there is also a 2 ... And numbers as args is something I haven't seen that often. Maybe using something like `-n` from head so that you do `-n 1` for record?

ribalba (approved):
I really like this PR. Will give us really nice data on how machines cool down. Great work

metric_providers/lm_sensors/source.c
ArneTR: Jap. I like the `-n 1`.

Will add

PR #650, opened 2024-01-12
standardized test-config

dan-mm:
Notes:
- only CpuUtilizationProcfsSystemProvider && PsuEnergyAcSdiaMachineProvider turned on at the moment
- if we want, we can use PsuEnergyAcXgboostMachineProvider instead of PsuEnergyAcSdiaMachineProvider, either works (but not both)
- no changes needed for VM as these are the only providers that were running in the VM, but we no longer need to use the disable-metrics-provider script (the script itself is still in, just in case we ever want to use that functionality... but maybe cleanup would be preferred?)
    - example applications will need a small update after merge in order to call the GMT-pytest action correctly (without disable script)
    
- no changes to test-compose.yml or structure.sql (they are also copied). 
    - for test-compose.yml, that is because it copies from compose.yml.example directly, not the user's compose.yml
    - for structure.sql, its because i don't expect the user to be editing that file anyways (and copying makes it easier to keep in sync)
    
- is it ok to have the test-db pw hardcoded here? Other option would be to have it be a variable that is set during the setup-test-env script, but then we get into .gitignore issues (unless we have a test-config.yml.example which is copied over, but that just feels silly at this point)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ metric-provider-binary
 static-binary
 .pytest_cache
 test-compose.yml
-test-config.yml
 tests/structure.sql
 tools/sgx_enable
 venv/
--- a/test-config.yml
+++ b/test-config.yml
@@ -1,67 +1,79 @@
+postgresql:
+  host: test-green-coding-postgres-container
+  user: postgres
+  dbname: test-green-coding
+  password: testpw
+  port: 9573
+
+smtp:
+  server: null
+  sender: null
+  port: null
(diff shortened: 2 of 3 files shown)


ArneTR:
Somehow my comment from last week got lost, so replaying the comment now:

- Hardcoded PW in the test config is fine
- Please use XGBoost reporter with some average defaults
- Please make the test-config.yml in the same ordering and indenting as the config.yml.example.

PR #651, opened 2024-01-12
Parallelize tests

### LINKED ISSUE #439: Add parallel testing
Currently the tests run for quite some time and there are some tests we could run in parallel. There is the great `pytest-xdist` package that uses workers to run tests in parallel. At some stage we should maybe look through all the tests and categorize them into two groups so we can seep up the ones that we can parallelize.

Very low prio though.

dan-mm:
Not ready yet, making PR to test in VM


ArneTR:
I do understand it is not ready yet. Two questions still:
- Can you make a draft PR in the future for these? I think it is even more clear to outsiders and us internally what the state of the PR is then. Looks like this: https://github.com/green-coding-berlin/green-metrics-tool/pull/630
- The implementation route is now so that the runner.py is changed. Wasn't the idea to patch the usage_scenario and variables of the runner at runtime? Or does the PR not reflect this discussion we had yet?

dan-mm:
> * Can you make a draft PR in the future for these? I think it is even more clear to outsiders and us internally what the state of the PR is then. Looks like this: [Monitor mode #630](https://github.com/green-coding-berlin/green-metrics-tool/pull/630)
Aye, I can make drafts for the future

> * The implementation route is now so that the runner.py is changed. Wasn't the idea to patch the usage_scenario and variables of the runner at runtime? Or does the PR not reflect this discussion we had yet?
I think there was a miscommunication here, I thought we had agreed that the runner.py implementation was to be changed? Patching the usage_scenario will not be enough, as we also need to change the /tmp/green-metrics-tool folder that gets used to be unique.

The discussion regarding patching the usage_scenario was in response to a checking I was doing against the container names from the flow section to match against the parallel_id. You pointed out that patching the object at runtime would make more sense than regex matching, but I thought this was still in context of within the runner.py - which is what I've implemented.

dan-mm:
@ArneTR - ready for review!

ArneTR:
- Can you give me a summary for that commit / design description
- Also: Can you tell me how many tests are parallelized in the end?
  - I see that the test time on Github Actions has gone down from 7 Minutes to 5.30 minutes. Is that what you are seeing on local too? Is that what you expect?

dan-mm:
Summary:

- test_functions does the setup_runner differently now: it parses through the usage_scenario and docker compose files that are passed in, and appends a parallel_id (usually a random 12 digit number) to any container names and networks
- after creating a Runner object, it also edits its _tmp_folder and _folder to also be unique (tmp/gmt_tests-{parallel_id}/green-metrics-tool/")

- not all tests were properly using dev_no_build as that wasn't the default in setup_runner (this was the root cause of the docker issue I was seeing). dev_no_build is now the default

- tests that need a uri now more uniformly create (and cleanup) a tmp folder in the tests subdirectory to use as its project folder

- most tests had to be updated to properly check for the parallel_id in their expectations and call the setup_runner correctly

- the base gcb_stress_gmt_tmp_image is now created before any tests are run in conftest.py
- similarly the conftest.py cleanup step is a bit more thorough (getting rid of anything under tests/tmp and clearing db, and done once after all tests are finished)

- I have added a pytest marker (pytest.mark.serial) to mark any tests that need to be run in serial. I've added documentation in the tests/Readme regarding this and running the tests now.

- Custom yaml loader moved from inline definition in runner.py into yml_helper.py file
- similarly, join_paths also moved to utils.py
- I was reusing these functions to the the usage_scenario yaml loading/parsing during setup_runner



As for the reduction, it is what I expect. These are number and times of the kinds of tests:

Parallel: 57 tests; time: 0:01:23
Serial: 15 tests; time: 0:01:01

so of that 5.30, 2.24 minutes test time, 3,06 workflow overhead

From the last main branch workflow, 72 tests; 0:03:41 test time and 3,32 workflow overhead. Which is almost 50% reduction in the test time itself.

It is what I expect as I see a similar percentage reduction in local:

Serialized: 0:07:16 total test time
With parallelization (+ the 15 serial): 0:02:11 + 0:02:09, 4:20 total test time



However.... I have to take back that this PR is ready. When I was running the tests locally to benchmark these times some more I'm seeing some sporadic failures still. It didn't show up when I tested earlier today during pushing, or in the PR tests, but they're there. As best as I can tell right now I think some cleanups are happening early, but I'm not sure yet. At least one more commit is incoming.

ArneTR:
@ribalba Please give it a review and validate if the approach is feasible and satisfies the complexity for the expected gains. Then I would like to plan further steps on this

ArneTR:
I wil make this PR as a draft to have it as concept in the backlog.

The idea and general implemementation of the parallelization is nice but the PR has some issues:
- Currently it is experiencing unknown race conditions
- many additional refactorings and small changes are added which are unclear if they might be causal for the breaking functionality or just refactorings
- Time gain is not as big as expected due to many synchronous downgrades

ArneTR:
Closing this as heavily outdated

PR #653, opened 2024-01-12
618 check systems test rewrite

dan-mm:
One thing I noticed while doing this is the following: here is the check_one_psu_provider function that is called in systems_check.py:

```
def check_one_psu_provider():
    metric_providers = utils.get_metric_providers(GlobalConfig().config).keys()
    energy_machine_providers = [provider for provider in metric_providers if ".energy" in provider and ".machine" in provider]
```
it checks for keys that have .energy and .machine in the key and determines those as psu providers. Why do we not check for keys that have .psu instead?


ArneTR:
> One thing I noticed while doing this is the following: here is the check_one_psu_provider function that is called in systems_check.py:
> 
> ```
> def check_one_psu_provider():
>     metric_providers = utils.get_metric_providers(GlobalConfig().config).keys()
>     energy_machine_providers = [provider for provider in metric_providers if ".energy" in provider and ".machine" in provider]
>     return len(energy_machine_providers) <= 1
> ```
> 
> it checks for keys that have .energy and .machine in the key and determines those as psu providers. Why do we not check for keys that have .psu instead?

The naming of the test is misleading. What we actually want to check is that we do not have two providers that provider energy and scope machine. Where the value is coming from is irrelevant. It can be a PSU, but also a PDU would be fine.

Please change the name of the test and maybe also the error message if that claims PSU is the problem rather than energy_ machine_ . Then fine to merge

PR #656, opened 2024-01-14
chore: add info about forking workflow for contributors

alexzurbonsen:
Suggestion: I find it helpful to explicitly state that GMT wants to receive pull requests via forks. Since you can also create pull requests without a fork.


ArneTR:
Thanks for the PR! A very welcome clarification in the contribution guideline.

PR #676, opened 2024-02-09
Initial version of the carbonDB

ribalba:
This is the first version of the CarbonDB GMT implementation. 
I will push the data generator ASAP. There are still a few minor things to fix there.

### DIFF
--- a/api/api_helpers.py
+++ b/api/api_helpers.py
@@ -1,9 +1,12 @@
+import io
 import ipaddress
 import json
 import uuid
 import faulthandler
 from functools import cache
 from html import escape as html_escape
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import ORJSONResponse
 import numpy as np
 import requests
 import scipy.stats
@@ -669,6 +672,10 @@ def get_carbon_intensity(latitude, longitude):
     if not (token := GlobalConfig().config.get('electricity_maps_token')):
         raise ValueError('You need to specify an electricitymap token in the config!')
 
+    if token == 'testing':
+        # If we are running tests we always return 1000
+        return 1000
+
     headers = {'auth-token': token }
     params = {'lat': latitude, 'lon': longitude }
 
@@ -682,3 +689,61 @@ def get_carbon_intensity(latitude, longitude):
         return resp_data.get('carbonIntensity')
 
     return None
+
+def carbondb_add(client_ip, energydatas):
+
+    latitude, longitude = get_geo(client_ip)
+    carbon_intensity = get_carbon_intensity(latitude, longitude)
+
+    data_rows = []
+
+    for e in energydatas:
+
+        if not isinstance(e, dict):
+            e = e.dict()
+
+        e = html_escape_multi(e)
+
+        fields_to_check = {
+            'type': e['type'],
+            'energy_value': e['energy_value'],
+            'time_stamp': e['time_stamp'],
+        }
+
+        for field_name, field_value in fields_to_check.items():
+            if field_value is None or str(field_value).strip() == '':
+                raise RequestValidationError(f"{field_name.capitalize()} is empty. Ignoring everything!")
+
+        if 'ip' in e:
+            # An ip has been given with the data. Let's use this:
+            latitude, longitude = get_geo(e['ip'])
+            carbon_intensity = get_carbon_intensity(latitude, longitude)
+
+        energy_kwh = float(e['energy_value']) * 2.77778e-7
+        co2_value = energy_kwh * carbon_intensity
+
+        company_uuid = e['company'] if e['company'] is not None else ''
+        project_uuid = e['project'] if e['company'] is not None else ''
+        tags_clean = "{" + ",".join([f'"{tag.strip()}"' for tag in e['tags'].split(',') if e['tags']]) + "}" if e['tags'] is not None else ''
+
+        row = f"{e['type']}|{company_uuid}|{e['machine']}|{project_uuid}|{tags_clean}|{int(e['time_stamp'])}|{e['energy_value']}|{co2_value}|{carbon_intensity}|{latitude}|{longitude}|{client_ip}"
+        data_rows.append(row)
+        print(row)
+
+    data_str = "\n".join(data_rows)
+    data_file = io.StringIO(data_str)
+
+    columns = ['type', 'company', 'machine', 'project', 'tags', 'time_stamp', 'energy_value', 'co2_value', 'carbon_intensity', 'latitude', 'longitude', 'ip_address']
+
+    DB().copy_from(file=data_file, table='carbondb_energy_data', columns=columns, sep='|')
+
+    DB().query("""
+        DELETE FROM carbondb_energy_data
+        WHERE ctid NOT IN (
+            SELECT min(ctid)
+            FROM carbondb_energy_data
+            GROUP BY time_stamp, machine, energy_value
+        )
+    """)
+
+    return ORJSONResponse({'success': True}, status_code=202)
--- a/api/api_helpers.py
+++ b/api/api_helpers.py
@@ -1,8 +1,11 @@
+import ipaddress
+import json
 import uuid
 import faulthandler
 from functools import cache
 from html import escape as html_escape
 import numpy as np
+import requests
 import scipy.stats
 
 from psycopg.rows import dict_row as psycopg_rows_dict_row
@@ -12,6 +15,7 @@ from pydantic import BaseModel
 faulthandler.enable()  # will catch segfaults and write to STDERR
 
 from lib.db import DB
+from lib.global_config import GlobalConfig
 
 def rescale_energy_value(value, unit):
     # We only expect values to be mJ for energy!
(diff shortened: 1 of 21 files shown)


tools/carbondb_compress.py
```diff
@@ -0,0 +1,55 @@
...
+        GROUP BY
+            e.type,
+            e.company,
+            e.machine,
+            e.project,
+            e.tags,
+            DATE_TRUNC('day', TO_TIMESTAMP(e.time_stamp / 1000000))
+        ON CONFLICT (machine, date) DO UPDATE
+        SET
+            energy_sum = EXCLUDED.energy_sum,
```
ArneTR, line 46: What is this EXCLUDED? I have never seen that

tools/carbondb_compress.py
```diff
@@ -0,0 +1,55 @@
...
+            SUM(e.energy_value) AS energy_sum,
+            SUM(e.co2_value) AS co2_sum,
+            AVG(e.carbon_intensity) AS carbon_intensity_avg,
+            COUNT(*) AS record_count
+        FROM
+            carbondb_energy_data e
+        LEFT JOIN
+            carbondb_energy_data_day d ON e.machine = d.machine
+                                        AND e.project = d.project
+                                        AND DATE_TRUNC('day', TO_TIMESTAMP(e.time_stamp / 1000000)) = d.date
```
ArneTR, line 36: what is `d.date` ? The table d is nowhere defined if I am not overlooking something.

tools/carbondb_compress.py
```diff
@@ -0,0 +1,55 @@
...
+            energy_sum = EXCLUDED.energy_sum,
+            co2_sum = EXCLUDED.co2_sum,
+            carbon_intensity_avg = EXCLUDED.carbon_intensity_avg,
+            record_count = EXCLUDED.record_count;
+    """
+    print(DB().query(query))
+
+# pylint: disable=broad-except
+if __name__ == '__main__':
+    compress_data()
```
ArneTR, line 55: Where do you delete in order to save space?

frontend/js/carbondb-details.js
```diff
@@ -0,0 +1,71 @@
...
+            var measurements = await makeAPICall(`/v1/carbondb/machine/day/${machine_uuid}`);
+        } catch (err) {
+            showNotification('Could not get data from API', err);
+            return;
+        }
+        if (measurements.data.length == 0){
+            showNotification('No data', 'We could not find any data. Did you follow a correct URL?')
+            return
+        }
+        console.log(measurements)
```
ArneTR, line 20: !

frontend/js/carbondb-details.js
```diff
@@ -0,0 +1,71 @@
+$(document).ready(function () {
+    function getURLParameter(name) {
+        return new URLSearchParams(window.location.search).get(name);
+    }
+
+    (async () => {
+        let mData
```
ArneTR, line 7: unused?

frontend/carbondb-details.html
```diff
@@ -0,0 +1,106 @@
...
+    <title>Green Metrics Tool</title>
+    <meta name="referrer" content="no-referrer-when-downgrade" />
+    <meta name="description" content="Measuring the energy of your software" />
+    <script src="/dist/js/jquery.min.js" defer></script>
+    <script src="/dist/js/datatables.min.js" defer></script>
+    <script src="/dist/js/echarts.min.js" defer></script>
+    <script src="/dist/js/toast.min.js" defer></script>
+    <script src="/dist/js/transition.min.js" defer></script>
+    <script src="/dist/js/transition.min.js" defer></script>
+    <script src="/dist/js/modal.min.js" defer></script>
```
ArneTR, line 22: do you need modals and dimmers here?

frontend/carbondb-details.html
```diff
@@ -0,0 +1,106 @@
...
+    <link rel="shortcut icon" href="/images/favicon.ico">
+    <link rel="icon" href="/images/favicon.ico">
+    <link rel="icon" href="/images/favicon-16x16.png" sizes="16x16">
+    <link rel="icon" href="/images/favicon-32x32.png" sizes="32x32">
+    <!-- Site Properties -->
+    <title>Green Metrics Tool</title>
+    <meta name="referrer" content="no-referrer-when-downgrade" />
+    <meta name="description" content="Measuring the energy of your software" />
+    <script src="/dist/js/jquery.min.js" defer></script>
+    <script src="/dist/js/datatables.min.js" defer></script>
```
ArneTR, line 17: do you need datatables here?

docker/structure.sql
```diff
@@ -289,3 +289,81 @@ CREATE TRIGGER hog_tasks_moddatetime
...
+    BEFORE UPDATE ON carbondb_energy_data
+    FOR EACH ROW
+    EXECUTE PROCEDURE moddatetime (updated_at);
+
+
+CREATE INDEX idx_carbondb_company ON carbondb_energy_data(company);
+CREATE INDEX idx_carbondb_machine ON carbondb_energy_data(machine);
+CREATE INDEX idx_carbondb_project ON carbondb_energy_data(project);
+CREATE INDEX idx_carbondb_time_stamp ON carbondb_energy_data(time_stamp);
+CREATE INDEX idx_carbondb_energy_value ON carbondb_energy_data(energy_value);
```
ArneTR, line 324: You are often querying with multiple conditions. Shall we also add a a compound key?

docker/structure.sql
```diff
@@ -289,3 +289,81 @@ CREATE TRIGGER hog_tasks_moddatetime
...
+
+CREATE TRIGGER carbondb_energy_data_day_moddatetime
+    BEFORE UPDATE ON carbondb_energy_data_day
+    FOR EACH ROW
+    EXECUTE PROCEDURE moddatetime (updated_at);
+
+
+CREATE INDEX idx_carbondb_hour_company ON carbondb_energy_data_day(company);
+CREATE INDEX idx_carbondb_hour_machine ON carbondb_energy_data_day(machine);
+CREATE INDEX idx_carbondb_hour_project ON carbondb_energy_data_day(project);
```
ArneTR, line 351: You are often querying with multiple conditions. Shall we also add a a compound key?

api/main.py
```diff
@@ -1204,5 +1206,115 @@ async def get_ci_badge_get(repo: str, branch: str, workflow:str):
...
+            'type': e.type,
+            'energy_value': e.energy_value,
+            'time_stamp': e.time_stamp,
+        }
+
+        for field_name, field_value in fields_to_check.items():
+            if field_value is None or field_value.strip() == '':
+                raise RequestValidationError(f"{field_name.capitalize()} is empty. Ignoring everything!")
+
+        if DB().fetch_one('SELECT id FROM carbondb_energy_data WHERE time_stamp=%s AND machine=%s AND energy_value=%s',
```
ArneTR, line 1252: Why does this guard clause exist? Is that a typical problem?

In larger data ingests this seems rather inefficient. 

Maybe just do batch insert and not line by line and a de-duplication after all inserts?

api/api_helpers.py
```diff
@@ -561,3 +565,120 @@ def get_t_stat(length):
...
+    print(f"Accessing https://ipapi.co/{ip}/json/")
+    if response.status_code == 200:
+        resp_data = response.json()
+
+        if 'error' in resp_data or 'latitude' not in resp_data or 'longitude' not in resp_data:
+            return (None, None)
+
+        resp_data['source'] = 'ipapi.co'
+
+        query = "INSERT INTO ip_data (ip_address, data) VALUES (%s, %s)"
```
ArneTR, line 607: refactor to have this outside of the vendor specific functions?

ArneTR (changes requested):
Code readability is very good! Nice PR!

My main questions are:
- I could not test it as I have no driver currently for the API. Can you link the one you have been using in development?
- The carbonDB feels like an addon on top of what Eco-CI is currently doing. What are your thoughts of unifying these two functionalites / API signatures?

tools/carbondb_compress.py
ribalba: To do updates I only allow a record which are unique for machine and date. If there is a conflict I update the fields. The EXCLUDED fields are the one of the conflicting record. Like this I can do the update in one SQL statement. 

https://www.postgresql.org/docs/current/sql-insert.html and then search for EXCLUDED

ribalba:
> My main questions are:
> 
> * I could not test it as I have no driver currently for the API. Can you link the one you have been using in development?
> * The carbonDB feels like an addon on top of what Eco-CI is currently doing. What are your thoughts of unifying these two functionalites / API signatures?

- Just finishing up the code. Will be here https://github.com/green-coding-solutions/eco-server-energy-estimation
- Yes I am planning of having Eco-CI and the Hog submitting values to the CarbonDB. While EcoCi and Hog will give you insights into more details carbonDB will give you an overview. But I see Hog and Eco-CI as part projects of CarbonDB.

tools/carbondb_compress.py
ribalba: LEFT JOIN clause: This joins another table to the primary table mentioned in the FROM clause. The table being joined is carbondb_energy_data_day, aliased as d. A LEFT JOIN ensures that all records from the left table (carbondb_energy_data aliased as e) are included in the joined table, even if there is no matching record in the right table (carbondb_energy_data_day aliased as d).
ON clause: This specifies the condition used to match rows from the two tables being joined. There are three conditions here, all of which must be true for the rows to be considered a match:
e.machine = d.machine: The machine column of both tables must have the same value.
e.project = d.project: The project column of both tables must have the same value.
DATE_TRUNC('day', TO_TIMESTAMP(e.time_stamp / 1000000)) = d.date: This is a bit more complex. It involves converting a timestamp value in the e.time_stamp column to a date, and then truncating it to the day component, to match it against the d.date column. Here's what each function does:
TO_TIMESTAMP(e.time_stamp / 1000000): Since the time_stamp column is likely in microseconds (given the division by 1,000,000), this converts the time_stamp value from microseconds to a PostgreSQL timestamp.
DATE_TRUNC('day', ...): This truncates the timestamp obtained in the previous step to the start of the day (00:00:00), removing the time component.
The resulting date is then compared to the date column in the d table. If they are the same, the row from e is considered to match with the row from d.
The purpose of this join is to correlate entries in the carbondb_energy_data table with corresponding entries in the carbondb_energy_data_day table based on matching machine and project identifiers, as well as ensuring that the dates (ignoring time beyond the day) match between the two tables. This allows for combining or comparing daily aggregated data (from carbondb_energy_data_day) with more granular timestamped data (from carbondb_energy_data).

All my words in form of ChatGPT.

tools/carbondb_compress.py
ribalba: I don't in the first iteration. I want to check if IPs change. I want to revisit this in a few weeks and look at the data. Once I have confirmed that everything works as expected I will delete the data.

frontend/js/carbondb-details.js
ribalba: Well spotted.

api/api_helpers.py
ribalba: I thought about this but the lines we could refactor are actually only 2 the `insert` which are 2 lines. So the refactor would be more lines of code and not make it a lot clearer. Would make refactoring the insert better. I could go either way thinking about it. I see pro and cons in both

api/main.py
ribalba: Changed to batch insert

ribalba:
@ArneTR ready for next review

tools/carbondb_compress.py
ArneTR: Thanks for the lenghty explanation. My question was actually quite simpler. I know all the mechanis but searched for the "AS" keyword. I have never seen an alias without it. But now know that `LEFT JOIN table as t` = `LEFT JOIN table t`

PR #344, opened 2023-06-24
Macos utilization provider

ArneTR:
It uses the `host_statistics` syscall in macos, same way as [htop does](https://github.com/htop-dev/htop/blob/main/darwin/DarwinMachine.c)

However sometimes I am getting 0 results for no reason ... 

2. The provider atm cannot be installed through the local scripts and even linux install will fail. We need to alter our directory structure to easily automate this.

Issue #330, opened 2023-06-15
utilization reporter for macOS


ArneTR:
We currently have no reporter that could provide that for macOS. For linux the `cpu.utilization.procfs.system.provider.CpuUtilizationProcfsSystemProvider` reporter is used to generate that data. It acutally just reads from the `/proc/stat` file and spits out a file with timestamps and cpu utilizations.

See in the screenshot as an example. The 208 etc. means 2.08%
<img width="1162" alt="Screenshot 2023-06-15 at 7 21 01 PM" src="https://github.com/green-coding-berlin/green-metrics-tool/assets/250671/31614972-d69c-454c-8865-dfeca365ea49">

So in short: The GMT currently lacks this functionality for macOS but it could be really easily implemented I guess.

What needs to be done in the PR is a small piece of code, either C-code / Python or bash or similar, that reads the current CPU utilization from the macOS subsystem. 
We pull that in as metric provider plugin and do the refactoring to make the reporter work with the alternate data source.

@ribalba Maybe this issue is interesting for you, or we open it up as good first issue?

Do you have any idea how to easily get this data out of the macOS subsystem? Where to read cpu utilization "raw" ?

ArneTR:
Because of https://github.com/green-coding-berlin/green-metrics-tool/pull/650 we do not have a standard working macOS test setup anymore.

I propose to prioritize this issue to have that running again.

@ribalba What is your time expectation for this task?

dan-mm:
The common providers (XGBoost) should work, no? That one is still available in the test-config

ribalba:
I think the power metrics provider will still work but this will only be on Mac. I will take a look next week. Is that ok @ArneTR ?

dan-mm:
@ribalba - the providers should all still work in the cases where they work, the linked PR is more about what our "default" test setup is. Before, local tests were choosing which metric-providers to use based on people's personal config.yml file, but this had issues, so now we have a static test-config.yml. 

Currently this does not have power-metrics turned on by default, as I don't think that's a guarantee for all macos devices (I can confirm this provider doesn't work in Github mac VM's for example) - this is what Arne is referring to here, I think he would like a mac-os provider that is more or less general for all mac owners, like how we have the linux util provider

ArneTR:
good call, I forgot about the powemetrics one ... reasons unclear :)

[...]

PR #681, opened 2024-02-14
Adds support for cpu utilization on mac with resolution of under 100ms

ribalba:
Closes https://github.com/green-coding-solutions/green-metrics-tool/issues/330
Closes https://github.com/green-coding-solutions/green-metrics-tool/pull/344

I choose to make a new pr as it seemed cleaner.


ribalba:
Why are you renaming it to mach? We don't use mach anywhere. It is mac everywhere?

metric_providers/cpu/utilization/mac/system/source.c
```diff
@@ -35,7 +35,7 @@ void loop_utilization(unsigned int msleep_time) {
             }
 
             gettimeofday(&now, NULL);
-            printf("%ld%06i %i\n", now.tv_sec, now.tv_usec, (int)(ut_total * 100 / numCPUsU));
+            printf("%ld%06i %i\n", now.tv_sec, now.tv_usec, (int)( (ut_total / (float)numCPUsU)*100*100) );
```
ribalba: Good catch. I forgot that we multiply by 100 again.

ArneTR:
Works great, really nice commit!

I patched only that the utiliatzion was in percent and not in ratio. Other than that some name changes and display additons.

TY!

@dan-mm Please use this provider for your mac tests

ArneTR:
> Why are you renaming it to mach? We don't use mach anywhere. It is mac everywhere?

Yeah sorry, I did not clarify. Our [naming scheme](https://docs.green-coding.io/docs/measuring/metric-providers/metric-providers-overview/#naming-convention) specifies that this part of the string should declare the source. Mac is not really the source, but the mach syscall library is. Similar to the procfs as the vfs as source in the other CPU utilization provider. Hope that makes it more clearer.

PR #693, opened 2024-02-27
Diffing with artifacts

ArneTR:
Diffing has been added to the compare view.

Since this operation is quite costly I also introduce the *artifact store*

This is a table that can hold JSON items and will be indexed by a text-like structure and a type. @ribalba Do you think this will be sufficient for our caching purposes?

I imagine:
- Comparison Diffs
- Statistical data from comparisons as structures object data (json)

What other artifacts might we have?

metrics green-coding internal_9142_compare html_ids=f814b51e-e953-4984-932c-1e537dcf8cc0,709fe691-1813-4006-8bb3-781582f9dc6c
metrics green-coding internal_9142_compare html_ids=f814b51e-e953-4984-932c-1e537dcf8cc0,709fe691-1813-4006-8bb3-781582f9dc6c (1)


ArneTR:
After having a good nights sleep another question came up: @ribalba 
Does it even have any use of making the artifcat store a JSON column? It will anyway only be read from and never searched I guess. 

So just having a blob / text column should be sufficient. Parsing it as JSON we can still do afterwards ... but save where it is not needed to parse.

docker/structure.sql
```diff
@@ -289,3 +289,20 @@ CREATE TRIGGER hog_tasks_moddatetime
     EXECUTE PROCEDURE moddatetime (updated_at);
 
 CREATE INDEX idx_task_coalition ON hog_tasks(coalition);
+
+
+CREATE TABLE artifacts (
+    id SERIAL PRIMARY KEY,
+    type text,
+    key text,
+    data json,
```
ribalba, line 298: As mentioned in your comment this could probably be a text. Should make things a little sleeker

frontend/js/compare.js
```diff
@@ -4,20 +4,37 @@ const getURLParams = () => {
     return url_params;
 }
 
+async function fetchDiff() {
+    document.querySelector('#loader-question').remove();
```
ribalba, line 8: Maybe you could add a https://fomantic-ui.com/elements/loader.html

frontend/js/compare.js
```diff
@@ -4,20 +4,37 @@ const getURLParams = () => {
...
+    const url_params = getURLParams();
+    document.querySelector('#diff-pre').insertAdjacentHTML('beforebegin', '<h2>Configuration and Settings diff</h2>')
+    try {
+        let api_url = '/v1/diff?ids=';
+        params.forEach( id => {
+            api_url = `${api_url}${id}`
+        });
+        const diff_data = (await makeAPICall(api_url)).data
```
ribalba, line 16: Why don't you do
```
let api_url = `/v1/compare?ids=${params.join()}`;
```

requirements-dev.txt
```diff
@@ -6,6 +6,7 @@ pylint==3.0.3
 fastapi==0.109.2
 starlette>=0.32
 anybadge==1.14.0
+deepdiff==6.7.1
```
ribalba, line 9: doesn't this belong in the `requirements.txt` Why is it only in `-dev`?

ribalba (changes requested):
Few minor things. Really nice feature

ribalba:
Maybe something like this would make it look nicer
https://github.com/rtfpessoa/diff2html

ArneTR:
This PR needs following rework:

- Redis Cache  only for production
- Unique constraint for artifacts
- TTL for artifacts. 30-60 days
- Diff2HTML testing how it looks by just giving unified diffs as string

ArneTR:
Redi implemented + caching of stats, compare, badge routes. Also with nicer diff view
Screenshot 2024-03-12 at 11 44 28 AM

ArneTR:
See it live on https://metrics.green-coding.io/compare.html?ids=6f34b31e-f35c-4601-ae0d-6fd04a951aaf,f24748ed-5399-4f42-a6c1-9c9ffe586815

ArneTR:
@dan-mm I Introduced a new service in the docker compose file (a redis server), but the tests fail. can you give it a look please?

ArneTR:
@ribalba I found the reason for the PR failing.

In https://github.com/green-coding-solutions/green-metrics-tool/blob/853322ee5f9fac1496d2b3a0892195647ea19e70/optimization_providers/base.py#L68 you are requesting an API endpoint internally that now has a caching implemented.

Since we currently have Redis only available local to the server we cannot access it.

I see two options here:
- Opening up redis from remote and putting a password on it
- Running the optimizations only on the server and never on the measurement nodes

I actually favor the latter as it makes the process more async, which is anyway where we want to go in the future and also it keeps security more tight. But wanted your opinion on this.

ArneTR:
@ribalba Do you have an ETA here on this? Merging is currently blocked due to the implementation of the optimizations

ribalba:
This was on the to discuss list :)

I don't think running the optimisations on the server makes sense as you would need to checkout all the repos again if you want to do something with code. Maybe even build docker containers if you want to see something/ optimise in them. 

I am in favour of having a failover to DB if the cache is not available. I think this should be the standard behaviour as you can't rely on a cache being there. Redis gives you no guarantees on availability. Also I want to develop locally without redis. So we need failover if redis fails anyway.

PR #698, opened 2024-02-29
Added no_python flag and ported flags to mac

ArneTR:
@dan-mm I introduced a new flag in the install script. does this also need to be set in the example application tests somewhere?


dan-mm:
@ArneTR  - nope, just changing the gmt-pytest action is enough, the examples directory workflow calls this action for the installation/setup.

PR #700, opened 2024-03-01
Dynanmic system checks

ArneTR:
@ribalba I added a function call mechansim. How do you like it? Can it be leaner / more neat?


ArneTR:
@dan-mm Why can the variable "system_check_threshold" not be found? It is put in the config.yml.example. Should this not be copied? To which file do i have to add it?

PR #702, opened 2024-03-05
Adds optimisation functionality

ribalba:


ribalba:
@ArneTR ready for review before I start writing all the optimisers

ArneTR:
Did give it a look. Works alright and looks great.

Screenshot attached.

I would change the wording in the end for the box rather to something like this:

"GMT is a FOSS software that shall empower users to make software more sustainable by measuring energy and carbon. We believe that FOSS is the most beneficial way to go here as transparency is key to create trust, adoption and awareness of the carbon emissions of software. Some features like hosting, automated measurements and also optimizations we use as our angle to support the free continious development of the tool. So if you use this tool professionally and want to take the step from awareness and measurement to optimization maybe consider supporting us. If you want to use this tool for research, are an NGO or a student please reach out to us - We have special free and reduced priced options. If you don't want to see this message and you are running your own cluster just add the `message_optimization` to the ignore list into your config file."

I think this reflects our intention better

Screenshot 2024-03-06 at 12 14 39 PM

ribalba:
@ArneTR Feedback added

ArneTR:
then go. Please clear up the conflicts and merge in

PR #710, opened 2024-03-13
Quickfix for the hog not to create so much data

ribalba:


ArneTR:
FYI: The parameters were reduced, but not the placeholders. Fixed it in main

ribalba:
https://www.youtube.com/watch?v=cnaeIAEp2pU

I thought that should be caught in the tests. Will write a test to find this in the future

PR #716, opened 2024-03-21
DB Auto Reconnect on connection drop

ArneTR:
The easiest way is to use the implemented psycopg_pool class which implements dynamic retry times with increasing wait times: https://www.psycopg.org/articles/2021/01/17/pool-design/

Since we have no async or mutli-processing code the pool has no further advantage. I still opted for two connections open to have a failover

The pool itself is not opened through a context manager to have the connections be reused

Furthermore all rollback and error handling code has been removed, as the context manager takes care of this. The error only bubbles if the retry count of 30 seconds is exceeded.

Using this mechanic can be problematic for a benchmarking tool in the first place and allows us only to make DB connections when NO benchmarking is happing, which means in between any actual measurement phases.

@ribalba Is this assumption valid? Did I forget any place where excessive retries might happen that then not fail since after some time the connection comes up again? But then have skewed measurements?

### DIFF
--- a/lib/db.py
+++ b/lib/db.py
@@ -1,9 +1,8 @@
 #pylint: disable=consider-using-enumerate
 
-import psycopg
+from psycopg_pool import ConnectionPool
 
 from lib.global_config import GlobalConfig
-
 class DB:
 
     def __new__(cls):
@@ -13,7 +12,7 @@ class DB:
 
     def __init__(self):
 
-        if not hasattr(self, '_conn'):
+        if not hasattr(self, '_pool'):
             config = GlobalConfig().config
 
             # Important note: We are not using cursor_factory = psycopg2.extras.RealDictCursor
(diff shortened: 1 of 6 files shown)


lib/db.py
```diff
@@ -23,43 +22,51 @@ def __init__(self):
             # force domain socket connection by not supplying host
             # pylint: disable=consider-using-f-string
             if config['postgresql']['host'] is None:
```
ribalba, line 24: Why are you doing this with an if? Couldn't you do 
```
config['postgresql'].get('host', 'localhost')
```

lib/db.py
```diff
@@ -23,43 +22,51 @@ def __init__(self):
...
-                    config['postgresql']['port']))
+                self._pool = ConnectionPool(
+                    "postgresql://%s:%s@localhost:%s/%s" % (
+                        config['postgresql']['user'],
+                        config['postgresql']['password'],
+                        config['postgresql']['port'],
+                        config['postgresql']['dbname'],
+                    ),
+                    min_size=1,
+                    max_size=2,
```
ribalba, line 33: Why are you setting this to `max_size=2` Shouldn't this be the amount of threads the server can have? As threads share the obj?
```
/var/www/startup/venv/bin/gunicorn \
--workers=2 \
--access-logfile=- \
--error-logfile=- \
--worker-tmp-dir=/dev/shm \
--threads=4 \
--worker-class=gthread \
--bind unix:/tmp/green-coding-api.sock \
-m 007 \
--user www-data \
--chdir /var/www/green-metrics-tool/api \
-k uvicorn.workers.UvicornWorker \
main:app
```

lib/db.py
```diff
@@ -23,43 +22,51 @@ def __init__(self):
...
+                        config['postgresql']['user'],
+                        config['postgresql']['password'],
+                        config['postgresql']['host'],
+                        config['postgresql']['port'],
+                        config['postgresql']['dbname'],
+                    ),
+                    min_size=1,
+                    max_size=2,
+                    open=True
+                )
```
ribalba, line 48: Maybe we should think about implementing a 'check' function?

https://www.psycopg.org/psycopg3/docs/api/pool.html#psycopg_pool.ConnectionPool

ribalba (approved):
Minior refactorings

PR #717, opened 2024-03-22
Small fixes

ribalba:


ArneTR:
I still do not know the code for the optimizations well enough to review these fixes.

Can you provide me some notes of what you are intending to fix and describe the changes a bit?

ribalba:
The main thing is 
```
  if 'deploy' in service:
                if memory := service['deploy'].get('resources', {}).get('limits', {}).get('memory', None):
                    docker_run_string.append('-m')
                    docker_run_string.append(str(memory))
                if cpus := service['deploy'].get('resources', {}).get('limits', {}).get('cpus', None):
                    docker_run_string.append('--cpus')
                    docker_run_string.append(str(cpus))
```
in which we extend the services section for resource usage. 
- https://docs.docker.com/config/containers/resource_constraints/#limit-a-containers-access-to-memory
- https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler

This is needed so that we know how much % the resource has been using. 

Otherwise we change the way we handle temporary directories/ files. 

Not really a big change. No Idea why GitHub is showing so many edits. I merged main into the branch.

ArneTR:
I redid the merge to clear up the view.

Cannot really tell if that code is doing what it should and will rather look at this in my full review of the optimizations. So merging in for now.

PR #723, opened 2024-03-25
Adds more tests

ribalba:


tests/api/test_api.py
```diff
@@ -110,3 +114,42 @@ def todo_test_get_runs():
...
     assert res_json['data'][0][0] == str(pid)
+
+
+
+def test_hogDB_add():
+    hog_data_obj  = [
+    {
+        "time": 1710668240000,
+        "data": hog_data.hog_string,
+        "settings": json.dumps({"powermetrics": 5000, "upload_delta": 3, "upload_data": True, "resolve_coalitions": ["com.googlecode.iterm2", "com.apple.terminal", "com.vix.cron"], "client_version": "0.5"}),
```
ArneTR, line 125: @ribalba  can we use orjson here?

tests/api/test_api.py
```diff
@@ -110,3 +114,42 @@ def todo_test_get_runs():
...
+        'time_stamp': str(int(time.time() * 1e6)),
+        'company': '',
+        'project': '',
+        'machine': 'f6d93b14-31c3-4565-9833-675371c67f2f',
+        'tags': "x,y"
+    }]
+
+
+    response = requests.post(f"{API_URL}/v1/carbondb/add", json=energydata, timeout=15)
+    assert response.status_code == 202
```
ArneTR, line 152: @ribalba This here caught my interest. In order to have a proper caching in Cloudflare it is nice to always work with one none-cache return code. So far we have used 204 No-Content, which I think reflects most closest to what is happening.

Why did you choose 202 Accepted ? Can we unify this to 204? Or do you think it makes sense to keep both and adjust on CLoudflare side?

tests/api/test_api.py
ribalba: Yes, what would be the improvement?

tests/api/test_api.py
ribalba: I felt 202 was more what was happening. But 204 is totally ok for me. I didn't know the caching thing.

ribalba:
Shall I make a PR? Or are you on it ?

ArneTR:
I just commented and currently looking at the other PRs. Please do a PR

tests/api/test_api.py
ribalba: I just wanted to fix this. Why do you want to use orjson?

ribalba:
@ArneTR fixed

PR #725, opened 2024-03-29
orjson requirement added

ArneTR:
orjson is a requirement since the optimizations were introduced for the normal runner.py too.

Currently we only install it through `requirements-dev.txt` 

Since our unit tests install this file this dependency issue was not spotted.

This PR normalizes the install by not having functional requirements in the `requirements-dev.txt` but only such needed for pure testing.

Additionally all functional requirements for API includes are now fully installed to reduce complexity as a tradeoff for unneeded install on non-API machines (measurement nodes)

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,10 +1,4 @@
 -r requirements.txt
 pydantic==2.6.4
 pytest==8.1.1
-requests==2.31.0
 pylint==3.1.0
-starlette>=0.32
-anybadge==1.14.0
-
-# just to clear the pylint errors for the files in /api
-scipy==1.12.0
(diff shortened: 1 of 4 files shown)


ArneTR:
@ribalba Put in as reviewer for your opinion, not for technical review per-se

requirements-dev.txt
```diff
@@ -1,11 +1,4 @@
 -r requirements.txt
```
ribalba, line 1: I would add the `docker/requirements.txt` here too as a dependency.

ribalba (approved):
I think that makes total sense to split it up like that. Good idea

PR #726, opened 2024-04-01
Adds co2 values to eco-ci

ribalba:


ribalba:
You can use this script to test
```
curl -X POST "http://api.green-coding.internal:9142/v1/ci/measurement/add" -H 'Content-Type: application/json' -d "{
                \"energy_value\":\"100\",
                \"energy_unit\":\"mj\",
                \"cpu\":\"cpu\",
                \"commit_hash\":\"abc\",
                \"repo\":\"http://google.com\",
                \"branch\":\"main\",
                \"workflow\":\"test\",
                \"run_id\":\"1\",
                \"project_id\":\"\",
                \"label\":\"label\",
                \"source\":\"source\",
                \"cpu_util_avg\":\"50\",
                \"duration\":\"2\",
                \"workflow_name\":\"workflow_name\",
                \"cb_company_uuid\":\"\",
                \"cb_project_uuid\":\"\",
                \"cb_machine_uuid\":\"\",
                \"lat\":\"123\",
                \"lon\":\"456\",
                \"city\":\"CITY\",
                \"co2i\":\"1234\",
                \"co2eq\":\"4567\"
            }"
```

api/main.py
```diff
@@ -1111,6 +1111,11 @@ class CI_Measurement(BaseModel):
     cb_company_uuid: Optional[str] = ''
     cb_project_uuid: Optional[str] = ''
     cb_machine_uuid: Optional[str] = ''
+    lat: Optional[str] = ''
```
ArneTR, line 1114: is that really a string and not a float?

tests/api/test_api.py
```diff
@@ -77,6 +77,36 @@ def test_ci_measurement_add():
...
+                        co2i="100",
+                        co2eq="0.1234567893453245"
+                        )
+    response = requests.post(f"{API_URL}/v1/ci/measurement/add", json=measurement.model_dump(), timeout=15)
+    assert response.status_code == 201, Tests.assertion_info('success', response.text)
+    query = """
+            SELECT * FROM ci_measurements WHERE run_id = %s
+            """
+    data = DB().fetch_one(query, (measurement.run_id, ), row_factory=psycopg.rows.dict_row)
+    assert data is not None
```
ArneTR, line 107: Why not assert that the data is identical? THis would be more strict and also tests for retrieval errors

ArneTR (changes requested):
Looks good to me, apart from some minor questions.

I would in any case keep the PR open until we merge the eco-ci side if some changes are needed there that we need to bring in here

api/main.py
ribalba: Yes it is. When coding I thought about it and then decided to use string as floats are not guaranteed to be precise and I thought we are not really going to do anything with the data computational so I decided that it could be a string. A float would be smaller in footprint though. I also thought about using something like postgis as we could then do distance measurements between where the jobs are run. But decided to KISS it.

ribalba:
@ArneTR added more tests

api/main.py
ArneTR: Thanks for the clarification. Makes sense

ArneTR:
getting errors after merge. @ribalba : can you investigate?

ribalba:
I couldn't find the error. Couldn't reproduce locally. Remerged main and rerun the tests and now it magically works.

PR #730, opened 2024-04-07
Psys

ArneTR:


ArneTR:
@ribalba Can you confirm if this works on your PSYS domain too

ribalba:
yes I can get the PSYS_0 energy when calling the rapl metric provider with `-p`. `-d` is not possible on my machine. I am a little confused as there is also some msr code in this PR. Am I correct that the psys does not work as a metric provider right now?

ribalba (approved):
The naming somehow confused me coming from the other PR. Looks and works great.

PR #732, opened 2024-04-08
Mcp energy fix

ArneTR:
The MCP was experiencing issues when the measurement interval set below 10 ms.

The results where close to zero or even negative apparent power. By identifying the protocol on https://ww1.microchip.com/downloads/aemDocuments/documents/OTH/ProductDocuments/DataSheets/20005473B.pdf it shows that the MCP is using an internal measurement interval to estimate the current frequency and also the power factor.
This estimation can guess a negative power factor, which is physically incorrect and thus a definite bug. Why this error internally occurs is unclear, but it can be mitigated by setting a different measurement interval.

The equation is 2^n*(1/f) (f = 50 Hz for typical networks in europe):
- n=1 => 40 ms
- n=2 => 80 ms
- n=3 => 160 ms
- n=4 => 320 ms
etc.

Good values can be achieved with n=1 so this PR sets this as default.

2. Furthermore when making shorter intervals the current way of calculating energy showed a problem with the OS. The sleep times where getting more unreliable, so calculating the energy beforehand in the C-Code was getting to low values.

By only calculating the energy only when the timestamps are set by reading the output stream gives more reliable and accurate results.

### DIFF
--- a/metric_providers/psu/energy/ac/mcp/machine/provider.py
+++ b/metric_providers/psu/energy/ac/mcp/machine/provider.py
@@ -13,3 +13,33 @@ class PsuEnergyAcMcpMachineProvider(BaseMetricProvider):
             current_dir=os.path.dirname(os.path.abspath(__file__)),
             skip_check=skip_check,
         )
+
+    def read_metrics(self, run_id, containers=None):
+        df = super().read_metrics(run_id, containers)
+
+        '''
+        Conversion to Joules
+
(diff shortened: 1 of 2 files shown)


ArneTR:
@ribalba Please also install it on the green-coding.ai box. Where the MCP code is there located

ribalba:
I assume you edited a few files on the machine?

PR #733, opened 2024-04-09
Normalized -i switch to always be ms

ArneTR:
Some reporters where using `-i 0.1` to sleep for 100 ms and some `-i 100` to sleep for 100 ms

This is now normalized.

The initial design decision was to to the least amount possible in the reporters in terms of conversion when collecting data. However we could not measure any difference in energy or time when using the non-conversion approach. So we opted for better consistency.


ArneTR:
@ribalba Can you give it a quick look? Should be all fine though. IPMI I could not test, since we have no machine atm.

ribalba (approved):
looks good

PR #766, opened 2024-05-06
Trying to force keyword args

ArneTR:


ArneTR:
@ribalba My intention here was to force a signature for the `Job.insert` *@classmethod*. 

Reason being that this method changed signature a couple of times and after a while by just supplying positional arguments the wrong parameters where supplied in the wrong buckets.

This, a bit weird signature how I find, forces keyword arguments and does not allow them to be positional. What do you think?

ribalba:
Good idea. Why not

ribalba (approved):

PR #767, opened 2024-05-10
Contexts and volume included can now happen in parent directories

ArneTR:
It was not possible to include a Dockerfile that was in the repository, but in a parent directory of where the *usage_scenario.yml* currently is.

This PR adds this functionality. To do so it was needed to give the tests some needed refactoring:
- Calling run_until is now done with a context_manager to avoid that cleanup is not called and tests get in unstable state
- tmp directory creation in tests removed and now being explicit with the usage_scenarios instead of joining and importing them together
- Added guard tests for directory escape as well
- Importing a compose file from a non-same directory where the usage_scenario is is now forbidden. Since both can specifiy relative paths they can become out of sync and it is unclear to which working_dir the relative path is relatived


runner.py
```diff
@@ -46,49 +46,9 @@
...
-    if filename == path.rstrip('/'):
-        return filename
-
-    if not filename.startswith(path):
-        raise ValueError(f"{path2} must not be in folder above {path}")
-
-    # To double check we also check if it is in the files allow list
-
-    if mode == 'file':
-        folder_content = [str(item) for item in Path(path).rglob("*") if item.is_file()]
```
ArneTR, line 69: @ribalba I removed this file list scanning as it became quite large when I now check only against the repository root. A repo with 10k+ files could be very memory intensive here.

Do you see a security issue? Since we check against the path and resolve strict=true it should be fine, or?

ribalba: yes, this was just to be double sure

PR #769, opened 2024-05-12
Added sorting by date and unified ci and measurement runs frontend

ArneTR:
CI Runs and Measurement runs can now be sorted by date and show directly when the last run happened.

Since the list became quite long this feature is needed for us to have an overview quickly which repos are active and receiving current runs.

Also sorting by name was made optional through an extra button
Screenshot 2024-05-12 at 1 22 53 PM

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -174,12 +174,15 @@ async def get_machines():
     return ORJSONResponse({'success': True, 'data': data})
 
 @app.get('/v1/repositories')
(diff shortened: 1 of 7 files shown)


ArneTR:
@ribalba I changed one GET reading API endpoint of the CI runs. Eco-CI is not reading from the endpoints ever, just POSTing , correct?

ribalba (approved):
Looks good

Eco-Ci never uses a `get` as far as I know

PR #779, opened 2024-05-21
Allow local URI to be a symbolic link

davidkopp:
Provided path can already be a symbolic link, but not a local URI. This PR fixes this.


ArneTR:
Looks legit and I makes sense. Just for my understanding what your use case is: Why did you not supply the correct path in the first place? Is there a location where you store your repos that are typically symlinked? If so: Why?

My question is to understand the context better and maybe expand and generalize the patch if needed.

davidkopp:
I came across this issue because I'm currently using a new machine with a different user name and therefore a different URI. To be able to use the same commands and paths across my machines, I have added a symbolic link.

ArneTR:
ty for the explanation. merged!

PR #780, opened 2024-05-21
Fix print of docker run string when using integer parameter

davidkopp:
Using the `retries` parameter of the healthcheck feature led to the following error:

```
Error: Base exception occured in runner.py

Exception (<class 'TypeError'>): sequence item 21: expected str instance, int found
Run_id (<class 'uuid.UUID'>): fb0156a1-df46-42b9-85fc-6a94eb6f157b
```

This PR fixes that by converting the `retries` parameter to a string.

While inspecting the issue I'm noticed that the same error could occur when using a single port definition ([ephemeral ports](https://docs.docker.com/guides/docker-concepts/running-containers/publishing-ports/#publishing-to-ephemeral-ports)). I think it makes sense to forbid using single port definitions.

### DIFF
--- a/lib/schema_checker.py
+++ b/lib/schema_checker.py
@@ -97,7 +97,7 @@ class SchemaChecker():
                     Optional("build"): Or(Or({str:str},list),str),
                     Optional("networks"): self.single_or_list(Use(self.contains_no_invalid_chars)),
                     Optional("environment"): self.single_or_list(Or(dict,str)),
-                    Optional("ports"): self.single_or_list(Or(str, int)),
+                    Optional("ports"): self.single_or_list(str),
                     Optional("depends_on"): Or([str],dict),
                     Optional("healthcheck"): {
                         Optional('test'): Or(list, str),
--- a/runner.py
+++ b/runner.py
@@ -859,7 +859,7 @@ class Runner:
                         docker_run_string.append(service['healthcheck']['timeout'])
                     if 'retries' in service['healthcheck']:
                         docker_run_string.append('--health-retries')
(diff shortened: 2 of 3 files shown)


ArneTR:
Thanks for bringing that up.

I think this patch is helpful, but might not cover the whole issue that you uncovered.

From a quick look into the code it is related to how `subrprocess` is treating the list which holds the pieces of the command to execute.

Another item that might be a number and in the end not understood as a string is `container_name`.

Let me give this a more thourough look in the next days as I believe it makes more sense to convert all elements to string before handing it to `subprocess`

2. The `ports` argument I think can be patched in that form although the explanation does not seem to be valid. According to the docker docs it can hold just a single port definition (https://docs.docker.com/compose/compose-file/05-services/#ports)

An unassigned port on the host will then be assigned. Pretty weird behaviour, but seems at least up to specs :)

davidkopp:
I forgot to mention the line that caused the exception:
https://github.com/green-coding-solutions/green-metrics-tool/blob/516584821bbf74d82c58dbdf0cbbf6b0c6cf23fd/runner.py#L933

Regarding the ports: I thought that this weird behavior of assigning a random port on the host doesn't make sense in the context of GMT. Of course, if GMT wants to be compliant with the Docker Compose specification it should be supported.

ArneTR:
I have superseeded this PR with https://github.com/green-coding-solutions/green-metrics-tool/pull/785 and will close it.

Please give https://github.com/green-coding-solutions/green-metrics-tool/pull/785 a quick review then it will be merged.

PR #783, opened 2024-05-22
Improve Healthcheck with start-interval parameter

davidkopp:
As discussed in #781 this PR implements the healthcheck parameter `start-interval`. In addition to that the logging around the healthcheck functionality is improved.

After some testing I came to the conclusion, that it is sufficient to just check every second what the health of the dependent container is without any additional logic.
There are 3 possible outcomes:
- container health becomes `healthy` -> pass
- container health becomes `unhealthy` -> exception
- configured max waiting time is reached -> exception

As proposed in #781 the default max waiting time should be increased (20 seconds is too short).

A compose file using the `start-interval` parameter can be found [here](https://github.com/t2-project/devops/blob/main/energy-tests/gmt/microservices-compose.yml). Currently, I have defined the healthcheck configurations with the following parameters:
```yaml
start_period: "30s"
start_interval: "1s"
interval: "1h"
```

### DIFF
--- a/tests/data/slow-start-application/start.sh
+++ b/tests/data/slow-start-application/start.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+echo "Starting up..."
+sleep 3 # simulting slow start up ...
+echo "Finished startup."
+tail -f /dev/null
--- a/tests/data/usage_scenarios/healthcheck_using_start_interval.yml
+++ b/tests/data/usage_scenarios/healthcheck_using_start_interval.yml
@@ -1,5 +1,5 @@
 ---
-name: Test depends_on
+name: Test Healthcheck
 author: David Kopp
 description: test
 
@@ -10,10 +10,12 @@ services:
       test-container-2:
         condition: service_healthy
   test-container-2:
-    image: alpine
+    build:
+      context: ../slow-start-application
     healthcheck:
-      test: ls
-      interval: 1s
+      test: pgrep 'tail'
+      start_interval: 1s
+      start_period: 5s
 
 flow:
   - name: dummy
--- a/tests/data/usage_scenarios/healthcheck_using_start_interval.yml
+++ b/tests/data/usage_scenarios/healthcheck_using_start_interval.yml
@@ -16,6 +16,7 @@ services:
       test: pgrep 'tail'
       start_interval: 1s
       start_period: 5s
+      interval: 1h # should not influence the measurement
 
 flow:
   - name: dummy
--- a/lib/schema_checker.py
+++ b/lib/schema_checker.py
@@ -105,7 +105,7 @@ class SchemaChecker():
                         Optional('timeout'): str,
                         Optional('retries'): int,
                         Optional('start_period'): str,
-                        # Optional('start_interval'): str, docker CLI does not support this atm
+                        Optional('start_interval'): str,
                         Optional('disable'): bool,
                     },
                     Optional("setup-commands"): [str],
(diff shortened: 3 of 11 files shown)


ArneTR:
The implementation looks sound.

It would be quite nice to see a test that actually validates if the feature is working as expected.

In my opinion the follow test layout would be helpful:
- A container that becomes healthy after 3 seconds. Then a test that will setup interval to 10s and does not set start_interval at all. After 5s the test will faill since the container did not update the health status
- A container that also becomes healthy after 3 seconds. This time start_interval is set to 1s and the container will become healthy after 3s. The test will have a timeout of 5s that will however never trigger as the test will finish before. This can be reached by using the `run_until` to bring GMT just before the container boot and then executing the boot phase separately.

A container that matches this specification needs to set a healthcheck command, otherwise it will automatically be detected as healthy as soon as it has started.

Here is a sample Dockerfile that would create such a container:

```docker
# Use an official lightweight image as a base
FROM alpine:latest

# Add a startup script
COPY start.sh /start.sh

# Make the script executable
RUN chmod +x /start.sh

# Add a health check
HEALTHCHECK CMD [ "sh", "-c", "ps aux | grep '[s]leep 5'" ]

# Run the startup script
CMD ["/start.sh"]

```

`start.sh`
```bash
#!/bin/sh
echo "Starting up..."
sleep 5
echo "Finished startup."
```

This code is untested but the concept should work ... maybe with minor tweaks.

## Summary

If you can implement a test, maybe even a better one if you can think of one, it would be very helpful.

If you cannot find the time we will try to take this on in the next two weeks.

davidkopp:
I have now improved the tests for the healthcheck functionality in general.
Thanks for your explanations for a good test setup. There were some changes necessary, so please have a look if I have covered what you wanted.

ArneTR:
Running the tests and giving it a look now

tests/data/slow-start-application/start.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+echo "Starting up..."
+sleep 3 # simulting slow start up ...
+echo "Finished startup."
+tail -f /dev/null
```
ArneTR, line 5: Interesting way to keep the container open indefinitely ❤️

tests/data/usage_scenarios/healthcheck_using_start_interval.yml
```diff
@@ -0,0 +1,25 @@
...
+    depends_on:
+      test-container-2:
+        condition: service_healthy
+  test-container-2:
+    build:
+      context: ../slow-start-application
+    healthcheck:
+      test: pgrep 'tail'
+      start_interval: 1s
+      start_period: 5s
```
ArneTR, line 18: In order to differentiate from `interval` here, which will also run to my understanding by default, you should force `interval` to a unreasonably high value of 1h or similar to disable it effectively

ArneTR (changes requested):

ArneTR:
Looks quite nice! Made a review and had a small change request

tests/data/usage_scenarios/healthcheck_using_start_interval.yml
davidkopp: done.

davidkopp:
Added another test with `start_interval` defined, however without `start_period` -> no healthcheck gets executed. If you want to use `start_interval`, it's necessary to also define `start_period`.

ArneTR:
sweet. thank you again so much for contributing!

Merged!!

PR #799, opened 2024-05-31
Adds a script to remove carbonDB duplicates

ribalba:
I went for a KISS version which needs to be called via `cron`. As we will be reworking carbonDB soonish I didn't want to do something overly complex like copying DBs around. We can call this every `n` minutes. We could consider setting a flag in redis if something has changed but I think we should think about this bigger. Maybe we need a general queuing system like jobs. But more flexible. Not something I see vital for now though

### DIFF
--- a/api/api_helpers.py
+++ b/api/api_helpers.py
@@ -778,13 +778,3 @@ def carbondb_add(client_ip, energydatas):
     columns = ['type', 'company', 'machine', 'project', 'tags', 'time_stamp', 'energy_value', 'co2_value', 'carbon_intensity', 'latitude', 'longitude', 'ip_address']
 
     DB().copy_from(file=data_file, table='carbondb_energy_data', columns=columns, sep='|')
-
-# TODO: Fix dedupliation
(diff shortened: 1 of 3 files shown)


ArneTR (changes requested):
My remarks:
- How often do you estimate this script has to run?
- Can you please fix the filename, it contains a typo
- The query takes at the moment already > 20 Minutes (it is still running ...).  If feels like this is a horrible inefficient operation. What is it trying to achieve by looking at all the data? Can we not close the window of acceptance in CarbonDB by let's say accepting at max 30 day old data? Then the window for duplicates can be narrowed?
- Further adding to my point before: The query sets a lot of locks. This will block the service. Even if running at night I think this is too intense. See screenshot
Screenshot 2024-06-01 at 10 26 39 AM

ArneTR:
Update on this: The script cannot finish in it's current form. Taking longer than 12 hours ... then I stopped.

ArneTR:
Looks good

PR #819, opened 2024-06-21
Kaniko enable multi-stage builds

ArneTR:
Kaniko could not create multi-stage images as the filesystem was mounted read only.

It will place images in the `/workspace` folder, which seems like a pre-set directory.

This PR mounts the repo into a `/tmp/repo` folder and mitigates this issue thus allowing multi stage builds

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -579,13 +579,15 @@ class Runner:
                 self.join_paths(context_path, dockerfile)
 
                 docker_build_command = ['docker', 'run', '--rm',
-                    '-v', f"{self._repo_folder}:/workspace:ro", # this is the folder where the usage_scenario is!
+                    '-v', '/workspace',
+                    '-v', f"{self._repo_folder}:/tmp/repo:ro", # this is the folder where the usage_scenario is!
                     '-v', f"{temp_dir}:/output",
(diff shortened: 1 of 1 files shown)


runner.py
```diff
@@ -579,13 +579,15 @@ def build_docker_images(self):
                 self.join_paths(context_path, dockerfile)
 
                 docker_build_command = ['docker', 'run', '--rm',
-                    '-v', f"{self._repo_folder}:/workspace:ro", # this is the folder where the usage_scenario is!
+                    '-v', '/workspace',
```
ribalba, line 582: @ArneTR is this valid?

ArneTR: Unsure what you mean ... so I try to describe what the line does. Maybe this will answer the question.

The line `'-v', f"{self._repo_folder}:/tmp/repo:ro", # this is the folder where the usage_scenario is!` is still present, it is just moved one line down and has now a different target

- `-v /workspace` is a an additional line before that that makes docker mount a writeable filesystem at /workspace. Before it was readonly. readonly is fine for single-stage images, but not for multi-stage as knaiko needs to write a cache.

So we now mount the repo the `/tmp/repo` and just copy stuff into `/workspace`

PR #825, opened 2024-06-30
Added DiskIO Reporter

ArneTR:
This PR adds a Disk I/O reporter.

To make it container agnostic the data is pulled directly from the `sysfs`. This needs the IO delegator to be set as specified in the documentation.

I am currently under the assumption that this controller is always active for docker in root mode, but will double check soon.

The approach here is similar to the other cgroup reading metric providers. However while crafting this the idea came up to actually split Input and Output and not supply the value as a cumulated number.

I see here two options:
- Making the reporter something like a base reporter that always reads everything and then only a part of the value gets extracted
- Making a shallow reporter that works like the SDIA or XGBoost one and actually relies on another reporter to be present on the system.

Here I think @ribalba it makes sense to discuss how this was solved in the lm_sensors

### DIFF
--- a/metric_providers/disk/io/cgroup/container/source.c
+++ b/metric_providers/disk/io/cgroup/container/source.c
@@ -1,0 +1,189 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <unistd.h>
+#include <sys/time.h>
+#include <string.h> // for strtok
+#include <getopt.h>
+
+typedef struct container_t { // struct is a specification and this static makes no sense here
+    char path[BUFSIZ];
+    char *id;
+} container_t;
+
+typedef struct disk_io_t { // struct is a specification and this static makes no sense here
+    long int rbytes;
+    long int wbytes;
+} disk_io_t;
+
+
+// All variables are made static, because we believe that this will
+// keep them local in scope to the file and not make them persist in state
+// between Threads.
+// in any case, none of these variables should change between threads
+static int user_id = 0;
+static unsigned int msleep_time=1000;
+
+
+static disk_io_t get_disk_cgroup(char* filename) {
+    long int rbytes = -1;
+    long int rbytes_sum = 0;
+    long int wbytes = -1;
+    long int wbytes_sum = 0;
+
+    FILE * fd = fopen(filename, "r");
+    if ( fd == NULL) {
+        fprintf(stderr, "Error - Could not open path for reading: %s. Maybe the container is not running anymore? Are you using --rootless mode? Errno: %d\n", filename, errno);
+        exit(1);
+    }
+
+    while (fscanf(fd, "%*u:%*u rbytes=%ld wbytes=%ld rios=%*u wios=%*u dbytes=%*u dios=%*u", &rbytes, &wbytes) == 2) {
+        rbytes_sum += rbytes;
+        wbytes_sum += wbytes;
+    }
+    ;
+
+    fclose(fd);
+
+    if(rbytes < 0 || wbytes < 0) {
+        fprintf(stderr, "Error - io.stat could not be read or was < 0.");
+        exit(1);
+    }
+
+    disk_io_t disk_io;
+    disk_io.rbytes = rbytes_sum;
+    disk_io.wbytes = wbytes_sum;
+    return disk_io;
+}
+
+static int output_stats(container_t *containers, int length) {
+
+    struct timeval now;
+    int i;
+
+    gettimeofday(&now, NULL);
+    for(i=0; i<length; i++) {
+        disk_io_t disk_io = get_disk_cgroup(containers[i].path);
+        printf("%ld%06ld %ld %s\n", now.tv_sec, now.tv_usec, disk_io.rbytes+disk_io.wbytes, containers[i].id);
+    }
+    usleep(msleep_time*1000);
+
+    return 1;
+}
+
+static int parse_containers(container_t** containers, char* containers_string, int rootless_mode) {
+    if(containers_string == NULL) {
+        fprintf(stderr, "Please supply at least one container id with -s XXXX\n");
+        exit(1);
+    }
+
+    *containers = malloc(sizeof(container_t));
+    char *id = strtok(containers_string,",");
+    int length = 0;
+
+    for (; id != NULL; id = strtok(NULL, ",")) {
+        //printf("Token: %s\n", id);
+        length++;
+        *containers = realloc(*containers, length * sizeof(container_t));
+        (*containers)[length-1].id = id;
+        if(rootless_mode) {
+            sprintf((*containers)[length-1].path,
+                "/sys/fs/cgroup/user.slice/user-%d.slice/user@%d.service/user.slice/docker-%s.scope/io.stat",
+                user_id, user_id, id);
+        } else {
+            sprintf((*containers)[length-1].path,
+                "/sys/fs/cgroup/system.slice/docker-%s.scope/io.stat",
+                id);
+        }
+    }
+
+    if(length == 0) {
+        fprintf(stderr, "Please supply at least one container id with -s XXXX\n");
+        exit(1);
+    }
+    return length;
+}
+
+static int check_system(int rootless_mode) {
+    const char* check_path;
+
+    if(rootless_mode) {
+        check_path = "/sys/fs/cgroup/user.slice/io.stat";
+    } else {
+        check_path = "/sys/fs/cgroup/system.slice/io.stat";
+    }
+
+    FILE* fd = NULL;
+    fd = fopen(check_path, "r");
+
+    if (fd == NULL) {
+        fprintf(stderr, "Couldn't open io.stat file at %s\n", check_path);
+        exit(127);
+    }
+    fclose(fd);
+    return 0;
+}
+
+int main(int argc, char **argv) {
+
+    int c;
+    int check_system_flag = 0;
+    int rootless_mode = 0; // docker root is default
+    char *containers_string = NULL;  // Dynamic buffer to store optarg
+    container_t *containers = NULL;
+
+    setvbuf(stdout, NULL, _IONBF, 0);
+    user_id = getuid();
+
+    static struct option long_options[] =
+    {
+        {"rootless", no_argument, NULL, 'r'},
+        {"help", no_argument, NULL, 'h'},
+        {"interval", no_argument, NULL, 'i'},
+        {"containers", no_argument, NULL, 's'},
+        {"check", no_argument, NULL, 'c'},
+        {NULL, 0, NULL, 0}
+    };
+
+    while ((c = getopt_long(argc, argv, "ri:s:hc", long_options, NULL)) != -1) {
+        switch (c) {
+        case 'h':
+            printf("Usage: %s [-i msleep_time] [-h]\n\n",argv[0]);
+            printf("\t-h      : displays this help\n");
+            printf("\t-s      : string of container IDs separated by comma\n");
+            printf("\t-i      : specifies the milliseconds sleep time that will be slept between measurements\n");
+            printf("\t-c      : check system and exit\n\n");
+            exit(0);
+        case 'i':
+            msleep_time = atoi(optarg);
+            break;
+        case 'r':
+            rootless_mode = 1;
+            break;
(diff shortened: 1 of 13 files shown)


metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+    long int wbytes = -1;
+    long int wbytes_sum = 0;
+
+    FILE * fd = fopen(filename, "r");
+    if ( fd == NULL) {
+        fprintf(stderr, "Error - Could not open path for reading: %s. Maybe the container is not running anymore? Are you using --rootless mode? Errno: %d\n", filename, errno);
+        exit(1);
+    }
+
+    while (fscanf(fd, "%*u:%*u rbytes=%ld wbytes=%ld rios=%*u wios=%*u dbytes=%*u dios=%*u", &rbytes, &wbytes) == 2) {
```
ribalba, line 40: isn't `%*u` an unsigned int. But the variable is signed.

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+
+// All variables are made static, because we believe that this will
+// keep them local in scope to the file and not make them persist in state
+// between Threads.
+// in any case, none of these variables should change between threads
+static int user_id = 0;
+static unsigned int msleep_time=1000;
+
+
+static disk_io_t get_disk_cgroup(char* filename) {
```
ribalba, line 28: I would make filename a const

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+    char *id = strtok(containers_string,",");
+    int length = 0;
+
+    for (; id != NULL; id = strtok(NULL, ",")) {
+        //printf("Token: %s\n", id);
+        length++;
+        *containers = realloc(*containers, length * sizeof(container_t));
+        (*containers)[length-1].id = id;
+        if(rootless_mode) {
+            sprintf((*containers)[length-1].path,
```
ribalba, line 90: Wouldn't using `snprintf` here make more sense?
```
sprintf((*containers)[length-1].path, BUFSIZ, ....
```

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+                "/sys/fs/cgroup/user.slice/user-%d.slice/user@%d.service/user.slice/docker-%s.scope/io.stat",
+                user_id, user_id, id);
+        } else {
+            sprintf((*containers)[length-1].path,
```
ribalba, line 94: same as above

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+    return 1;
+}
+
+static int parse_containers(container_t** containers, char* containers_string, int rootless_mode) {
+    if(containers_string == NULL) {
+        fprintf(stderr, "Please supply at least one container id with -s XXXX\n");
+        exit(1);
+    }
+
+    *containers = malloc(sizeof(container_t));
```
ribalba, line 80: Don't you need to check the return of malloc? Same below

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+            printf("\t-c      : check system and exit\n\n");
+            exit(0);
+        case 'i':
+            msleep_time = atoi(optarg);
+            break;
+        case 'r':
+            rootless_mode = 1;
+            break;
+        case 's':
+            containers_string = (char *)malloc(strlen(optarg) + 1);  // Allocate memory
```
ribalba, line 164: When is this freed? 
Also what happens if malloc fails.

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+typedef struct disk_io_t { // struct is a specification and this static makes no sense here
+    long int rbytes;
+    long int wbytes;
+} disk_io_t;
+
+
+// All variables are made static, because we believe that this will
+// keep them local in scope to the file and not make them persist in state
+// between Threads.
+// in any case, none of these variables should change between threads
```
ribalba, line 23: Are you using threads?

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+    if ( fd == NULL) {
+        fprintf(stderr, "Error - Could not open path for reading: %s. Maybe the container is not running anymore? Are you using --rootless mode? Errno: %d\n", filename, errno);
+        exit(1);
+    }
+
+    while (fscanf(fd, "%*u:%*u rbytes=%ld wbytes=%ld rios=%*u wios=%*u dbytes=%*u dios=%*u", &rbytes, &wbytes) == 2) {
+        rbytes_sum += rbytes;
+        wbytes_sum += wbytes;
+    }
+    ;
```
ribalba, line 44: ?

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+    } else {
+        check_path = "/sys/fs/cgroup/system.slice/io.stat";
+    }
+
+    FILE* fd = NULL;
+    fd = fopen(check_path, "r");
+
+    if (fd == NULL) {
+        fprintf(stderr, "Couldn't open io.stat file at %s\n", check_path);
+        exit(127);
```
ribalba, line 121: Why `127`?

https://stackoverflow.com/questions/1763156/127-return-code-from

you use `1` everywhere else

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+    int i;
+
+    gettimeofday(&now, NULL);
+    for(i=0; i<length; i++) {
+        disk_io_t disk_io = get_disk_cgroup(containers[i].path);
+        printf("%ld%06ld %ld %s\n", now.tv_sec, now.tv_usec, disk_io.rbytes+disk_io.wbytes, containers[i].id);
+    }
+    usleep(msleep_time*1000);
+
+    return 1;
```
ribalba, line 71: Why do you return a 1 here?

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+
+    *containers = malloc(sizeof(container_t));
+    char *id = strtok(containers_string,",");
+    int length = 0;
+
+    for (; id != NULL; id = strtok(NULL, ",")) {
+        //printf("Token: %s\n", id);
+        length++;
+        *containers = realloc(*containers, length * sizeof(container_t));
+        (*containers)[length-1].id = id;
```
ribalba, line 88: I would copy the id here instead of referencing a pointer. You don't know what will happen with containers_string in the future

Also I would trim the string as someone could specify container_string:

```"a, b"```

File: metric_providers/disk/io/cgroup/container/source.c [hunk 12]
```diff
@@ -0,0 +1,189 @@
...
+    long int rbytes;
+    long int wbytes;
+} disk_io_t;
+
+
+// All variables are made static, because we believe that this will
+// keep them local in scope to the file and not make them persist in state
+// between Threads.
+// in any case, none of these variables should change between threads
+static int user_id = 0;
```
ribalba, line 24: I would probably move this into the main as if the `getuid` is somehow deleted you are assuming root. Or set it to -1 so the program fails

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -0,0 +1,189 @@
...
+    const char* check_path;
+
+    if(rootless_mode) {
+        check_path = "/sys/fs/cgroup/user.slice/io.stat";
+    } else {
+        check_path = "/sys/fs/cgroup/system.slice/io.stat";
+    }
+
+    FILE* fd = NULL;
+    fd = fopen(check_path, "r");
```
ribalba, line 117: couldn't you do?
```
FILE* fd = fopen(check_path, "r");
```

ribalba (changes requested):
My C is a little rusty. Not sure I found everything.

metric_providers/disk/io/cgroup/container/source.c
ArneTR: Fair point. We are never using the user_id though to check for user id but only to construct a path in the sysfs.

I changed it in all the files though as I believe if we use it in the future then you are right and it can be a security bug.

metric_providers/disk/io/cgroup/container/source.c
ArneTR: legacy :/ 
Changed in all files

metric_providers/disk/io/cgroup/container/source.c
ArneTR: normalized to 1 where possible in all files

metric_providers/disk/io/cgroup/container/source.c
ArneTR: there was an idea to do so at the time of writing, thus the warning commentary. atm we don't though.

metric_providers/disk/io/cgroup/container/source.c
ArneTR: %*u means the variable is ignored. The two first numbers mark the offset on the data storage.

What I am reading in is rbytes and wbytes. They are parsed as long int.

Did I understand the question correctly?

metric_providers/disk/io/cgroup/container/source.c
ArneTR: why? Then I cannot pass a pointer but have to pass by value, or?

metric_providers/disk/io/cgroup/container/source.c
ribalba: @ArneTR I couldn't find the free for this malloc

metric_providers/disk/io/cgroup/container/source.c
ribalba: no idea why I wrote this. I had something clever but can't remember

metric_providers/disk/io/cgroup/container/source.c
ArneTR: I don't understand what question you have. Can you elaborate a bit?

metric_providers/disk/io/cgroup/container/source.c
ArneTR: I had the following idea: "Since I am using BUFSIZ as I did in the code we ingested from the MSR reading code, the array is anyway the maximum possible path lenghts".

Now revising that this is bogus for multiple reasins:
- First of all BUFSIZ is not the maximum path size on an OS. it is rather just the maximum buffersize for STDOUT etc. This was taken from some other code where BUFSIZ was used for reading from IO streams, which then also misused the constant for filesystem path constructions
=> Using PATH_MAX now everywhere

- Added snprintf where possible with PATH_MAX as maximum copy length

metric_providers/disk/io/cgroup/container/source.c
ArneTR: done everywhere. Also in your code of the lm_sensors :)

metric_providers/disk/io/cgroup/container/source.c
ArneTR: Yeah, I agree copying is here the better choice. Especially since strtok anyway has modified the string it should be considered no-reuse memory.

Also added some more NUL-Byte terminations which where missing.

2. I would not trim the string. This is a 5-10 line function in C and I would rather consider this a user-submitted-data error. It will fail cause the container cannot be found.

metric_providers/disk/io/cgroup/container/source.c
ArneTR: removed

ArneTR:
@ribalba All done. Quick Re-review please as I changed this also in all other C source files we have

ribalba (approved):
lgtm

ribalba (commented):
LGTM

PR #835, opened 2024-07-12
Security fix for python interpreter replacement

ArneTR:
It was possible to replace the `python3` interpreter at the venv location with a malicious file.

Since we were giving access to the interpreter via a sudoers entry this would open a privilege escalation.

It is recommended to update to the newest version of the GMT and just re-run the `install_linux.sh` or `install_mac.sh` file. The fix will be applied then.


ribalba (approved):
Looks good and what we discussed

PR #836, opened 2024-07-12
Rapl energy filtering check

ArneTR:
RAPL energy filtering register is now checked.

Mechanism is achieved through allowing requests to the hardware_info_root file which has sudoers rights

### DIFF
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -154,14 +154,13 @@ make -C lib/sgx-software-enable
 mv lib/sgx-software-enable/sgx_enable tools/
 rm lib/sgx-software-enable/sgx_enable.o
 
-print_message "Adding hardware_info_root.py to sudoers file"
-PYTHON_PATH=$(which python3)
-PWD=$(pwd)
-echo "ALL ALL=(ALL) NOPASSWD:$PYTHON_PATH $PWD/lib/hardware_info_root.py" | sudo tee /etc/sudoers.d/green-coding-hardware-info
+print_message "Adding python3 lib.hardware_info_root to sudoers file"
+# Please note the -m as here we will later call python3 without venv. It must understand the .lib imports
+# and not depend on venv installed packages
+echo "ALL ALL=(ALL) NOPASSWD:/usr/bin/python3 -m lib.hardware_info_root" | sudo tee /etc/sudoers.d/green-coding-hardware-info
 sudo chmod 500 /etc/sudoers.d/green-coding-hardware-info
 # remove old file name
 sudo rm -f /etc/sudoers.d/green_coding_hardware_info
-
 
 print_message "Setting the hardare hardware_info to be owned by root"
 sudo cp -f $PWD/lib/hardware_info_root_original.py $PWD/lib/hardware_info_root.py
@@ -236,6 +235,8 @@ if [[ $no_python != true ]] ; then
     python3 -m pip install -r metric_providers/psu/energy/ac/xgboost/machine/model/requirements.txt
 fi
 
+# kill the sudo timestamp
+sudo -k
 
 echo ""
 echo -e "${GREEN}Successfully installed Green Metrics Tool!${NC}"
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -23,10 +23,16 @@ no_build=false
 no_python=false
 no_hosts=false
 ask_tmpfs=true
+no_ipmi=false
+no_sensors=false
+# The system site packages are only an option to choose if you are in temporary VMs anyway
+# Not recommended for classical developer system
+system_site_packages=false
+no_msr_tools=false
 
 reboot_echo_flag=false
 
-while getopts "p:a:m:nhtb" o; do
+while getopts "p:a:m:nhtbisyr" o; do
     case "$o" in
         p)
             db_pw=${OPTARG}
@@ -48,6 +54,19 @@ while getopts "p:a:m:nhtb" o; do
             ;;
         t)
             ask_tmpfs=false
+            ;;
+        i)
+            no_ipmi=true
+            ;;
+        s)
+            no_sensors=true
+            # currently unused
+            ;;
+        r)
+            no_msr_tools=true
+            ;;
+        y)
+            system_site_packages=true
             ;;
 
     esac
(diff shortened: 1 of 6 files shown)


install_linux.sh
```diff
@@ -174,19 +197,33 @@ sudo chown root:root $PWD/tools/cluster/cleanup.sh
...
+    print_message "Installing IPMI tools"
+    print_message "Important: If this step fails it means ipmitool is not available on you system"
+    print_message "If you do not plan to use IPMI you can skip the installation by appending '-i'"
+    if lsb_release -is | grep -q "Fedora"; then
+        sudo dnf -y install ipmitool
+    else
+        sudo apt-get install -y freeipmi-tools ipmitool
+    fi
+    print_message "Adding IPMI to sudoers file"
+    echo "ALL ALL=(ALL) NOPASSWD:/usr/sbin/ipmi-dcmi --get-system-power-statistics" | sudo tee /etc/sudoers.d/green-coding-ipmi-get-machine-energy-stat
```
ribalba, line 222: This might be a problem as you are opening up the ipmi-dcmi program to non root users which then could access the Data Center Manageability Interface. As you can add more parameters which are not checked

install_mac.sh
```diff
@@ -106,10 +106,11 @@ source venv/bin/activate
 print_message "Setting GMT in include path for python via .pth file"
 find venv -type d -name "site-packages" -exec sh -c 'echo $PWD > "$0/gmt-lib.pth"' {} \;
 
-print_message "Adding hardware_info_root.py to sudoers file"
-PYTHON_PATH=$(which python3)
-PWD=$(pwd)
-echo "ALL ALL=(ALL) NOPASSWD:$PYTHON_PATH $PWD/lib/hardware_info_root.py" | sudo tee /etc/sudoers.d/green_coding_hardware_info
+print_message "Adding python3 lib.hardware_info_root to sudoers file"
+echo "ALL ALL=(ALL) NOPASSWD:/usr/bin/python3 -m lib.hardware_info_root" | sudo tee /etc/sudoers.d/green_coding_hardware_info
```
ribalba, line 110: Why are you not checking for permissions here?

lib/hardware_info.py
```diff
@@ -148,11 +92,6 @@ def get_list():
 
     return linux_info_list
 
-
-def get_values(list_of_tasks):
-    '''Creates an object with all the data populated'''
-    return {x[1]: x[0](*x[2:]) for x in list_of_tasks}
-
 def get_default_values():
```
ribalba, line 95: For consistency I would also move this

ribalba (changes requested):

install_linux.sh
```diff
@@ -154,15 +177,15 @@ make -C lib/sgx-software-enable
...
 
-print_message "Adding hardware_info_root.py to sudoers file"
-PYTHON_PATH=$(which python3)
-PWD=$(pwd)
-echo "ALL ALL=(ALL) NOPASSWD:$PYTHON_PATH $PWD/lib/hardware_info_root.py" | sudo tee /etc/sudoers.d/green-coding-hardware-info
+print_message "Adding python3 lib.hardware_info_root to sudoers file"
+# Please note the -m as here we will later call python3 without venv. It must understand the .lib imports
+# and not depend on venv installed packages
+echo "ALL ALL=(ALL) NOPASSWD:/usr/bin/python3 -m lib.hardware_info_root" | sudo tee /etc/sudoers.d/green-coding-hardware-info
+echo "ALL ALL=(ALL) NOPASSWD:/usr/bin/python3 -m lib.hardware_info_root --read-rapl-energy-filtering" | sudo tee -a /etc/sudoers.d/green-coding-hardware-info
```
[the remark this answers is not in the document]
ArneTR: Unsure what you mean here. If parameters are added after the called binary it restrics to exactly these parameters.

If you want to allow different calls with multiple parameters you have to add them all one-by-one or create a wildcard. All parameters are only allowed if I only specify the binary.

Did you remember it differently? Or did I misunderstand the question?

install_linux.sh
```diff
@@ -49,6 +55,19 @@ while getopts "p:a:m:nhtb" o; do
...
+        s)
+            no_sensors=true
+            # currently unused
+            ;;
+        r)
+            no_msr_tools=true
+            ;;
+        y)
+            system_site_packages=true
+            ;;
```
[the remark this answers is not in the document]
ArneTR: the reason is that the default call shall be as short as possible, aka no parameters at all needed.

If you want to deviate from that you can add parameters. -y for instance for the system packages is a very untypical call. Same as not installing IPMI.

I understand that it changes true/false logic, but I see only very selective cases for even using one parameter at all ...

install_linux.sh
```diff
@@ -174,19 +197,33 @@ sudo chown root:root $PWD/tools/cluster/cleanup.sh
...
-    sudo dnf -y install ipmitool
-else
-    sudo apt-get install -y freeipmi-tools ipmitool
-fi
 
+if [[ $no_msr_tools != true ]] ; then
+    print_message "Installing msr-tools"
+    print_message "Important: If this step fails it means msr-tools is not available on you system"
+    print_message "If you do not plan to use RAPL you can skip the installation by appending '-r'"
+    if lsb_release -is | grep -q "Fedora"; then
```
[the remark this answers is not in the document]
ArneTR: jap, interesting. As soon as we find somebody with a working centOS setup this person can tackle it.

Opened an issue for it: https://github.com/green-coding-solutions/green-metrics-tool/issues/842

install_linux.sh
ArneTR: no, this is not how sudoers entries work .... see also the other comment. Can you post a link to your understanding of the sudoers entry? Mine is: As soon as you specify a paramters it is locked to exactly this call.

install_mac.sh
ArneTR: That is done in a different PR

lib/hardware_info.py
ArneTR: why, it is unique to this file?

install_mac.sh
ArneTR: maybe it will become cleaner when I rebase this PR with the now updated main branch ...

install_linux.sh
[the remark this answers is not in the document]
ribalba: I meant to just change the naming of the variables.

install_linux.sh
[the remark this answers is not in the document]
ribalba: You are right. I got mixed up with the `-m`

install_linux.sh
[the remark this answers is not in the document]
ArneTR: understood. changed.

PR #860, opened 2024-08-13
Fix using depends_on in conjunction with container_name

### LINKED ISSUE #859: State and health check does not work if container_name is used
At the moment GMT uses the provided service name to check if the container is running/healthy. However, if a `container_name` is set, the check (that uses the service name) can't find the container and will fail after the configured max waiting time (60 s).

I plan to create a PR this evening that fixes this.

davidkopp:
If `container_name` was used in a setup with `depends_on` the state and health check did not work. This PR fixes that by using the provided `container_name` in the check.

I also tried to improve the wording in the log messages regarding the terms "service" and "container".

Fixes #859

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -703,15 +703,15 @@ class Runner:
 
             if 'container_name' in service:
                 container_name = service['container_name']
+                print(TerminalColors.HEADER, '\nSetting up container (', container_name, ') for service:', service_name, TerminalColors.ENDC)
             else:
                 container_name = service_name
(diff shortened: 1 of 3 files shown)


ArneTR:
Thanks for the PR!

I have seen the issue https://github.com/green-coding-solutions/green-metrics-tool/issues/859 and understand that the healthcheck is currently not working if both `container_name` and a `service` name is specified.

Implementation looks sound to me. Also renaming the variable to dependent_service actually makes clear what the array contains in accordance with the docker compose specifications.

Running the tests now!

PR #869, opened 2024-08-21
Checks for duplicate flow names

ribalba:
Flow names need to be unique. This is something we check while executing the flow but it will fail quite late, once the second flow name is encountered. This adds a check when parsing the yml file


ArneTR:
Good idea. I would however also output the name which was not unique to make the error clear. I will add this in main as an additional commit after merge

PR #872, opened 2024-08-25
Authentication in GMT

ArneTR:
This PR introduces the authentication functionality to GMT.

Authentication serves following functional needs:
- Restricting certain API routes or optimization features to only authenticated users
- Allowing certain authentication tokens unlimited measurement minutes while as restricting others / public to have limited measurement minutes
- Mapping certain runs to specific authentication tokens and thus allowing for longer data rentention
- Mapping submitted runs, ci_measurements and PowerHog data to certain authentication tokens and thus creating CarbonDB overviews 

## Design considerations
GMT is designed to be a fully self-hostable and open source application. In many cases no authentication is needed for open source teams using thus.
By default GMT will install a "DEFAULT" user which however has access to all ressources, has no quotas or limits.
The authentication is designed to be transparent when not configured meaning that the frontend as well as the CLI interface will transparently inject the "DEFAULT" user whenever accessing ressources.

Only when the system is updated, new machines are added or the code is modified the authentication token ACLs must be altered.

This is the current data structure of the ACL

```JSON
{
    "api": {
        "quotas": {}, # Empty object means no quotas are applied
        "routes": [
            "/v1/carbondb/add",
            "/v1/ci/measurement/add",
            "/v1/software/add",
            "/v1/hog/add",
            "/v1/authentication/data"
        ]
    },
    "data": {
        "ci_measurements": {
            "retention": 2678400  # 31 days
        },
        "hog_coalitions": {
            "retention": 2678400  # 31 days
        },
        "hog_measurements": {
            "retention": 2678400  # 31 days
        },
        "hog_tasks": {
            "retention": 2678400  # 31 days
        },
        "measurements": {
            "retention": 2678400  # 31 days
        },
        "runs": {
            "retention": 2678400 # 31 days
        }
    },
    "jobs": {
        "schedule_modes": [
            "one-off",
            "daily",
            "weekly",
            "commit",
            "variance"
        ]
    },
    "machines": [ # List will be filled automatically with future script
        1
    ],
    "measurements": {
        "quotas": {},  # Empty object means no quotas are applied
        "settings": {
            "flow-process-duration": 86400, # 24h
            "total-duration": 86400 # 24h
        }
    },
    "optimizations": [ # List will be filled automatically with future script
        "container_memory_utilization",
        "container_cpu_utilization",
        "message_optimization",
        "container_build_time",
        "container_boot_time",
        "container_image_size"
    ]
}
```

## Missing features
- At the moment this alteration can only be done in the database directly. For the future it is planned to create a "Create User / Update User" script that can handle this token update transparently.
- ~~Measurement quotas are not applied yet. Only API quotas so far~~
- ~~data retention limit is not implemented yet~~


ArneTR:
- Data Retention was implemented. Cronjob can now be done via `prune_db.py` script
- Measurement quota was implemented
- Test for functionality added

frontend/js/authentication.js
```diff
@@ -0,0 +1,33 @@
...
+    $('#save-authentication-token').on('click', async function(){
+        localStorage.setItem('authentication_token', $("#authentication-token").val());
+        try {
+            $('#token-details-message').hide();
+            var user_data = await makeAPICall('/v1/authentication/data');
+
+            $('#token-details-message').show();
+            $('#token-details').text(JSON.stringify(user_data.data, null, 2));
+        } catch (err) {
+            showNotification('Could not read authentication token data', err);
```
ribalba, line 29: I would think about deleting the token here. Otherwise the user might have added a wrong token, got the error and navigated somewhere else and the frontend stops working because the auth keeps failing.

frontend/js/helpers/main.js
```diff
@@ -149,9 +153,14 @@ async function makeAPICall(path, values=null) {
             }
         }
     }  else {
-        var options = { method: 'GET' }
+        var options = { method: 'GET', headers: {} }
     }
 
+    let authentication_token = localStorage.getItem('authentication_token');
+    if(authentication_token == null) authentication_token = 'DEFAULT';
```
ribalba, line 160: I would not set a token default value as this might confuse people when debugging. As as far as I understand the backend it defaults to `DEFAULT` if there is no header.

api/main.py
```diff
@@ -660,7 +698,10 @@ def validate_measurement_data(data):
     return True
 
 @app.post('/v1/hog/add')
-async def hog_add(measurements: List[HogMeasurement]):
+async def hog_add(
+    measurements: List[HogMeasurement],
+    user: User = Depends(authenticate), # pylint: disable=unused-argument
+    ):
```
ribalba, line 704: The way this is currently implemented if there is no header present the default user will be user. We need to make sure that data is not shared per accident. The same with carbonDB. Let's say I start the hog and don't add a key all my process data will be uploaded with the default user. This will happen with everyone, so we need to make sure that the data can not be retrieved for the default user. Or the default user can't access the hog and carbondb add api endpoints.

lib/db.py
```diff
@@ -68,6 +68,17 @@ def fetch_one(self, query, params=None, fetch_mode=None):
     def fetch_all(self, query, params=None, fetch_mode=None):
         return self.__query(query, params=params, return_type='all', fetch_mode=fetch_mode)
 
+    def import_csv(self, filename):
+        with self._pool.connection() as conn:
+            conn.autocommit = True
+            cur = conn.cursor()
+            with open(filename, 'r', encoding='utf-8') as sql_file:
+                sql_script = sql_file.read()
+                for statement in sql_script.split(';'):
```
ribalba, line 77: This will fail if there is a `;` in a text value for example

lib/db.py
```diff
@@ -68,6 +68,17 @@ def fetch_one(self, query, params=None, fetch_mode=None):
...
+                    if statement.strip():
+                        cur.execute(statement)
```
ribalba, line 79: - What happens if there is an exception? I would rollback on exception.
- Where do you commit?

lib/secure_variable.py
```diff
@@ -0,0 +1,32 @@
...
+        return self._value
+
+class SecureVariableEncoder(json.JSONEncoder):
+    def default(self, o):
+        if isinstance(o, SecureVariable):
+            return repr(o)
+        return super().default(o)
+
+
+if __name__ == '__main__':
```
ribalba, line 23: I would put this in a test

lib/user.py
```diff
@@ -0,0 +1,178 @@
...
+        if token is None or token.get_value() is None:
+            sha256_hash.update("DEFAULT".encode('UTF-8'))
+        else:
+            sha256_hash.update(token.get_value().encode('UTF-8'))
+
+        user = DB().fetch_one("""
+                SELECT id, name
+                FROM users
+                WHERE token = %s
+                """, params=((sha256_hash.hexdigest()), ))
```
ribalba, line 80: We need to make sure that every SQL connection is encrypted otherwise the tokens are exposed

api/main.py
```diff
@@ -1443,5 +1496,17 @@ async def carbondb_get_company_project_details(cptype: str, uuid: str):
 
     return ORJSONResponse({'success': True, 'data': data})
 
+# @app.get('/v1/authentication/new')
+# async def get_authentication_token(name: str = None):
+#     if name is not None and name.strip() == '':
+#         name = None
+#     return ORJSONResponse({'success': True, 'data': User.get_new(name)})
```
ribalba, line 1503: This will fail if the DB insert fails but still report `'success': True`

ribalba (commented):
Looks good. The only major thing I could find is that currently you can upload hog data and carbonDB that will be saved under the `default` user and then everyone can access that data as the frontend also defaults the open user. So if two people don't add credentials to the hog both can see the others data. I would block the URLs for the default users and force people to auth. 

Otherwise very nice PR

frontend/js/authentication.js
ArneTR: Jap, that would be very confusing. Changed it so that the DEFAULT is now only applied in the API. Not in the Users.py class and not in the frontend anymore. This is clearer.
Also submitting an empty string is now disallowed.

I also added a unit test

frontend/js/helpers/main.js
ArneTR: also resolved. See other comment

api/main.py
ArneTR: But why? This was the behaviour before ...? If you do not use a token your data will get pooled. I think this is perfectly fine ...?

Happy for your feedback. In any case it is nothing to be done in code. This is done via the DB data

lib/db.py
ArneTR: This sadly cannot be done. I must set the adapter in autocommit mode. Otherwise it cannot execute this block statement.

lib/secure_variable.py
ArneTR: Whenever I do these mini codes I see them more as a tooling and manual evaluation if in question. I would keep it in for now and remove later if it becomes overwhelming

api/main.py
ribalba: No, before every instance generated a unique UUID that was used to identify a user.

lib/db.py
ArneTR: hmm, ja. That is an issue. It worked fine for my testing but definitely has to be reworked. Since I do not use that function anymore as I choose a different implementation to ingest the data I added a NotImplementedError Guard clause for now

lib/user.py
ArneTR: Let's have a separate discussion how far we want to go here. Technically GMT itself has no SSL encryption to begin with. It is only there because we have Cloudflare in front of it.
Also we have no in-memory encryption which is also something that could be helpful ...

Moved to a discussion: https://github.com/green-coding-solutions/green-metrics-tool/discussions/918

Please follow up there

api/main.py
ArneTR: that code is not ready and just for future refernce. Added your comment thoug. ty

PR #873, opened 2024-08-25
Trims values from requests submit and allows value addition through url

ribalba:


ArneTR:
I do not understand the use case for this. When will the user be landing on such a prepared URL? Through a direct link? Or is this a feature of the admin atm to produce these links?

ribalba:
a) I want to save the url so I can resubmit a cluster job. A pita at the moment, that I have to reenter the data all the time. I am currently saving a curl request.
b) I want to extend the stats.html that if a job has failed on the cluster you can resubmit it. Once you have fixed the error. But I don't want it to go to the API directly as I don't want people to flood the cluster. Like this they are redirected to the submit form but with all the data filled. (Except email)

ArneTR:
Thanks for the clarification. Makes now total sense to me. Will review now

frontend/js/request.js
```diff
@@ -1,4 +1,32 @@
 (async () => {
+
+    function populateFieldsFromURL() {
+        const urlParams = new URLSearchParams(window.location.search);
+
+        if (urlParams.has('name')) {
+            document.querySelector('input[name="name"]').value = urlParams.get('name');
```
ArneTR, line 7: All parameters need to be XSS escaped

ArneTR (changes requested):

frontend/js/request.js
ribalba: I thought about this, but my way of thought was that all values are added as values of inputs. So there is no real way this could be exploited. But now revisiting it there isn't really and disadvantage of escaping and it can only help.

ribalba: Added escapes

PR #877, opened 2024-08-26
Carbondb unique

ArneTR:
The constraint for the uniqueness in CarbonDB was not looking at all relevant rows.

We had this already implemented on our live system but not backported it to the repository.

The implementation so far was however not respecting NULL values correctly. The new unique index now treats two NULL values as identical and will trigger now.

@ribalba Data volume is now down from 13 GB to 9 MB

### DIFF
--- a/docker/structure.sql
+++ b/docker/structure.sql
@@ -365,7 +365,6 @@ CREATE TABLE carbondb_energy_data_day (
     co2_sum FLOAT,
     carbon_intensity_avg FLOAT,
     record_count INT,
-
     created_at timestamp with time zone DEFAULT now(),
     updated_at timestamp with time zone
 );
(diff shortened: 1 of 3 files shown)


docker/structure.sql
```diff
@@ -378,10 +377,7 @@ CREATE TRIGGER carbondb_energy_data_day_moddatetime
 CREATE INDEX idx_carbondb_hour_company ON carbondb_energy_data_day(company);
 CREATE INDEX idx_carbondb_hour_machine ON carbondb_energy_data_day(machine);
 CREATE INDEX idx_carbondb_hour_project ON carbondb_energy_data_day(project);
-
-
-ALTER TABLE IF EXISTS public.carbondb_energy_data_day
-    ADD CONSTRAINT unique_machine_project_date UNIQUE (machine, date);
+CREATE UNIQUE INDEX unique_entry ON carbondb_energy_data_day (type, company, machine, project, tags, date) NULLS NOT DISTINCT;
```
ribalba, line 380: Why are you adding eh company and the project? Initially the idea was that you only need a machine. But I think we are going to rework this anyway

ribalba (commented):
We should talk about how to change this. After we have auth we can implement this totally different and far nicer. I only used uuid because we didn't have auth. I think this should be refactored

PR #886, opened 2024-09-09
Adds checks to the install scripts

ribalba:
I don't have a Windows system to test this on. But apparently WSL returns  "Linux" in WSL 1 and WSL 2


ArneTR:
Haha, that was quick! Probably following up on the issue we currently have open :)

LGTM

PR #897, opened 2024-09-15
Hog upload fix

ArneTR:
This is currently a draft PR to make the [PowerHog](https://github.com/green-coding-solutions/hog) data upload usable again.

It was deactivated as we were seeing regular errors.

It has now been deactivated with some learnings. More are to come in this PR.

## Learnings
- PowerHog is posting often frames larger than 1 MB
  - NGINX POST body size was increased to 3 MB. Not too large to block the server, but a bit of headroom
- For reasons unknown many requests where issued to `GET /v1/hog/add`. Since we monitor 404s this was clogging up our logs. I have deactivated the route for now ... but why is this happening?

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -678,8 +678,7 @@ async def hog_add(measurements: List[HogMeasurement]):
         try:
             _ = Measurement(**measurement_data)
         except (ValidationError, RequestValidationError) as exc:
-            print('Caught Exception in Measurement()', exc.__class__.__name__, exc)
-            print('Hog parsing error. Missing expected, but non critical key', str(exc))
+            error_helpers.log_error('Caught Exception in Measurement(), but not critical', exception_class=exc.__class__.__name__, exception=exc)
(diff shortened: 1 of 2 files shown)


ribalba:
- Large uploads:
We limit the upload to 10 rows [0] which I found to be under 1MB. This could have increased over time the more apps are running. Setting the upload to  3MB seems like a good solution

- Access to `GET /v1/hog/add`. 
I can not explain why this is happening. The upload is with POST [1]. Do you see some correlation in the log files? What IPs are the GET requests coming from?

[0] https://github.com/green-coding-solutions/hog/blob/ea2f5970b8281650ac21687613b5ce098ddf8de4/power_logger.py#L180
[1] https://github.com/green-coding-solutions/hog/blob/ea2f5970b8281650ac21687613b5ce098ddf8de4/power_logger.py#L207

ArneTR:
@ribalba Can you please close this and move code to the EE repository

PR #927, opened 2024-09-30
Added uninstall script

ArneTR:


uninstall.sh
```diff
@@ -0,0 +1,36 @@
...
+set -euo pipefail
+
+echo "Please note that this script does not remove pre-requisites like gcc, docker, curl etc. as it does not want to alter the system state"
+
+function uninstall_python() {
+    python3 -m pip uninstall -y -r requirements.txt
+    python3 -m pip uninstall -y -r requirements-dev.txt
+    python3 -m pip uninstall -y -r docker/requirements.txt
+    python3 -m pip uninstall -y -r metric_providers/psu/energy/ac/xgboost/machine/model/requirements.txt
+}
```
ribalba, line 11: Why don't you just delete the `venv`. We create it and we should remove it. Like this you have an empty venv lying around. If people install things themselves in the venv they should be warned but I would expect that from a venv. Or tell them at the end that they should delete the venv manually with `rm`


Looking a few lines down you do a `rm -rf . ` which will remove the venv folder? Why this this in the first place then?

ArneTR: It will also try to delete the packages as cached system packages. The pip uninstall with active venv might be unnecessary but is just deleting of symlinks at best, so I guess no harm

PR #928, opened 2024-10-01
Ssl

ArneTR:
This PR introduces SSL functionality to GMT.

Previously it was advised to use Cloudflare in front of GMT if using it in a live version. Howver even that needed an origin certificate (https://developers.cloudflare.com/ssl/origin-configuration/origin-ca/)

This PR does:
- remove port 80 binding
- force DB SSL connection (without CA check though as we use snakeoil certs @ribalba I think this is fine ...?)
- make SSL in install procedure


ribalba (approved):
LGTM

PR #929, opened 2024-10-02
Cron dir refactoring

ArneTR:
This PR clears the logic of our directory strucutre quite a bit AND it adds error handling to all cronjobs.

We now have three main dirs:
- cron
- lib
- tools

Under tools I have extracted all parts that are considered libraries in a separate but identically named file in *lib*


ArneTR:
Tests are fine. It is just a weird race condition that I cannot reproduce locally ... :(

ribalba (approved):

PR #957, opened 2024-10-16
Carbondb 2.0 - New

ArneTR:
This PR brings the new version of CarbonDB to the GMT.

Some features are still missing:
- Compress Job not done atm
- Insert blocking on old data not active
- Embodied Carbon / Operational carbon distinction
- Weighted Averaged Carbon Intensity instead of plain average

This PR is thus Draft as we are still looking for inputs from people currently using CarbonDB.

Preview version online at: http://478gtuzwbe72.green-coding.io:9142/carbondb.html

Example Screenshot:

Screenshot 2024-10-16 at 11 10 24 AM


ArneTR:
@gmaze : As discussed via email this is the new proposed version for a more usable Dashboard for CarbonDB.

Really looking forward to your feedback if this is more usable and solves your needs.

ArneTR:
@mrchrisadams : Are you actively using CarbonDB? In any case: If you have a minute would also really love your feedback on this.

In case you are not aware what CarbonDB does: It is an aggregation tool that pulls data from sources like the Green Metrics Tools Runs, Eco-CI, PowerHOG etc. and brings it into an overview dashboard so that you can overview all your IT emissions.

We have activated it in your CI integration for Eco-CI but so far the data was not really displayed well.

mrchrisadams:
hi @ArneTR !

> Are you actively using CarbonDB? In any case: If you have a minute would also really love your feedback on this.

We aren't actively using it at present, but I think I am starting to understand what this dashboard is intended to show now - essentially it would be all the carbon emissions from digital services across your 'digital estate', right? 

i.e. all our production machines, staff devices, use of hosted third party services like github and so on

If that's the case, I can see the value but it's not something we've invested time in buliding ourselves in the way carbondb works. I'm happy to have more of a chat about it later though, because we did these calculations manually for the various systems we operate to come up with the numbers listed in the article below.

https://www.thegreenwebfoundation.org/news/an-experience-report-using-the-tech-carbon-standard/

Anyway, it's ok to drop the data associated with the green web fdn - we are not relying on it at present.

ArneTR:
@ribalba This PR is heavy and I would say quite impossible to fully review by just looking at it. I tried to pre-empt that by adding quite a lot of tests for the new functionality.

However also a lot of refactorings are included which were needed to bring CarbonDB in with less friction and without old baggage.

The only left baggage is the old `/v1/ci/measurement/add` endpoint which I left in and am just migrating data. As soon as we release the next-next Eco-CI version this shall be removed.

@ribalba Please skim read it if you spot something obvious. Would be good if we can go live with it by tomorrow.

ribalba (approved):
So I read all the code but this can not be seen as a code review as I just skimmed it. Looks like you did what we talked about :) 

Great work what a challenge to do all that.

PR #958, opened 2024-10-17
Fixes wrong o vs. O parameter

ribalba:
-o is for output file -O is for optimisations

### DIFF
--- a/metric_providers/cpu/energy/rapl/msr/component/Makefile
+++ b/metric_providers/cpu/energy/rapl/msr/component/Makefile
@@ -1,4 +1,4 @@
-CFLAGS = -o3 -Wall -lm -I../../../../../../lib/c
+CFLAGS = -03 -Wall -lm -I../../../../../../lib/c
 
 metric-provider-binary: source.c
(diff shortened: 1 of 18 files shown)


ArneTR:
Interesting. Your code still has a ZERO (0) instead of a big O, so this is why it fails.

On a test I see sadly zero improvement. With 14k the binary is already as small as it get's it seems

ribalba:
But is binary size really and indication of performance? Couldn't the binary be the same size but use far better CPU optimisations? Or the binary could even be bigger but faster in execution if some debug symbols are left out but a simple loop is flattened.

ArneTR:
If you want to go wild about it maybe. Actually if I compare not in the kB dimension but in Bytes the O3 is *bigger* than the O2.
What happens internally is hard to say as the load of the code is already close to unmeasureable. I love the optimization, but I believe it is largely irrelevant in a bigger picture given we are already below our planned thresshold of being < 1% CPU utilization per reporter.

ArneTR:
Merging in as only a fronted test flakes. Will monitor if this happens more often. Seems like a race condition

PR #960, opened 2024-10-18
Adding tag and commit mode to Timeline Projects scheduling

ArneTR:
This PR implements the currently functionality for scheduling projects to:
- New Commits 
- New Tags

Currently the implementation works for *GitHub*, *GitLab Hosted*  and also *GitLab* self-hosted.

Commits / Tags are checked via the GitHub or GitLab API.

## Side Benefits
Besides enabling the functionality the user is now alerted in the frontend when the repository could not be found, which is a nice side effect.

## Potential drawbacks

GitHub has a strong limit on public API requests of 60/h. We might hit that. Will need to monitor errors.


lib/user.py
```diff
@@ -21,6 +21,11 @@ def __init__(self, user_id: int):
         self._capabilities = user[2]
 
 
+    def to_dict(self):
+        values = self.__dict__.copy()
+        del values['_id']
+        return values
+
     def __repr__(self):
```
ribalba, line 29: couldn't this call `to_dict`?

ribalba (approved):
LGTM

lib/user.py
ArneTR: yes :) will fix

PR #972, opened 2024-10-31
Phase Stats are now also saved if runs fails

ArneTR:


ArneTR:
@ribalba Tests are now fine. The network flaky test will only get fixed once we have the Green Testing Cluster. It is very hard to make it right on the shared testing env from GitHub

ribalba (approved):
Gave it a quick read and makes sense

PR #973, opened 2024-11-01
Monotonic and low value check

ArneTR:
When decreasing the measurement resolution down to the limit of what the provider can handle (< 1mJ for energy for instance) we now check for an underflow in the data.


ribalba (approved):
lgtm

PR #993, opened 2024-11-22
Codespaces merge into main

ArneTR:


ArneTR:
@davidkopp I have merged the codespaces branch, refactored the CGroup reporters and made all changes explicit in the codespaces setup file.

After this merge it should now be possible to just open a codespace on main without any special setting or selecting the "Workshop" template.

If you are still using it feel free to give it a try if it works for you. Feedback would be greatly appreciated!

davidkopp:
That's great!

There is one issue: The added parameter `VHost_Ratio` for the xgboost model does not work:

```sh
TypeError: PsuEnergyAcXgboostMachineProvider.__init__() got an unexpected keyword argument 'VHost_Ratio'
```

ArneTR:
Thanks. I had already addressed that in https://github.com/green-coding-solutions/green-metrics-tool/pull/994 but not merged. Should be working now.

The outputted energy is also now tailored to the Codespaces VM. Which was not the case before.

davidkopp:
Perfect. It works now 😀

PR #995, opened 2024-11-23
Gitignore improvements

davidkopp:
Some minor improvements on how some files are ignored by Git:

- Ignore `sgx_enable` in tools directory (is a copy of the submodule that gets created during install)
- Ignore all output files in `lib/c` (`/detect_cgroup_path.o` was not ignored)
- Because the exclusion of the lmsensors build happens now in `install_shared.py` if the argument `install_sensors` is set to false, the old workaround in `codespace-setup.sh` is not relevant anymore → ignore of `Makefile.bak` is not relevant anymore

### DIFF
--- a/.devcontainer/codespace-setup.sh
+++ b/.devcontainer/codespace-setup.sh
@@ -1,12 +1,6 @@
 #!/bin/bash
 set -euxo pipefail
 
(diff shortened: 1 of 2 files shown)


ArneTR:
Great PR, thank you! Forgot about the hack regarding the lmsensors. Good find

PR #1053, opened 2025-01-19
(Frontend): Clearer split between modular frontend and "Runner"

ArneTR:
GMT has long been comprised of different stuff we cram into one frontend.

For new users it was often hard to understand what GMT exactly is and how the modularity works.

Althoug we still want to keep all components in one application but since in professionaly setups different install types exist:
- Only Eco CI API
- Only runner machines
- Full fledged setup with all API endpoints running also cronjobs etc.

This PR allows for more modularity in the install and introduces a separate name for the component that actually does the measurement and the dashboard.

The Dashboard is renamed to Green Metrics Tool.

The component making the runs is now the MetricsRunner.

The Dashboard is reworked to have Eco CI, PowerHOG, CarbonDB and MetricRunner to be selectable on install.

Futhermore a Homepage is created that shows the different components active with some basic stats.

@ribalba  See the screenshot. Which more basic stats would you like to see?

@davidkopp If you have some spare time I would also really value your opinion. Hope the screenshots suffice to understand the PR

Screenshot 2025-03-15 at 4 29 26 PM
Screenshot 2025-03-15 at 4 29 35 PM

## Greptile Summary

This PR improves the modularity and clarity of the Green Metrics Tool frontend by separating components and introducing more flexible installation options.

- Fixed NVIDIA SMI source name typo in `/frontend/js/helpers/config.js.example` from 'NVIDA SMI' to 'NVIDIA SMI'
- Removed `/tests/test-config.js` in favor of more modular configuration approach
- Added installation flags for ScenarioRunner, Eco CI, PowerHOG, and CarbonDB components
- Renamed Dashboard component to "Green Metrics Tool" and measurement component to "MetricsRunner" for better clarity
- Created new homepage showing active components with basic statistics

The changes make the tool's architecture clearer for new users while maintaining all functionality in a single application with flexible installation options.


davidkopp:
I like the split of the components in the frontend! That makes totally sense.

Here some thoughts regarding the names:
- When I hear the name "Green Metrics Tool" I think about the functionality of running scenarios, not about the whole tool suite. I assume that is the case for the most (or even all) people. "Green Metrics Tool" sounds like it consists of only one tool, not a suite of tools. So "Green Metrics Tool Suite" would be more precise, but I'm not sure if you would like to have a product name change.
- "MetricRunner": Why not "ScenarioRunner" / "Scenario Runner"? Including the term _scenario_ would make it much more intuitive (IMHO), at least with the current functionality. The runner requires a `usage_scenario.yml` file and runs a measurement of an use case scenario. Or do you plan to include the monitor mode into the runner? If so, including "scenario" into the name would be again counterintuitive. Side note: For me the functionality of monitoring a (complex) system could be even a complete different tool of the GMT tool suite, independent of the runner.
- Renaming the page "energy timeline" to "projects": the term "projects" is similar to "repos", so that could be confusing. "Monitored Projects" would be more intuitive, but probably too long.

ribalba:
Somehow this slipped through sorry.

- I am not a big fan of adding the "suite" to the name. It makes it even longer.  I would keep the name Green Metrics Tool and market it as the foundation on which different things build through modules. 

```
+-------+ +-------+ +-------+
|       | |       | |       |
|       | |       | |       |
| POWER | |  ECO  | |Runner |
|  HOG  | |   CI  | |       |
|       | |       | |       |
|       | |       | |       |
+-------+ +-------+ +-------+
+---------------------------+
|                           |
|             GMT           |
|                           |
+---------------------------+
```

I think people will understand that the modules are part of the GMT as it is more than just the frontend but also the database. CarbonDB can not work without the GMT for example. 

- I don't really like `MetricsRunner` as this is confusing as we have the `metrics_provider` so it somehow implies running those. I think something like `ScenarioRunner` is ok. I remember we also talked about a totally different name like `squid` or something that we can say what it means in the GMT context. 

- Maybe you can also add a little `>` to things that pull out. 
- I also wouldn't understand what projects is
- Maybe we should move Auth and Settings to the bottom of the menu to show that these are module independent features?
- Should we remove the Data Analysis point and maybe move it to the documentation?
- I would also highlight the Home a little to make sure that it is the heading
- I really like the overview page. 

I haven't looked at the code

ArneTR:
Thanks for all your great comments.

- I do see the value of reducing confusion about "Green Metrics Tool" to make the "Suite" behaviour clearer, but I also believe adding an Term is too long and a change here is the most critical. So maybe no change for now here.
- I see the problem with projects. Will rename to "Monitored repos" or similar
- Thanks for the feedback on MetricRunner ... something else then should be used I guess :)
  - The term ScenarioRunner feels kind of "matching" but only if you have understood the GMT mindset already I feel. If you do start and do not know what a scenario means here it leaves you very confused.
  - So how about going more simple and just doing: "CodeRunner" ?

davidkopp:
[CodeRunner](https://coderunnerapp.com/) is a Programming Editor for macOS. You should take this in considerations. The name ScenarioRunner is used from the [Carla Simulator](https://github.com/carla-simulator/scenario_runner). However, I think that is less problematic. 

I disagree that CodeRunner is clearer than ScenarioRunner. With CodeRunner you might have the idea that you are running code snippets (e.g. an algorithm), which is an overly simplistic understanding. ScenarioRunner implies that you are running something more, called a "scenario". You may not know exactly what the term "scenario" means in this context. But you will quickly understand and see that the purpose of the tool is to run real use case scenarios, so the term ScenarioRunner makes perfect sense.
I personally prefer the name ScenarioRunner.

ribalba:
I agree with @davidkopp. Code Runner implies that you are running a piece of code. Other ideas `BechmarkRunner` just `Bench`, `Profile`, `ProfileRunner`

ArneTR:
Alright guys, I would like to take one final round on this. Hope you have still some juice to tag along :)

I see the point with CodeRunner and agree that ScenarioRunner is very descriptive. Still it feels like a term that a German agency would use and that makes me feel sad ...

How do you feel about:
- ScenarioEngine
- GreenRunner
- GreenRun
- FooRun (for Footprint-Runner)
- FRun (for Footprint Runner)

ribalba:
... and we don't want a sad Arne 😿

I would vote for ScenarioEngine as everything with Green will be confusing when we explain the Green Metrics Tool (I know I suggested them)

FooRun reminds me of FooBar and [Foo Fighters](https://www.youtube.com/watch?v=1VQ_3sBZEm0)

davidkopp:
I agree with Didi. From the proposed names I prefer ScenarioEngine as well. 

However, ScenarioRunner still feels more fitting than ScenarioEngine for me 🙈
An engine is a core component of something that you usually don't touch as an end user. But from my understand as an end user of the Green Metrics Tool, I use/see this component directly, similar to the other components as well:
- Eco CI: add it explicitly to my pipelines
- PowerHOG: run it explicitly on my desktop machine
- CarbonDB: view the resulting carbon emissions
- Scenario* component: start explicitly a measurement of a usage scenario

Do you have a different understanding of the scenario component than me?

PR #1059, opened 2025-01-24
Github gives you the url for the repo with a .git at the end

ribalba:


ArneTR:
I would sanitize / modify the URL on submission and save the non .git in the DB.

Otherwise you are doing a strreplace here, but then later when we actually pull the repo we do that on a different URL.
Though this works for github and seems fine for a .git URL we might add some more changes later.

Thus my wish: Make operations for checking and pulling on the same sanititzed URL and put the check in the API routine of v1/software/add

ArneTR:
The hack you introduced assumes that GitHub will always treat X and X.git as the same repo. 
This does not need to be the case and can change in the future.

I believe the correct way is to use what was verfied against the API. If .git URLs currently do not verify then we need to sanitize and use the non-.git. This we should then do throughout.
Worst case is that you pull a different repo later than what was verified.

To be fair, this can also happen the other way round. So the safest way would be to just reject .git URLs entirely ...

PR #1067, opened 2025-01-31
Adds-ping-on-install

ribalba:
Will fix the tests and and add some once we have agreed that this is the way we want to implement this. 

I could also imagine that we do it on a per run basis. That would be more interesting data but privacy should be paramount here. 

## Greptile Summary

Adds installation telemetry functionality to send anonymous system information to plausible.io, introducing new privacy considerations and data collection mechanisms.

- Added telemetry functions in `lib/install_shared.sh` to collect machine ID and system info with opt-in prompts
- Added `-Z` flag in `.devcontainer/codespace-setup.sh` and `-z` flag in `.github/actions/gmt-pytest/action.yml` for automated telemetry
- Machine ID collection in `lib/install_shared.sh` reads sensitive system identifiers without fallback options
- Hardcoded analytics endpoint (plausible.io) in `lib/install_shared.sh` lacks error handling
- Missing documentation and global configuration options for telemetry settings

### DIFF
--- a/lib/install_shared.sh
+++ b/lib/install_shared.sh
@@ -24,9 +24,8 @@ ask_ssl=true
 cert_key=''
 cert_file=''
 enterprise=false
-send_ping=true
-send_ping_override=false
-
+ask_ping=true
+force_send_ping=false
 
 function print_message {
     echo ""
@@ -306,18 +305,6 @@ function send_ping() {
         --data "{\"name\":\"install\",\"url\":\"http://hello.green-coding.io/install\",\"domain\":\"hello.green-coding.io\",\"props\":{\"unique_hash\":\"${unique_hash}\",\"arch\":\"${arch}\",\"arch\":\"${os}\",\"os\":\"${os}\",\"os_version\":\"${os_version}\"}}" > /dev/null
 }
 
-function ask_for_ping() {
-    if [[ $send_ping_override == true ]]; then
-        send_ping
-    else
-        echo ""
-        read -p "Developing software can be a lonely business. Want to let us know you are installing the GMT? No personal data will be shared! (y/N) : " send_ping_input
-        if [[ "$send_ping_input" == "Y" || "$send_ping_input" == "y" ]] ; then
-            send_ping
-        fi
-    fi
-}
-
 while getopts "p:a:m:nhtbisyrlc:k:e:zZ" o; do
     case "$o" in
         p)
@@ -371,10 +358,10 @@ while getopts "p:a:m:nhtbisyrlc:k:e:zZ" o; do
             enterprise=true
             ;;
         z)
-            send_ping=false
+            ask_ping=false
             ;;
         Z)
-            send_ping_override=true
+            force_send_ping=true
             ;;
     esac
 done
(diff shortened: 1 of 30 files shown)


ArneTR:
Looks like a good and compact implementation.

I am however not in favor of having the receiver end inside of the GMT. It is not a lot of code, but is has nothing to do there. In a worst case scenario people use it to flood other installations with bogus information because it is a an all-receiving endpoint.

Happy to merge this in when the tests work. Please just send the data to the same endpoint from Plausible as EE does it. We should install a CNAME soon to have this analytics vendor agnostic

ArneTR:
I installed a CNAME

hello.green-coding.io

ribalba:
@ArneTR changed to plausible

ArneTR:
I don't understand the send_ping_override ... why is that needed?

Codespaces will send the ping if no parameter is supplied, as this is the default behaviour. Why supply -Z additionally?

lib/install_shared.sh
```diff
@@ -282,9 +285,36 @@ function finalize() {
...
+    curl -i -X POST https://plausible.io/api/event \
+        -H 'User-Agent: ${random_hash}' \
+        -H 'Content-Type: application/json' \
+        --data "{\"name\":\"install\",\"url\":\"http://hello.green-coding.io/install\",\"domain\":\"hello.green-coding.io\",\"props\":{\"unique_hash\":\"${unique_hash}\",\"arch\":\"${arch}\",\"arch\":\"${os}\",\"os\":\"${os}\",\"os_version\":\"${os_version}\"}}" > /dev/null
+}
 
-while getopts "p:a:m:nhtbisyrlc:k:e:" o; do
+function ask_for_ping() {
+    echo ""
+    read -p "Developing software can be a lonely business. Want to let us know you are installing the GMT? No personal data will be shared! (y/N) : " send_ping_input
```
ribalba, line 311: @ArneTR read is here

ArneTR: I think your code has a bug then. You refactored it to trigger the ask_for_ping function and not send_ping if -Z is set. So it will always ask 

I am in favor of skipping the question with appropriate params. But this is not happening to my understanding atm

ArneTR: no, I stand corrected. Was a misunderstanding

ribalba: I think you are right actually. The check is in the wrong place.

ribalba:
I changed the logic. Good catch

ArneTR:
I have reworked it a little. I believe it is much clearer now, to speak of ask_ping and not send_ping and have the function.

Also I believe having a function called send_ping() and a variable $send_ping was confusing;

Also added some local definitions where applicable

ArneTR:
@greptileai

lib/install_shared.sh
```diff
@@ -282,9 +284,28 @@ function finalize() {
...
+    local unique_hash=$(generate_unique_hash)
+    local random_hash=$(openssl rand -hex 8)
+    local arch=$(uname -m)
+    local os=$(uname -s)
+    local os_version=$(uname -r)
+
+    curl -i -X POST https://plausible.io/api/event \
+        -H 'User-Agent: ${random_hash}' \
+        -H 'Content-Type: application/json' \
+        --data "{\"name\":\"install\",\"url\":\"http://hello.green-coding.io/install\",\"domain\":\"hello.green-coding.io\",\"props\":{\"unique_hash\":\"${unique_hash}\",\"arch\":\"${arch}\",\"arch\":\"${os}\",\"os\":\"${os}\",\"os_version\":\"${os_version}\"}}" > /dev/null
```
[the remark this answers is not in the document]
ribalba: done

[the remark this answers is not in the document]
ribalba: done

ArneTR:
@greptileai

PR #1088, opened 2025-03-06
Add support for entrypoint attribute

### LINKED ISSUE #1085: Support entrypoint in docker compose
Container images often come with a predefined ENTRYPOINT. The usage of such a container image in a flow within GMT is currently not possible, because in a flow we need to be able to execute a command during runtime of the container and not only at the start of the container.

Therefore, common container images of various API/load testing tools can't be used at the moment because of their ENTRYPOINTs. Examples:
- [JMeter](https://github.com/justb4/docker-jmeter/blob/master/Dockerfile): `ENTRYPOINT ["/entrypoint.sh"]`
- [k6](https://github.com/grafana/k6/blob/master/Dockerfile): `ENTRYPOINT ["k6"]`
- [Artillery](https://github.com/artilleryio/artillery/blob/main/packages/artillery/Dockerfile): `ENTRYPOINT ["/home/node/artillery/bin/run"]`
- [grpcurl](https://github.com/datawire/docker-grpcurl/blob/master/Dockerfile): `ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]`

As far as I know at the moment there is only one solution: Build a custom container image without an entrypoint or an entrypoint that is waiting for commands. I did the latter for JMeter 1,5 years ago here:
https://github.com/t2-project/docker-jmeter/blob/master/entrypoint.sh

A better solution could be the support of the attribute `entrypoint` in the Docker Compose file. By specifying an entrypoint here, the entrypoint in the container image can be overwritten. The attribute is supported by [Docker Compose](https://docs.docker.com/reference/compose-file/services/#entrypoint) and by [docker run](https://docs.docker.com/reference/cli/docker/container/run/#options) (`--entrypoint`).

Is this really a solution or have I overlooked something?

davidkopp:
Closes #1085

Only the short form of the `entrypoint` attribute is implemented (see [Docker Compose specification](https://docs.docker.com/reference/compose-file/services/#entrypoint)).
This is also the case for the `command' attribute, where only the short form is currently implemented.

## Greptile Summary

Added support for Docker entrypoint configuration in service definitions, enabling overriding of container ENTRYPOINTs to allow execution of runtime commands in test flows.

- Added `entrypoint` validation in `lib/schema_checker.py` as optional string field in service schema
- Modified `runner.py` to handle entrypoint overrides using secure command parsing
- Replaced `clear-entrypoint` flag with standard empty string syntax (`entrypoint: ''`) 
- Updated test scenarios to use `tail -f /dev/null` instead of resource-intensive `yes` command
- Removed custom non-standard attributes to better align with Docker Compose specification

<sub>💡 (5/5) You can turn off certain types of comments like style [here](https://app.greptile.com/review/github)!</sub>


ArneTR:
Uh, this looks juicy, ty!

1. I will let our GreptileAI check it ... so beware for maybe incoming comments.
2. On first view I believe using shlex.split on the entrypoint may introduce an argument injection. The idea is to have whatever is supplied as entrypoint to be supplied as one contigous string.
=> I will give it a more thorough look tomorrow morning

First and foremost: Big thx!

ArneTR:
@greptileai

davidkopp:
My initial implementation was completely wrong. Sorry for that.
Before marking the PR as ready I would like to do some more tests. For example I would like to set the `entrypoint` to `entrypoint.sh` to be able to execute arbitrary commands instead of just one word.

davidkopp:
I think the PR is now ready for review. I added now 3 unit tests in total to have a good understanding of how the `entrypoint` can be used, also in conjunction with `command`.

I didn't do the changes greptile proposed, because then it would not be consistent with the rest of the code.

ribalba:
Cool idea and great way to KISS it. Something I do see myself wanting to use is a syntax like
```
    entrypoint: tail -f /dev/null
```
so I don't need to create a separate shell file. You did that in [your first version](https://github.com/green-coding-solutions/green-metrics-tool/pull/1088/commits/b309d45b0b507af34f7d8ee087317536922d7960#diff-34cde5e49b54d9b67952c80d3580b6c421ebcf16097996e215aef54e44f78562R1026) I think we should allow scripts and commands in the specification

ribalba:
I was thinking of then changing it to a list internally with the split. But we should keep the standard 

But maybe we should step back a step and think about how this feature would be used. We could still support the std compose entry point for people that want to use that in their compose file. 

But the main usage IMHO will be to keep the docker container running when we are using a container that normally exits after command completion. So maybe we should either:
a) have a std script in the docker container that everyone can use. Something like /tmp/run4ever.sh
b) add a new key to the compose section. Something like `run_container_4_ever: true` which then overwrites the entry point with `['tail', ...`. 
c) Creates a new docker file with the entrypoint changed. Not really in favour of doing this except if we can come up with some other things we might want to add to a new docker container for certain benchmarks or that makes something easier. But I can't think of something currently so I would not implement it like this.

davidkopp:
I like the idea to provide a simple way to keep the container alive.
Option a and b would be both ok for me. Option b, setting an attribute to true in the service definition, would be the easiest way from a user perspective. The only concern I have with this approach is that the values provided as `entrypoint` and `command` must be ignored.
An alternative name for the attribute could be `keep_alive` or `keep_container_alive`.

So which approach do we want to take?

I came across [this statement](https://stackoverflow.com/a/41655546), that suggests to use `sleep` or a while loop with sleep instead of using `tail` for blocking. `sleep infinity` does not work on all systems (e.g. busybox), so maybe a sleep in a while loop would be the best option to implement it.

ribalba:
- We could output a warning if we detect that `entrypoint` is set to make it clear to the user. 
- I like "keep_container_alive" far more than my hastily written idea
- I can't remember why I moved to the `tail` syntax at some stage. I used to use sleep but that failed on some systems. But I can't remember what that was. I would propose to go with the `sleep` for now till we encounter a problem as the `tail` solution does add overhead. 

@ArneTR is currently out so let's wait and see what he has to say about this :) I can imagine him having an opinion.

ArneTR:
Hey Hey, sorry for the late response but I am sadly sick and thus response times are bad ...

1. Long syntax
This is only possible as a workaround as in the end we pass params to the CLI, which has no native support for it. So we need to manually split it up ... I would prefer not to go that route

2. The idea of having a declarative argument like "keep_container_alive" is nice but overloads a lot of stuff whic h is unclear without readin the docs.
Furthermore I am not 100% sure if we can always detect if an entrypoint is set in the Dockerfile. 

3. My proposal would be instead to introduce the overhead of trying to implement entrypoint we just implement an entrypoint "remover".
All the functionality that entrypoint has can be mimicked with just using cmd.
So I propose introducing *clear_entrypoint* and then a user can craft a usage_scenario with a *cmd* command as wanted.
This saves users from re-building their container images, allows to run different commands as needed, and introduces no escaping need or security guarding in GMT.
Only downside I see so far is that the argument is non declarative and a user needs to go to the docs to understand how to keep a container running. We could however catch this state and output a hint ...?

@davidkopp Since you have the feature currently in use: Would that implementation in GMT work for you?

davidkopp:
An entrypoint remover would also work for me.

There are now multiple ways implemented in this PR that would work for me:

1. script:
   ```yaml
   entrypoint: entrypoint.sh
   ```

2. single word entrypoint and arguments provided with the command attribute:
   ```yaml
   entrypoint: tail
   command: -f /dev/null
   ``` 

3. empty entrypoint (ignores default entrypoint):
   ```yaml
   entrypoint: ""
   command: tail -f /dev/null
   ``` 

4. explicit entrypoint remover:
   ```yaml
   clear-entrypoint: true
   command: tail -f /dev/null
   ````    

The ways 1-3 are compliant with the Docker Compose specification, 4 not.

I personally like the option 3 the most. What do you think about it? If you like it as well, we would not need the custom attribute `clear-entrypoint` and we could remove it again.

btw: Option 3 and 4 are implemented internally the same way.

ArneTR:
Oh nice thinking. Yah, option 3 is way nicer and does introduce mental overhead with more attributes.

Happy to merge in variant 3 if you trim the PR down. Can also release it today then.

But first of all: Will run the tests in the current form and call Greptile.

ArneTR:
@greptileai

davidkopp:
I removed the attribute `clear-entrypoint` again and removed one test case, that is not really relevant.

So from my side the PR would be ready to merge 😊

davidkopp:
Ah, one last question: Should there be a check if the entrypoint consists only of a single word? If there are multiple words it will be a problem later on when the container is created with a incorrect docker run command.

ArneTR:
@greptileai

ArneTR:
Code looks great and I will be merged. 

Here my replies to your comments:
- As you noted starting the container will fail if more than one word is in the entrypoint. This is expected for the CLI calling, but is different from how docker compose behaves. I argue that it is fine for now and I would like to introduce this functionality step by step. The PR is already touching some files.
  - For instance, different to how I understood you earlier this code behaves totally fine in Docker Compose:
  ```docker
    name: Stress Container One Core 5 Seconds
    services:
      ubuntu:
        image: ubuntu
        entrypoint: 'echo "asd"'
        command: "what"
  ```
  - It will output "asd what". Docker just merges entrypoint and command, even when entrypoint is just a string and has more than one word. It does not read like that from the docs, but it seems the implemented behaviour.
  - Although I find this quirky, this is they way GMT should also behave I believe ... I will create this in a follow up PR

- Secondly: One major problem I see, not limited to this PR though, is that WHEN only a single word is supplied the error is very misleading:
```log

Error: RuntimeError occured in runner.py

Exception (RuntimeError): PowermetricsProvider returned error message: Metrics provider powermetrics seems to have not produced any measurements. Metrics log file was empty. Either consider having a higher sample rate or turn off provider.
...
``` 

Since multiple exceptions happended in this case Python shows only the last hereby masking what was actually the root cause.

I will add the exception context to make this clearer thus resulting in something like this:

```log

Error: RuntimeError occured in runner.py

Exception (RuntimeError): PowermetricsProvider returned error message: Metrics provider powermetrics seems to have not produced any measurements. Metrics log file was empty. Either consider having a higher sample rate or turn off provider.
Initial_exceptions (OSError): Docker run failed
Stderr: docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "sleep asd": executable file not found in $PATH: unknown.

```

davidkopp:
Thanks for merging!

> Docker just merges entrypoint and command, even when entrypoint is just a string and has more than one word. It does not read like that from the docs, but it seems the implemented behaviour.

You are absolutely right. I missed that point that if there are more than one word in the entrypoint, they will become the arguments, together with cmd.
To verify that I checked the [implementation of the Docker engine](https://github.com/moby/moby/blob/3e1b15dc97630946c87ff7c4c1bf43128d908806/daemon/container.go#L176):

```go
return configEntrypoint[0], append(configEntrypoint[1:], configCmd...)
``` 

Only the first element of the entrypoint configuration list is used as the final entrypoint. The rest are used as arguments, together with the cmd configuration.

PR #1091, opened 2025-03-10
Only requires sudo when really needed

ribalba:
When running ./install multiple times a day it is quite annoying to keep unlocking sudo. This will only do the sudo calls if it is really needed

Still needs to be tested under Linux

### DIFF
--- a/install_mac.sh
+++ b/install_mac.sh
@@ -27,11 +27,23 @@ make -C "lib/c"
 
 build_binaries
 
-print_message "Adding powermetrics to sudoers file"
+print_message "Setting up powermetrics in sudoers files"
 check_file_permissions "/usr/bin/powermetrics"
 check_file_permissions "/usr/bin/killall"
-echo "${USER} ALL=(ALL) NOPASSWD:/usr/bin/powermetrics" | sudo tee /etc/sudoers.d/green_coding_powermetrics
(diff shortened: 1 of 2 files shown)


ArneTR:
For my understanding: This PR skips the sudo input in case that the file is already present?

I guess with the idea of not having to type in the password too often as this could be considered too much of a hassle?

If so I am not sure if I see this as a useful addition for three reasons:
- It increases the complexity of the install script
- It has some use on macOS but very limited use on linux as for most of the operations of the script (like updating apt) sudo is anyway necessary and as soon as we have it sudo access is cached for the file
- In order for this time saver to work you need 544 rights on the file. This is not recommended practice as it can open a security issue for an attacker to know which sudo-free calls are possible and try to exploit them. (I know ... debateable espeically since GMT is open source ... but I am thinking of automated tools here that do not google the source code)

So all in all I see more downsides than gains here. Happy for your feedback and happy to be corrected.

ArneTR:
Sorry for not saying anything to your PR message ... seem to have overlooked that.

Reading that my response to your case would actually be: If you do it so many times a day (which I consider very unusual practice and is more for us devs, than the average user): Have a sudo shell open.

I don't think GMT should solve something here by opening even remotely security issues, which are for DEVs only, if it can be mitigated with simply having a sudo shell open to work with during the day

PR #1093, opened 2025-03-11
Add compose stack name + allow test stack to run in parallel

davidkopp:
When running the containers of GMT in Docker Desktop the stack was called "docker" because of the base path. This PR adds a proper stack name: `green-metrics-tool`.

Docs:
- https://docs.docker.com/compose/how-tos/project-name/
- https://docs.docker.com/reference/compose-file/version-and-name/#name-top-level-element

One use case of having different stacks is to have different environments. So I thought it could make sense to allow to run the test environment in parallel to the production/development environment.

## Greptile Summary

This PR adds a proper stack name to Docker Compose configurations and enables parallel execution of test and production environments by modifying port configurations and service names.

- Added stack name 'green-metrics-tool' in `/docker/compose.yml.example` for better organization in Docker Desktop
- Changed PostgreSQL port from 9573 to 9574 in test configurations to allow parallel environment execution
- Modified `/tests/setup-test-env.py` to handle new port mappings for nginx (9143) and postgres (9574)
- Updated `/tests/test_functions.py` to use dynamic database configuration from GlobalConfig instead of hardcoded values
- Added 'test-' prefix to volume and service names in test environment for clear separation

<sub>💡 (4/5) You can add custom instructions or style guidelines for the bot [here](https://app.greptile.com/review/github)!</sub>

### DIFF
--- a/api/api_helpers.py
+++ b/api/api_helpers.py
@@ -29,12 +29,14 @@ import redis
 from enum import Enum
 
 def get_artifact(artifact_type: Enum, key: str, decode_responses=True):
-    if not GlobalConfig().config['redis']['host']:
+    host = GlobalConfig().config['redis']['host']
+    port = GlobalConfig().config['redis']['port']
+    if not host:
         return None
 
     data = None
     try:
-        r = redis.Redis(host=GlobalConfig().config['redis']['host'], port=6379, db=artifact_type.value, protocol=3, decode_responses=decode_responses)
+        r = redis.Redis(host=host, port=port, db=artifact_type.value, protocol=3, decode_responses=decode_responses)
 
         data = r.get(key)
     except redis.RedisError as e:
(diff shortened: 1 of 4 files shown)


ArneTR:
Looks like a good idea to me. @ribalba can you test this please? Did only look at time bc sickness

ArneTR:
@greptileai

davidkopp:
I forgot to check if the frontend tests are still working. Sorry for that.
I have now added some fixes so the frontend tests work again. For that I needed a workaround, see my change in `tests/start-test-containers.sh`. Hope that's ok. I didn't wanted to duplicate the frontend config, so I came up with this workaround to temporarily change the ports only during the start of the containers.

ArneTR:
Test seem to work fine. Did now also a code review. Here my remakrs:

- Greptile mentioned that the stack name is set in the loop. I think that is unnecessary, or? If so, can you move it outside the service loop
- You are setting the DB to a different port for the tests inside the containers and then map TEST_DATABASE_PORT:TEST_DATABASE_PORT. To my understanding this makes it work by accident. I would argue for leaving the DB internally at 9573 and just changing the mapping
- When I try to have the test setup and the live setup running locally I get a conflict. Redis is not handled atm. Was that intended?

ArneTR:
Code looks good and works well on my local box. Running the tests now!

One thing that should be considered: It is still not possible to run the tests AND a benchmarking run in parallel as the providers will conflict each other. I however also see no valid reason to do that, other than some next level hyper parallel development. If that is needed we typically outsource the test to GitHub to run in parallel.

ArneTR:
Thanks for the PR ❤️. merged.

PR #1094, opened 2025-03-11
Check container health also for containers without dependencies

davidkopp:
I had the issue that a container has a long startup time (due to database migrations), but GMT didn't checked the health condition, so the flow was already starting, but the container was not ready yet.

I think it makes sense to check all containers if they are started and if they are healthy (if a healthcheck is implemented) in the end of the boot phase before starting with the next phase (idle / runtime). This should be independent of whether or not there are dependencies between services.

This PR adds such a health check to the end of the boot phase. To avoid duplicated code I had to refactor the depends_on implementation.

The PR also changes a config parameter (`wait_time_dependencies`). Is this is problem?

I have set the waiting time for a container to become healthy to 300 seconds. Is this too long? The application I have mentioned in the beginning takes on my laptop around 200 seconds to boot.

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -83,7 +83,8 @@ measurement:
   post-test-sleep: 5
   phase-transition-time: 1
   boot:
-    wait_time_dependencies: 60
+    wait_time_container_started: 30
+    wait_time_container_healthy: 300
   metric-providers:
 
   # Please select the needed providers according to the working ones on your system
--- a/runner.py
+++ b/runner.py
@@ -760,6 +760,38 @@ class Runner:
         print("Startup order: ", names_ordered)
         return OrderedDict((key, services[key]) for key in names_ordered)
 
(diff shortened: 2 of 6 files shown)


ArneTR:
How does docker compose behave if no *depends_on* is set? Does it abort / stall the startup process if a container does not get healthy?

I would like to mimick the behaviour to docker natively and not let GMT behave differently.

Furthermore: Is it not beneficial for your compose file to have *depends_on* also set? Or what is the use case of having a *healthcheck* but no *depends_on* ?

davidkopp:
If there is no *depends_on*, docker compose only updates the health status from "starting" to "healthy" / "unhealthy", but as far as I can see without any consequences. If the status becomes unhealthy the startup proceeds nonetheless.

So you're right: it makes sense to just use `depends_on` with a healthcheck to achieve what I wanted. There is no reason for this pull request. I will close it.

PR #1106, opened 2025-03-21
Adds monotonic clocks

### LINKED ISSUE #1092: Monotonic check of procfs fails in WSL2
The check, if the metrics of procfs are monotonic increasing fails in my environment every time:

```
Error: Base exception occured in runner.py

Exception (ValueError): Time from metric provider cpu_utilization_procfs_system is not monotonic increasing
```

I guess I can blame WSL2 for that. For testing usage scenarios I'm using Windows 11 at the moment with WSL2. My CPU is AMD Ryzen 7 PRO 7840U.

There are 3 workarounds:

- disable the procfs metric provider
- using the runner.py switch `--dev-no-metrics`
- commenting out the following line:
  https://github.com/green-coding-solutions/green-metrics-tool/blob/5504a9a5c6751628112bd98634ad0d9d5862b904/metric_providers/base.py#L166

I'm thinking about if it makes sense to add another switch to runner.py (e.g. `--skip-metric-checks`) that allows to disable the metric checks. Not sure if it's worth it.

ribalba:
Closes #1092

I used `CLOCK_MONOTONIC_RAW`. I did some bench marking and it wasn't much slower as CLOCK_MONOTONIC even if some blogs say so. It will not be relevant in the ms space we are working in. 

Just as an FYI: CLOCK_MONOTONIC never experiences discontinuities due to NTP time adjustments, but it does change in frequency as NTP learns what error exists between the local oscillator and the upstream servers.

## Greptile Summary

This PR implements CLOCK_MONOTONIC_RAW timing across metric providers to address time monotonicity issues, particularly in WSL2 environments where NTP adjustments can cause discontinuities.

- Removed gettimeofday fallback (-m flag) from all metric providers, standardizing on CLOCK_MONOTONIC_RAW through new gmt-lib functions
- Added `get_time_offset()` and `get_adjusted_time()` in gmt-lib.c for reliable monotonic timing with real-time offset
- Several metric providers still include 'm' in getopt strings despite removing the flag functionality
- Multiple providers have memory management issues with containers_string not being freed and incorrect allocation checks
- Some files have outdated comments about static variables and thread safety that should be updated

The changes effectively address the WSL2 timing issues while consolidating timing functionality, though there are some cleanup tasks needed around error handling and memory management.


ribalba:
@ArneTR can you please ask greptile

ArneTR:
@greptileai

ArneTR:
Seems pretty straight forward!

Please look at the requests from Greptile
- Here especially: I am thinking of turning off everything that is categorized as "style: " bc I think this is just noise. What do you think?

2. I agree with greptile that the -m flag is unintuitive. I would expect the exact other way round behaviour: ACTIVATING the monotonic clock.

Having said that: What is the intention of keeping both implementations? I think this is an extra unnecessary branching and we do not plan to have both calls supported ,or? Do we need to?
In that path which should be rather quick I would remove it ... or?

ribalba:
2) Yes, what do you suggest. I couldn't come up with something better. Everything else would be random
3) The idea is, that we might use the providers for logging long running jobs. Let's say you want to log the RAPL energy into a file and collect it with Logstash. Then the times could be different, as syslog uses `gettime`. This might lead to confusion if you want to debug something.

ArneTR:
2. I suggest to remove -m entirely
3. I understand the use case you are seeing there, but I argue that this case is very fabricated. Syncing log times has never been our concern so far and if it happens BETWEEN systems we anyway have to do that manually. I would argue for removing this and bringing it in once we need it ... then we might even have more / different requirements

ribalba:
It's not about syncing between machines. It's about anything that runs long. As soon as the metrics provider runs for anything longer the possibility arises that the times it reports will be different from the system time creating confusion. 

But I can't really find this one case we need it right now. So I removed it

ArneTR:
@ribalba Can you please resolve all the converations that are still open if done. I have implemented a new branch rule that should block merging if still unresolved conversations exist. Will review then

ribalba:
done

ArneTR:
@greptileai

ArneTR:
@ribalba Greptile has more for you :)

Important though: can you check how that file ever did compile? If you are referencing an unknown variable it should fail IMHO ...

ArneTR:
Very nice. Just tested on my linux box and all values look good.  Also no overhead increase from using this timer in GMT overhead.

5g of carbon (according to Eco CI) well invested :)

PR #1107, opened 2025-03-21
SSH connection and other active user warning

ArneTR:
- Added warning if there are active SSH connections on the system which could influence measurement
 - Added warning if there are active users on the system whose activity could influence measurement

## Greptile Summary

Added system checks to warn users about active SSH connections and logged-in users that could affect measurement accuracy in the Green Metrics Tool.

- Added check in `lib/system_checks.py` to detect active SSH connections using pgrep sshd
- Added check in `lib/system_checks.py` to identify active users via the who command
- Both checks are implemented as WARN level to alert without blocking execution
- Checks align with GMT's goal of ensuring accurate energy consumption measurements

## Summary

- Added warning-level SSH session checks for active SSH connections and remote interactive sessions.
- Registered the check in the start and end system-check suites.
- Added `NOT_IMPLEMENTED` handling for unsupported platforms, including Windows encoding checks.
- Added output handling for checks that return `NOT_IMPLEMENTED`.
- Raised errors when SSH configuration cannot be read or contains no valid ports.

### DIFF
--- a/lib/system_checks.py
+++ b/lib/system_checks.py
@@ -44,6 +44,15 @@ def check_one_energy_and_scope_machine_provider():
     energy_machine_providers = [provider for provider in metric_providers if ".energy" in provider and ".machine" in provider]
     return len(energy_machine_providers) <= 1
 
+def check_ssh_connections():
+    ps = subprocess.run(['pgrep', 'sshd'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
+    return ps.stdout == b'' and ps.stderr == b'' and ps.returncode == 1
+
+def check_active_users():
+    ps = subprocess.run(['who'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
+    return ps.stdout == b'' and ps.stderr == b'' and ps.returncode == 0
+
+
 def check_tmpfs_mount():
     return not any(partition.mountpoint == '/tmp' and partition.fstype != 'tmpfs' for partition in psutil.disk_partitions())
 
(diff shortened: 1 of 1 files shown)


lib/system_checks.py
```diff
@@ -44,6 +44,15 @@ def check_one_energy_and_scope_machine_provider():
     energy_machine_providers = [provider for provider in metric_providers if ".energy" in provider and ".machine" in provider]
     return len(energy_machine_providers) <= 1
 
+def check_ssh_connections():
+    ps = subprocess.run(['pgrep', 'sshd'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
```
ribalba, line 48: But this checks if the demon is running not if someone is connected. Even if no one is connected there will still be a sshd demon running so people can connect. 

`who` should expose users that are connected via sshd or `ss -tnp | grep :22` will show open ssh connections

@ArneTR

ArneTR: I did look into that when creating the PR and apparently my testing was not sufficient.

On my test system the ssh port listening was done by systemd. Looked like this:
```bash
State           Recv-Q           Send-Q                     Local Address:Port                     Peer Address:Port          Process
LISTEN          0                128                              0.0.0.0:2798                          0.0.0.0:*              users:(("sshd",pid=597,fd=3))
LISTEN          0                128                                 [::]:2798                             [::]:*              users:(("sshd",pid=597,fd=4))
```

But now after your commented I probed more systems, where the *sshd* process directly sets the listener:
```bash
State            Recv-Q           Send-Q                     Local Address:Port                     Peer Address:Port           Process
LISTEN           0                4096                                   *:22                                  *:*               users:(("sshd",pid=5830,fd=3),("systemd",pid=1,fd=309))
```

The reason for not going for port 22 is that I also want to support ssh watching on non standard ports.

However given that I cannot identify the process I currently have no idea how to do it. Since the *sshd* is not readable the best shot would be to parse *ss* and check for port 22 or for sshd as process user ... 
Or, maybe even better, to not allow any listening connections at all in the *ss* list ... (apart from DNS and our DB connection)?

 You got a more sleek idea? Or any remarks?

ribalba: Couldn't you use `who | grep pts`?

ArneTR: I will look into this, but I have a feeling that forced commands via SSH are not captured through that mechanism ...

Bob-Sideshow: No, but isn’t that a server configuration? I would say that we add something into the documentation that this will then not work. My assumption is that 99+% will have a shell open.

ribalba: No, but isn’t that a server configuration? I would say that we add something into the documentation that this will then not work. My assumption is that 99+% will have a shell open.

ArneTR:
Branch has been reworked to look for listening on connections on SSH ports

ArneTR:
@coderabbitai review

ArneTR:
Error is unrelated and will investigate in different PR

PR #1111, opened 2025-03-23
Force compare mode

ArneTR:
This PR introduces an expert compare mode, where GMT can be forced to use a specifc mode.

Still some restrictions apply as that all the runs must fall in max. 2 buckets.

Screenshot 2025-03-23 at 5 49 28 PM
Screenshot 2025-03-23 at 5 49 22 PM

### DIFF
--- a/api/api_helpers.py
+++ b/api/api_helpers.py
@@ -65,21 +65,21 @@ def store_artifact(artifact_type: Enum, key:str, data, ex=2592000):
 # Function furthemore uses .substr instead of just replacing the unit, as some units have demominators like Bytes/s or
 # ugCO2e/ page request which we want to retain
 #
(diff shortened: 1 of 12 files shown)


davidkopp:
There is a little bug in the frontend if I want to compare runs that are not on the first page (>10 entries by default) on the runs page. If I'm on the e.g. on the second page and I click on the new dropdown menu in the compare column the UI jumps back to the first page and I can't select an option.
Current workaround: Display more entries on the first page.

ArneTR:
thanks, bug confirmed. Will brush it up tomorrow. Good to hear you found a temporary workaround

ArneTR:
Ok, I reworked the display component. It should be usable online now.

Will merge the PR once I added proper tests

PR #1119, opened 2025-03-25
Adds a requirements file to ee and skips the init if already done

ribalba:


lib/install_shared.sh
```diff
@@ -260,8 +262,13 @@ function checkout_submodules() {
         git submodule update --init lib/sgx-software-enable
     fi
     git submodule update --init metric_providers/psu/energy/ac/xgboost/machine/model
+
     if [[ $enterprise == true ]] ; then
-        git submodule update --init ee
+        if [ ! -d "ee/.git" ]; then
```
ArneTR, line 267: That condition will just check if the repo has been checked out before.

However the idea of the install script is to not only install, but also to update GMT.

the command `git submodule update --init ee ` does the update. It cannot be skipped on install.

Or am I missing something here?

What is the idea in the first place? To not overwrite the dir if it has unstashed changes?

ArneTR (changes requested):

lib/install_shared.sh
ribalba: Yes the way I modelled the checkout of the ee branch here

https://github.com/green-coding-solutions/gmt-enterprise/pull/13

is to already check out the brach in the GitHub workflow. 

What about doing a
```
cd ee
git pull
```

?

ArneTR: What is the git pull supposed to do? As an alternative to updating?
I would argue that this is not sufficient as:
- We want the connected version of the subrepo to main not the latest version of the enterprise repo. They might not align
- Making a git pull will fail if you have uncommited stuff ... actually that can be considered a good thing though and I like that! One should not overwrite stuff and I feel it is fine to fail then. The fail should be more graceful though. I would isse a git status before and only continue if that did not fail.

I understand that this implementati helps making the pipeline test correctly, but it prevents the GMT from updating.

Effetively we need different handling when GMT is updating normally (we want to check out the version bundled with GMT main) and when run in a pipeline (where we want to preseve the branch checked out).

So I would argue a new switch is in order ...

PR #1127, opened 2025-03-27
Entrypoint as list functionality

### LINKED ISSUE #1100: Entrypoint and command must support long form
Currently we support entrypoing and cmd in short form.

This is a great first step (ty @davidkopp )

However the handling of docker compose is suprisingly more complex than it was first implemented in GMT when command was implemented.

The long form (list form) is a direct command pattern whereas the short form starts a shell session (sh -c).

In our current implementation we suggest that the shell form is used ... but it is not!

Thus we need an implementation that is identical to docker compose.

@ribalba Can you please deep dive here and make a proper implementation

ribalba:
Closes #1100

I also fixed some failing tests that you seem to have also encountered in the meantime

## Greptile Summary

This PR enhances Docker entrypoint/command handling to support both shell (short form) and direct execution (long form) and updates tests and schema accordingly.

• Added tests at /tests/data/usage_scenarios/test_docker_compose_entrypoint.yml and /tests/data/usage_scenarios/test_docker_entry_command.yml validating both forms and parameter edge cases.  
• Updated /runner.py to split entrypoint/command tokens correctly, aligning behavior with Docker Compose, though error messages may misreference “command” instead of “entrypoint.”  
• Modified /lib/schema_checker.py to accept both string and list types for entrypoint and command.  
• Revised tests in /tests/test_usage_scenario.py for improved assertion flexibility.


ribalba:
@ArneTR Greptile please

ArneTR:
@greptileai

ArneTR:
Regarding: I also fixed some failing tests that you seem to have also encountered in the meantime

Your fix might be problematic. I do not believe this is an actual fail. The problem is: The process returns a different return code on macOS vs Linux ... I do not want to simply truncate the message here as the code is relevant ... currently I do not know why this discrepancy happens thoug. Can you please rever this part for now

ArneTR:
I referenced the issue I was talking about here: https://github.com/green-coding-solutions/green-metrics-tool/issues/1128

However the tests you have fixed works for me on macOS and Linux. Seems like you have tried on an outdated version. Please merge in main again

ribalba:
removed the failing tests

ArneTR:
LGTM

PR #1133, opened 2025-03-29
Moving to milli-watt hours

ArneTR:
This PR moves GMT to milli-watt-hours by default.

See Discussion: https://github.com/green-coding-solutions/green-metrics-tool/discussions/1132


ArneTR:
has been tested in the SaaS for a while and no complaints. I believe it is more valueable and thus will merge

PR #1135, opened 2025-03-29
Hidden phases feature

ArneTR:
Introduces the concept of the hidden phases feature.

Screenshot 2025-03-29 at 10 29 06 AM
Screenshot 2025-03-29 at 10 29 08 AM
Screenshot 2025-03-29 at 10 29 11 AM


ribalba:
I like the idea. I have this quite a few times that this would make the output nicer. `sleep` being the classical.

PR #1163, opened 2025-04-25
openSUSE & Rancher Desktop support (rootless)

thiloettelt:
Hello,

I made some rather naive changes to the install_linux.sh and install_shared.sh in order to support installation on openSUSE (tested on tumbleweed) and also triggering rootless mode when discovering Rancher Desktop.

Sincerely,
Thilo


ArneTR:
Uhh, great! Thanks for the PR! 🥰

The submitted additions to the install files look all sound to me.

Could you also double check which libraries need to be uninstalled again in: https://github.com/green-coding-solutions/green-metrics-tool/blob/main/uninstall.sh

The uninstall script gives the option to first remove all installed libraries via the install script. So it will be the same listing I guess ... unless package naming for uninstall works differently on *openSUSE* 

But also the uninstall script will remove pre-required dependencies, which we list in the docs under: https://docs.green-coding.io/docs/installation/installation-linux/#downloading-and-installing-required-packages

Could you list the packages and the respective names you had to install on your distribtion in the uninstall script?

I will then transport the information to the docs and update it with an `openSUSE` instruction given your PR.

Thanks again so much, we love that the tool gets used on openSUSE too :)

ArneTR:
Hey @thiloettelt ,

we haven't heard any feedback on this PR for a while. Did you read the reply?

Happy to talk futher! Would be great if you could provide an update if the PR is still "alive"

thiloettelt:
What a timing, I just resumed working on the PR.
I have just updated the PR and want to test it on a clean VM.
I'll give you notice today, when it's done.

thiloettelt:
For conformity I'm also using lsb_release for OpenSUSE right now.
However I have found out that the lsb-release package is not installed on OpenSUSE by default.
The common way would be to use /etc/os-release
It should be something that basically works on any distro.
Not sure how to proceed with this.

ArneTR:
Using `/etc/os-release` seems far cleaner. Please use that instead of `lsb_release`. We will also change the code on main. Thanks for the hint!

thiloettelt:
I have replaced lsb_release for the opensuse parts wirth /etc/os-release and added some missing optional uninstall commands for opensuse.

install_linux.sh
```diff
@@ -31,6 +33,10 @@ if [[ $activate_scenario_runner == true ]] ; then
             print_message "You can add -S to the install script to skip installing lm_sensors. However cluster mode and temperature reporters will not work then." >&2
             exit 1
         fi
+    elif cat /etc/os-release | grep -q "openSUSE"; then
+        if ! sudo zypper -n in sensors libsensors4-devel; then
```
ArneTR, line 37: Is it not needed to install `lm_sensors` ?

install_linux.sh
```diff
@@ -131,6 +143,8 @@ if ! mount | grep -E '\s/tmp\s' | grep -Eq '\stmpfs\s' && [[ $ask_tmpfs == true
     if [[ "$tmpfs" == "Y" || "$tmpfs" == "y" ]] ; then
         if lsb_release -is | grep -q "Fedora"; then
             sudo systemctl unmask --now tmp.mount
+        elif cat /etc/os-release | grep -q "openSUSE"; then
+            echo "You probably already have /tmp on a tmpfs"
```
ArneTR, line 147: Can this not be force-enabled? If the user selects yes and action should happen I would argue.

lib/install_shared.sh
```diff
@@ -344,7 +344,7 @@ function build_containers() {
 
     if [[ $build_docker_containers == true ]] ; then
         print_message "Building / Updating docker containers"
-        if docker info 2>/dev/null | grep rootless || [[ $(uname) == "Darwin" ]]; then
+        if docker info 2>/dev/null | grep rootless || docker info 2>/dev/null | grep lima-rancher-desktop || [[ $(uname) == "Darwin" ]]; then
```
ArneTR, line 347: lima to my knowledge is a VM based solution. What is the rationale here to check for this support?

ArneTR: is the sudo call failing on your system?

uninstall.sh
```diff
@@ -49,6 +51,8 @@ if [[ $(uname) == "Linux" ]]; then
     if [[  "$pre_install" == "Y" || "$pre_install" == "y" ]] ; then
         if lsb_release -is | grep -q "Fedora"; then
             sudo dnf remove -y curl git make gcc python3 python3-devel
+        elif cat /etc/os-release | grep -q "openSUSE"; then
+            sudo zypper rm --dry-run -n git make gcc python313 python313-pip python313-virtualenv
```
ArneTR: removal of curl here also? or can it not be removed as it is a core package?

install_linux.sh
ArneTR: or `sensors` ... this is how it is called in the uninstall.

ArneTR (changes requested):
I looked at the PR and have some minor questions but also a confusion with lima, then I feel it can be merged in and is a really nice addition! Thanks again!

install_linux.sh
thiloettelt: it's installing sensors, which is the equivalent to lm_sensors on opensuse and libsensors4-devel in that line. Same for uninstall.sh - did I miss something?

lib/install_shared.sh
thiloettelt: Yes, it's failing because Rancher Desktop registers nerdctl or corresponding docker aliases within the rc file of the user.
The intention of this line is to check if Rancher Desktop is being used and then handle everything rootless.
But not just because of the rc configuration thing, but also because the default mode of Rancher Desktop is rootless.

uninstall.sh
thiloettelt: I don't know about the curl, that's in the line that is not from me.
I believe all of these packages actually come installed on OpenSUSE, so I am not sure whether there should be an option to remove them for openSUSE.

install_linux.sh
ArneTR: sorry. I now see it and must have been blind

uninstall.sh
ArneTR: I understand. To my research it is not installed in minimal. So in the spirit of "removing all dependencies" we will remove it. No need to update the PR. Will do it after merge

lib/install_shared.sh
ArneTR: Ok, understood. I guess it should be changed then from our side. The `sudo docker` is legacy and since we now have the first quirk with it we remove it. The check for lima feels very specific and going for a general requirement to have the docker command be non-root runnable seems like the better option here. Feel free to remove the line. Otherwise I will remove it after merge.

install_linux.sh
ArneTR: This is the only question open. If you could force mount on tmpfs (typically through systemd) this is ready for merge

thiloettelt: Good question, I have found out that tmpfs has been the standard for OpenSUSE almost 5 years for now.
Maybe we can skip this thing for opensuse, instead of asking the user. Or even better find out if it is already using tmpfs (which I don't know yet how to do).
Source: https://en.opensuse.org/openSUSE:Tmp_on_tmpfs

uninstall.sh
thiloettelt: I had accidentally introduced --dry-run, which I just corrected.

install_linux.sh
ArneTR: It can be checked, yes.

What is the output of: `findmnt -n -o FSTYPE /tmp` for you?
If it is `tmpfs` than the command works as expected and I can bring it in

thiloettelt: Yes, the command works, it's tmpfs.

lib/install_shared.sh
thiloettelt: Please remove if after merge to your liking

PR #1164, opened 2025-04-27
Isaqb green

pkutschera:


ArneTR:
Hey @pkutschera,

thanks for the PR! I am unsure if that was intended. This seems like it should be on your fork ...?

In any case: If you folks want that the repository mentioned in the splash.md shall be continuously monitored if it is still working with the GMT you are happy to add it to the GMT as a "once per week" job. In case you make changes to it you can then also see the changes over time

See: https://metrics.green-coding.io/request.html

pkutschera:
It was not intended :-)

PR #1166, opened 2025-04-27
Resource limits rework

ArneTR:
This PR reworks how GMT handles Resource Limits.

Previously this was done through the `deploy` key in `services`.
However this key is intended for Docker Swarm.

In the future GMT will only support the `cpus` and `mem_limit` keys under `services`.

When using Docker Compose directly the `deploy` settings for CPU and Memory have no further implications and will just overwrite any settings given via services top level keys. Implementation is here: https://github.com/docker/compose/blob/5bb46035cf12e5b2fc2f3b09522a392b7b753120/pkg/compose/create.go#L713

Futhermore GMT understands now extra top level keys that instruct the cluster to "simulate" a smaller machine.
The values for energy, utilization etc. are then only accounted for this smaller box.

Example: The box has 4 cores and but the key `virtualization` (name might change) specifies only 2 CPUs.
Then the total energy consumption of the machine will have all "operational" overhead attributed but only half of the baseline, and half of the embodied carbon. This is because the other half goes to the "other VM" on the machine.

With this GMT incentivises users to use capactity planning and also experience how the carbon goes down if less ressources are requested.


ArneTR:
Closed due to implementation in different PR

PR #1172, opened 2025-05-01
Usage Scenario Variables and ScenarioRunner templates

ArneTR:
Adds measurement templates for beginners and quick measurements with GMT.

YOu can now do:

- Quickly measure a website
  - `./run-template website "https://www.google.de"`
- Quickly measure an AI prompt (will default to gemma3:1b)
  -  `./run-template ai "How cool is the GMT?"`

This makes the tool more beginner friendly, but also abstracts away some of the pitfalls that are happing with measuring these two cases for instance that for both you need to have sensible providers active and for websites the resolution should be at least 1/10 of the page load time (typically 10 ms)

We will iterate on these modes and add warnings / guard-clauses as needed, but will have them as bare like this for the moment.

## Greptile Summary

Added measurement templates and usage scenario variables to enable quick website and AI measurements, with significant updates to the ScenarioRunner and API endpoints.

- Added `usage_scenario_variables` JSONB column to runs, jobs, and watchlist tables with proper constraints
- Introduced `/v2/jobs` and `/v2/runs` endpoints, deprecating v1 versions with backward compatibility
- Added validation and error handling for usage scenario variables in ScenarioRunner
- Added test coverage for variable handling, template functionality and API changes
- Security concern: Potential shell injection risk in Docker command execution with `shell=True`


ArneTR:
@ribalba How do you like these modes? Do they fulfill your idea of an easy mode?

I am keeping the templates in the GMT repository and thus they are automatically versioned. To not muddy the overview of which version they have been changed in the GMT now overloads the commit-hash with the actual folder of the templates. Thus a change to the AI template will still keep the same commit_hash for the website template and vice versa.

ArneTR:
@greptileai

ArneTR:
@greptileai

PR #1176, opened 2025-05-06
Adds the * functionality for filenames, multiple filenames and --iterations

ribalba:
This is a patch I often use when running benchmarks. Basis for a discussion


ArneTR:
I am a bit skeptical of this PR.

I do see the need to run scenarios one after another. We also do have a functionality called "Variance" in our API that inserts a run 3 times.

However on the CLI I wonder if the gain is really given bc users could just do:
```bash
for i in {1..3}; do python3 runner.py --uri XXX --filename XXX ; done
```
if they really want to do this.

On the other hand I see other tools like `perf` that also have a *repetition* parameter to repeat tests and thus nudge the user to use it, but that is for quick introspection of stuff.

With the GMT we want users to only use CLI functionality to develop and setup stuff quickly. Proper measurements must be done in Cluster mode.

So why nudge the user to use behaviour that is good for deriving StdDev and Means of stuff, if you are connected via SSH and on the CLI, which is skewing results in the first place?

ribalba:
Then maybe we should have a script ./submit_to_cluster that does these things for you. For the work I am doing now I have 15 usage scenario files that I need to run and I want 5 repeats. Of course I could do this with shell magic but like this I can just it all in the GMT.

davidkopp:
Since I asked Arne about such a feature, it's time for this PR to be merged ;) I would like to have this feature to be able to check the standard deviation of multiple runs on my local system.

I (or Claude) added a few improvements in https://github.com/green-coding-solutions/green-metrics-tool/pull/1261. However, there is no significant change.

I was not sure how to handle the conflict with main. Normally I would do a rebase but, I did not want to destroy the branch history of Didi. Let me know if that's ok for you here or if I should create a new branch with your changes.

davidkopp:
btw: Claude complained about the usage of the glob pattern matching, because it could be a security risk to include any file on the file system. However, that should be ok here, since it's only used in the development environment and not on production.

ArneTR:
> btw: Claude complained about the usage of the glob pattern matching, because it could be a security risk to include any file on the file system. However, that should be ok here, since it's only used in the development environment and not on production.

I agree. I see no security issue here. System wide files can also be accessed with `--uri` and `--filename` already. Glob amplifies it for sure but opens nothing new. CLI mode is expecting user access to the machine and thus no security concern

tests/test_runner.py
```diff
@@ -151,3 +151,55 @@ def test_runner_run_invalidated():
...
+
+def test_runner_with_glob_pattern_filename():
+    """Test that runner works with glob pattern filenames like folder/*.yml"""
+    # Test runner.py with glob pattern that matches multiple files in a folder
+    ps = subprocess.run(
+        ['python3', 'runner.py', '--uri', GMT_DIR, '--filename', 'tests/data/usage_scenarios/runner_filename/*.yml',
+         '--skip-system-checks', '--dev-cache-build', '--dev-no-sleeps', '--dev-no-save'],
+        cwd=GMT_DIR,
+        capture_output=True,
+        text=True,
```
ArneTR: we are typically using encoding='utf-8' ... my last knowledge is they are synonyms. If so and text= not mandatory than standardize on encoding=utf-8

tests/test_runner.py
```diff
@@ -151,3 +151,55 @@ def test_runner_run_invalidated():
...
+        check=False
+    )
+
+    assert ps.returncode == 0, f"Runner failed with stderr: {ps.stderr}"
```
ArneTR: it can fail without stderr. Please tell returncode when failed and then output stderr alongside (as it might be empty)

tests/test_runner.py
```diff
@@ -151,3 +151,55 @@ def test_runner_run_invalidated():
...
+    assert 'tests/data/usage_scenarios/runner_filename/basic_stress_1.yml' in ps.stdout
```
ArneTR: this can easily flake, as it might also be part of a skip message. Please use matching that guarantees that it signals to be used in a run. If necessary patch scenario_runner to output this to stdout more clearly

ArneTR: applies for line below too

tests/test_runner.py
```diff
@@ -151,3 +151,55 @@ def test_runner_run_invalidated():
...
+    # Test runner.py with multiple files and iterations=2
+    ps = subprocess.run(
+        ['python3', 'runner.py', '--uri', GMT_DIR,
+         '--filename', 'tests/data/usage_scenarios/runner_filename/basic_stress_1.yml',
+         '--filename', 'tests/data/usage_scenarios/runner_filename/basic_stress_2.yml',
+         '--iterations', '2',
+         '--skip-system-checks', '--dev-cache-build', '--dev-no-sleeps', '--dev-no-save'],
+        cwd=GMT_DIR,
+        capture_output=True,
+        text=True,
```
ArneTR: standardize

tests/test_runner.py
```diff
@@ -151,3 +151,55 @@ def test_runner_run_invalidated():
...
+        check=False
+    )
+
+    assert ps.returncode == 0, f"Runner failed with stderr: {ps.stderr}"
```
ArneTR: return code

tests/test_runner.py
```diff
@@ -151,3 +151,55 @@ def test_runner_run_invalidated():
...
+    # Should see each file processed twice (2 iterations)
+    assert ps.stdout.count('tests/data/usage_scenarios/runner_filename/basic_stress_1.yml') == 2
```
ArneTR: better match

ArneTR: applies for line below too

tests/test_runner.py
```diff
@@ -151,3 +151,55 @@ def test_runner_run_invalidated():
...
+def test_runner_filename_pattern_no_match_error():
+    """Test that runner fails gracefully when filename pattern matches no files"""
+    # Test runner.py with pattern that matches no files
+    ps = subprocess.run(
+        ['python3', 'runner.py', '--uri', GMT_DIR,
+         '--filename', 'tests/data/usage_scenarios/nonexistent_*.yml',
+         '--skip-system-checks'],
+        cwd=GMT_DIR,
+        capture_output=True,
+        text=True,
```
ArneTR: standardize

runner.py
```diff
@@ -110,68 +113,90 @@
...
+            if using_default_filename:
+                print(TerminalColors.FAIL, f'Error: Default file not found: {pattern}', TerminalColors.ENDC)
+                print('Please create the file or specify a different file with --filename')
+            else:
+                print(TerminalColors.FAIL, f'Error: No valid files found for --filename pattern: {pattern}', TerminalColors.ENDC)
+            sys.exit(1)
+        filenames.extend(valid_files)
+
+    # Remove duplicates and repeat for each iteration
+    filenames = list(set(filenames)) * args.iterations
```
ArneTR: why remove duplicates? Maybe the user has a specifc repetition pattern in mind that shall be used? Vote to remove set()

ArneTR (changes requested):
Thanks for the PR. My remarks in the review.

Please also merge in main and care for the new arguments to be used correctly in the loop now. quite some new switches have been added

runner.py
davidkopp: Agree.
@ribalba The `set()` came from you. What's your opinion?

davidkopp: I have now removed the deduplication. This means that using the parameters `--filename usage_scenario.yml --filename usage_scenario.yml --iterations=2` would result in the file being executed four times.

davidkopp:
Found another bug in the code 🙈
--filename patterns were not resolved relative to URI path for local directories. That is now fixed.

I added all my tests to `tests/test_runner.py`. However, I just recognized, that all the other runner tests are in `tests/test_usage_scenario.py`. That was unexpected. Is this on purpose? I propose to move them all to `tests/test_runner.py` and merge a few tests that may be duplicated now in a future PR.

ArneTR:
PR looks good to me. Will merge.

@davidkopp Please document the new switches

davidkopp:
> * test_usage_scenario is supposed to test behaviour of the runner when given various usage_scenarios. Since they are such a big "argument" they have a separate class
> 
> * test_runner is supposed to test all other arguments like skip_system_checks, usage_scenario_variables etc.
> 
> 
> If that was not clear, maybe giving them other names would clarify?

That is also my understanding. However, the following tests are currently included in test_usage_scenario.py, but they would be a better fit for test_runner.py:

- `test_uri_local_dir`
- `test_uri_local_dir_missing`
- `test_uri_github_repo`
- `test_uri_local_branch`
- `test_uri_github_repo_branch`
- `test_uri_github_repo_branch_missing`
- `test_name_is_in_db`
- `test_different_filename`
- `test_different_filename_missing`
- `test_no_file_cleanup`
- `test_file_cleanup`
- `test_skip_and_allow_unsafe_both_true`
- `test_debug`
- `wip_test_verbose_provider_boot`

PR #1190, opened 2025-05-16
Trying to test disk.io providers under github actions

ArneTR:
## Greptile Summary

This PR enables disk I/O metric providers in the test configuration for GitHub Actions testing, adding both system-level procfs and container-level cgroup providers.

- Added `disk.io.procfs.system.provider.DiskIoProcfsSystemProvider` in `/tests/test-config.yml.example` for system-wide disk I/O monitoring
- Added `disk.io.cgroup.container.provider.DiskIoCgroupContainerProvider` in `/tests/test-config.yml.example` for container-specific disk I/O metrics
- Both providers configured with resolution: 99 to match existing metric provider settings

### DIFF
--- a/tests/test-config.yml.example
+++ b/tests/test-config.yml.example
@@ -70,8 +70,8 @@ measurement:
     linux:
       cpu.utilization.procfs.system.provider.CpuUtilizationProcfsSystemProvider:
         resolution: 99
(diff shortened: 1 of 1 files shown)


ArneTR:
We need to get the sector size through a different mean.

This could be an option: https://chatgpt.com/share/68273e8c-0744-8007-bb05-39530571f049

PR #1191, opened 2025-05-16
Hotfix: Compare values were 3 orders of magnitude to low due to double division

ArneTR:
This PR introduces a QoL feature by:
- Showing untransformed and unrounded values in title of values in metrics table
- If a value is lower than the rounding resolution and/or really 0 a tooltip warning is added

Fix:
- The frontend was displaying values wrongly as multiple orders of magnitude too low in comparison mode for diffing
  - Repeated runs were not affected
  - Single run display was not affected
  - API data retrieval was not affected.
  - Key metric boxes where unaffected
  - The relative deviation was not affected (percentage display in right column)

PR is already live on metrics.green-coding.io in this state.

However PR will stay open until tests are added to avoid regression

## Screenshots QoL
Hovering: Screenshot 2025-05-16 at 3 21 37 PM

Tooltip
Screenshot 2025-05-16 at 3 17 41 PM

## Greptile Summary

Fixed critical display issue in comparison mode where metric values were shown 3 orders of magnitude too low, and added tooltips for better value visibility.

- Fixed value calculation in `frontend/js/helpers/phase-stats.js` by directly using untransformed mean values from `detail_data['data'][key]?.mean`
- Added tooltips in `frontend/js/helpers/metric-boxes.js` to show original values when below rounding threshold
- Added comprehensive test suite in `tests/frontend/test_frontend.py` for comparison values and tooltips
- Note: Fix affects only comparison mode diffing; repeated runs, single run display, and key metric boxes were unaffected


ArneTR:
Regression tests added

ArneTR:
@greptileai

PR #1195, opened 2025-05-21
Checking outside symlinks when cloning from URL

ArneTR:
## Greptile Summary

Added security measures to prevent directory traversal attacks by checking for symlinks that point outside repository directories when cloning from URLs in the Green Metrics Tool.

- Added `is_outside_symlink()` and `find_outside_symlinks()` in `lib/utils.py` to detect potentially malicious symlinks using `os.path.realpath()`
- Integrated symlink security checks in `lib/scenario_runner.py` to prevent unauthorized file access when cloning repositories
- Added test cases in `tests/test_usage_scenario.py` to validate symlink security features
- Consider improving error handling for permission issues and access errors in symlink detection

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -12,6 +12,23 @@ from lib.db import DB
 
 CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 
+def is_outside_symlink(base_dir, symlink_path):
+    try:
+        target = os.readlink(symlink_path)
+        abs_target = os.path.abspath(os.path.join(os.path.dirname(symlink_path), target))
+        return not abs_target.startswith(os.path.abspath(base_dir)), abs_target
+    except OSError:
+        return False, None  # Not a symlink
+
+def find_outside_symlinks(base_dir):
+    for root, dirs, files in os.walk(base_dir):
+        for name in dirs + files:
+            full_path = os.path.join(root, name)
+            is_outside, target = is_outside_symlink(base_dir, full_path)
+            if is_outside:
+                return f"{full_path} → {target}"
+    return None
+
 def remove_git_suffix(url):
     if url.endswith('.git'):
         return url[:-4]
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -14,7 +14,7 @@ CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 
 def is_outside_symlink(base_dir, symlink_path):
     try:
-        abs_target = os.path.realpath(os.path.join(os.path.dirname(symlink_path), symlink_path))
+        abs_target = os.path.realpath(symlink_path)
         return not abs_target.startswith(os.path.realpath(base_dir)), abs_target
     except OSError:
         return False, None  # Not a symlink
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -14,9 +14,8 @@ CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 
 def is_outside_symlink(base_dir, symlink_path):
     try:
-        target = os.readlink(symlink_path)
-        abs_target = os.path.abspath(os.path.join(os.path.dirname(symlink_path), target))
-        return not abs_target.startswith(os.path.abspath(base_dir)), abs_target
+        abs_target = os.path.realpath(os.path.join(os.path.dirname(symlink_path), symlink_path))
+        return not abs_target.startswith(os.path.realpath(base_dir)), abs_target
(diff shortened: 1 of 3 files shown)


lib/utils.py
```diff
@@ -12,6 +12,23 @@
 
 CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 
+def is_outside_symlink(base_dir, symlink_path):
+    try:
+        target = os.readlink(symlink_path)
+        abs_target = os.path.abspath(os.path.join(os.path.dirname(symlink_path), target))
```
ribalba, line 18: I think this will still fail if you have a link pointing to a link. I think that is why we used `realpath` before

ribalba (changes requested):

lib/utils.py
ArneTR: You point out an interesting behaviour and apparently my understanding of .abspath was wrong. I thought it does what .realpath does. Thanks for pointing that out.

I added a test which created the crafted symlink `/tmp/repo/subfolder/../../../etc/passwd` bc then the .startswith comparison makes an abspath look like it is in the repo. but the realpath will resolve to `/etc/passwd`

I also thought about checking if a symlink actually resolves or not at the time of the repo checkout. But tampering afterwards is not really possible as the host os is not touched anymore and the folder is read-only anyway.

Can you give it another look? @ribalba

ArneTR:
@greptileai

ArneTR:
> Reading this https://amalmurali.me/posts/git-rce/ We might also think about
> 
> ```
> git clone -c core.symlinks=false -c protocol.file.allow=never
> ```
> 
> and just not allow symlinks on the cluster. If we want to make it even safer.

I think inside symlinks should be allowed. Added a test to make a regression check for that so we keep that in mind

lib/utils.py
```diff
@@ -12,6 +12,22 @@
 
 CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 
+def is_outside_symlink(base_dir, symlink_path):
+    try:
+        abs_target = os.path.realpath(symlink_path)
+        return not abs_target.startswith(os.path.realpath(base_dir)), abs_target
+    except OSError:
```
ribalba, line 19: In this new implementation what does trigger the `OSError`?

lib/utils.py
```diff
@@ -12,6 +12,22 @@
...
+        return False, None  # Not a symlink
```
ribalba, line 20: I also found the `os.path.commonpath` which might be a little better than `startswith` 

Also the `path` has a `resolve()` which resolves all symlinks.

lib/utils.py
```diff
@@ -12,6 +12,22 @@
...
+
+def find_outside_symlinks(base_dir):
+    for root, dirs, files in os.walk(base_dir):
```
ribalba, line 23: I think I might have found another attack. Not quite sure how to exploit it though right now. os.walk does not follow symlinks: https://docs.python.org/3.9/library/os.html?highlight=os%20walk
```
By default, walk() will not walk down into symbolic links that resolve to directories. Set followlinks to True to visit directories pointed to by symlinks, on systems that support them.
```
so maybe you could point to a dir that you can then somehow have a symlink in that will then not be checked. I am not going to spend too much time on this but maybe if you check in a dir where the permissions don't allow anything and then because you are root in the container you can access this dir. But I haven't tried it

ArneTR: good point to think about. But I think from a security point it does not matter. What we want to know is if the symlink goes out of the directory. As soon as walk finds one and it goes that will be alerted.

The user can link internally of the repository as much as they want. Since to be an effective escape they must link to the outside at some point. So even if one links to another symlink internally as long as THAT target does not link out it does not matter if we follow symlinks or not.

Actually it would be the other way round and you could trigger an endless recursion. So I would argue os.walk is fine

Do you agree?

ArneTR:
> > I think inside symlinks should be allowed. Added a test to make a regression check for that so we keep that in mind
> 
> I would probably vote to make it more secure. You can always create a symlink in the setup commands if you want to and getting this code right seems quite complex and prone for errors. For example if we ever allow write in the mounted folder as we have discussed before.

I agree that it would make it more secure as it would reduce possible stuff we overlooked.

However I would want people to have repositories that have symlinks and use them with GMT. It will be a very hard to find quirk if GMT does not check out symlinks.
If you have an NGINX configuration for instance it will most likely have symlinks as sites-enabled are all symlinked.

But happy to discuss further

PR #1213, opened 2025-06-03
POC: Generating OpenAPI client

winniehell:


api/main.py
```diff
@@ -173,6 +174,11 @@ async def update_user_setting(setting: UserSetting, user: User = Depends(authent
     from ee.api import ai_optimisations
     app.include_router(ai_optimisations.router)
 
+for route in app.routes:
```
winniehell, line 177: see c9e1aed9c84dadf9780490f638a1550cf8a7a848

frontend/js/stats.js
```diff
@@ -77,7 +79,8 @@ const fetchAndFillRunData = async (url_params) => {
     let run = null;
 
     try {
-        run = await makeAPICall('/v2/run/' + url_params['id'])
+        const api = new DefaultApi();
+        run = await api.getRun(url_params['id']);
```
winniehell, line 83: this is the relevant improvement. when using TypeScript, there could even be validation for parameters and responses on the client side

generate-api-client.sh
```diff
@@ -0,0 +1,18 @@
...
+
+set -euo pipefail
+
+output_directory="$(dirname $0)/frontend/js/api/"
+
+rm -rf "${output_directory}"
+
+openapi-generator-cli generate \
+  --generator-name javascript \
+  --input-spec http://127.0.0.1:8000/openapi.json \
```
winniehell, line 11: instead of using the backend port, a local file can be used

winniehell:
> Description missing

sorry about that! it was not meant to be a standalone changeset but only an additional proof of concept to highlight what we had talked about. I've added a proper description now.

> Is it a generator for OpenAI? Or is the generator bash script just linked for compeleteness and was initally used?

it uses a library (https://openapi-generator.tech/) to generate code from the specification provided by FastAPI. the bash script is just for documenting the command line flags that I have used. most of them could be added to a config file. whether or not to keep a script as a shortcut for the generator command is a matter of taste.

the command (or script) needs to be run whenever the backend API changes.

> This seems like a lot of boilerplate just to query the simple API.

that is true and most of it is due to the fact that OpenAPI is strictly typed while JavaScript is not. there is much less boilerplate when using TypeScript (but that would be a separate discussion). some of the boilerplate can be avoided by fine tuning the config and there is also an open feature request to skip client side validation and rely on server side validation instead: https://github.com/OpenAPITools/openapi-generator/issues/4168

> Can it not be used with a classical external tool instead of carrying everything inside the repository?

I think I have answered most of the question above. whether to add the (redundant) client code to the repository or not is a highly subjective matter. personally I prefer to do so because then you can search in it or run git blame without navigating to the backend (essentially avoiding context switches).

ArneTR:
We decided we do not want to have the boilerplate or a node client in the core repo. Thanks for the pointers though!

PR #1223, opened 2025-06-13
Fix script prune_db.py

davidkopp:
This PR fixes the script `tools/prune_db.py` to be able to remove failed runs or remove all measurement data.

### DIFF
--- a/tools/prune_db.py
+++ b/tools/prune_db.py
@@ -6,7 +6,7 @@ import faulthandler
 faulthandler.enable(file=sys.__stderr__)  # will catch segfaults and write to stderr
 
 from lib.db import DB
-from psycopg import errors
+from lib.global_config import GlobalConfig
 
 if __name__ == '__main__':
     import argparse
@@ -20,12 +20,17 @@ if __name__ == '__main__':
         print("This will remove ALL runs, measurement, CI, carbonDB and hog data from the DB. Continue? (y/N)")
         answer = sys.stdin.readline()
         if answer.strip().lower() == 'y':
-            tables = ['runs', 'ci_measurements', 'hog_measurements', 'carbondb_data', 'carbondb_data_raw']
-            for table in tables:
-                try:
-                    DB().query(f"TRUNCATE TABLE {table} CASCADE")
-                except errors.UndefinedTable:
-                    continue
+            DB().query(f"TRUNCATE TABLE runs CASCADE")
+            DB().query(f"TRUNCATE TABLE ci_measurements CASCADE")
+
+            if GlobalConfig().config.get('activate_carbon_db', False):
+                from ee.tools.prune_db_ee import prune_carbondb
+                prune_carbondb()
+
+            if GlobalConfig().config.get('activate_power_hog', False):
+                from ee.tools.prune_db_ee import prune_power_hog
+                prune_power_hog()
+
             print("Done")
     elif args.mode == 'failed-runs':
         print("This will remove all runs that have not ended, which includes failed ones, but also possibly running, so be sure no measurement is currently active. Continue? (y/N)")
--- a/tools/prune_db.py
+++ b/tools/prune_db.py
@@ -31,5 +31,5 @@ if __name__ == '__main__':
         print("This will remove all runs that have not ended, which includes failed ones, but also possibly running, so be sure no measurement is currently active. Continue? (y/N)")
         answer = sys.stdin.readline()
         if answer.strip().lower() == 'y':
-            DB().query('DELETE FROM runs WHERE failed = TRUE')
+            DB().query('DELETE FROM runs WHERE failed = TRUE OR end_measurement IS NULL')
             print("Done")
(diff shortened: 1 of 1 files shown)


tools/prune_db.py
```diff
@@ -19,15 +20,16 @@
         print("This will remove ALL runs, measurement, CI, carbonDB and hog data from the DB. Continue? (y/N)")
         answer = sys.stdin.readline()
         if answer.strip().lower() == 'y':
-            DB().query('TRUNCATE runs CASCADE')
-            DB().query('TRUNCATE ci_measurements CASCADE')
-            DB().query('TRUNCATE hog_measurements CASCADE')
-            DB().query('TRUNCATE carbondb_energy_data CASCADE')
-            DB().query('TRUNCATE carbondb_energy_data_day CASCADE')
+            tables = ['runs', 'ci_measurements', 'hog_measurements', 'carbondb_data', 'carbondb_data_raw']
```
ArneTR: I think it would be a cleaner option to introduce a switch here where you can select 'non-ee' and 'ee' tables.

Basically the pune_db script could have an '--ee' switch to decided then which tables to include

tools/prune_db.py
```diff
@@ -19,15 +20,16 @@
...
+                    DB().query(f"TRUNCATE TABLE {table} CASCADE")
+                except errors.UndefinedTable:
+                    continue
             print("Done")
     elif args.mode == 'failed-runs':
         print("This will remove all runs that have not ended, which includes failed ones, but also possibly running, so be sure no measurement is currently active. Continue? (y/N)")
         answer = sys.stdin.readline()
         if answer.strip().lower() == 'y':
-            DB().query('DELETE FROM runs WHERE end_measurement IS NULL')
+            DB().query('DELETE FROM runs WHERE failed = TRUE')
```
ArneTR: Did you experience issues with using `end_measurement is NULL`. At the moment the reality in the database is often that runs that had hard hickups do not get failed = TRUE set as the DB failed before that. But not having end_measurement is more reliable.

ArneTR (changes requested):
Thanks for the PR!

I totally see to update this script to work in non enteprise GMT mode.

I left some ideas and comments for discussion. Happy for your feedback

tools/prune_db.py
davidkopp: Yes, I had measurement runs in the database that ran until the end, but were correctly marked as failed. The attribute `end_measurement` was therefore not NULL and the script was not able to delete these failed measurements from the database.

So the query should be changed to `DELETE FROM runs WHERE failed = TRUE OR end_measurement IS NULL`?

tools/prune_db.py
davidkopp: Which ee tables are relevant here? (I don't have access to the enterprise version, so I can't it look up)

tools/prune_db.py
ArneTR: jap, that would be a better condition then.

This DB field is due for a refactor in the near future anyway. But checking for `failed = TRUE OR end_measurement IS NULL ` should be the best for now.

tools/prune_db.py
ArneTR: I am imagining more something like this:

```
if GlobalConfig().config.get('activate_carbon_db', False):
    from ee.tools.prune_db_ee import prune_carbondb
    prune_carbondb()

if GlobalConfig().config.get('activate_power_hog', False):
    from ee.tools.prune_db_ee import prune_power_hog
    prune_power_hog()
```

instead of the direct DB().query(...) calls.

In now feel that the code should have no knowlege of the tables necessarily. Only a function signature. That the calls are still in there seems like a missing code artifact from the refactor a while ago.

Since the code does not yet exist you can also suggest a different approach. I will then handle the appropriate implementation in the EE code

davidkopp: It makes sense to call a function here from the ee library. So I used your code now.

tools/prune_db.py
```diff
@@ -19,15 +20,21 @@
         print("This will remove ALL runs, measurement, CI, carbonDB and hog data from the DB. Continue? (y/N)")
         answer = sys.stdin.readline()
         if answer.strip().lower() == 'y':
-            DB().query('TRUNCATE runs CASCADE')
-            DB().query('TRUNCATE ci_measurements CASCADE')
-            DB().query('TRUNCATE hog_measurements CASCADE')
-            DB().query('TRUNCATE carbondb_energy_data CASCADE')
-            DB().query('TRUNCATE carbondb_energy_data_day CASCADE')
+            DB().query(f"TRUNCATE TABLE runs CASCADE")
```
ArneTR, line 23: fstring unnecessary

tools/prune_db.py
```diff
@@ -19,15 +20,21 @@
...
+            DB().query(f"TRUNCATE TABLE ci_measurements CASCADE")
```
ArneTR, line 24: fstring unnecessary

ArneTR:
Looks good! 

My pylint complained about the unnecessary f-strings. Will fix from my side after merge.

Thanks for the pro-active PR!

PR #1229, opened 2025-06-18
Switching to CLOCK_BOOTTIME to account for suspend states

ArneTR:
This PR switches the used Clock in GMT metric providers to`CLOCK_BOOTTIME` to mitigate following issues:

- if the system goes to suspend the clock times returned by the metric providers currently using `CLOCK_MONOTONIC_RAW` with just an initial calculated offset will get skewed. Since the output is a *Unix Timestamp* this is considered unintended behaviour and we want a proper *Unix Timestamp* as the output.
- Metric Providers that use easier syscalls from for instance the CLI world typically access 'CLOCK_REALTIME' by default (for instance `date`). Providers like `PsuEnergyAcIpmiMachineProvider` will thus be off from the `CLOCK_MONOTONIC_RAW` and we need to standardize  this to be the actual time for both.

The current implementation changes the clock but leads to mal-accounting in providers that account for power instead of energy as seen in screenshot #1. The time gap is understood as the system being running on the same power for the pause that has occured.

We need to decide if that is what we want to allow or not. Do cases exists where this is valid power accounting?
In the case of suspend it is definitely wrong. Since no power values are know to us and the system is in very low power mode accounting for a gap here seems like the best option.
Screenshot 2025-06-18 at 7 55 39 AM

To complete the analysis: All other metric reporter types account the gap correctly as plain missing data when it comes to raw data. However average calculations all flake, as the time window that is looked at has a huge gap.

Getting from energy to power becomes thus impossible without somehow removing the gap from the data.
Question is: Do we want to analyze the resulting measurement data and "classify" gaps ... which can be error prone.
Or just render the measurement as failed if a suspend occurs? And we just keep the behaviour of going into suspend as a non-breaking option for debug purposes?

Addon: Example screenshot of detected sampling rate which is heavily off and we could use it for detecting a suspend.
Screenshot 2025-06-18 at 8 10 41 AM

@ribalba ?

### DIFF
--- a/lib/c/gmt-lib.c
+++ b/lib/c/gmt-lib.c
@@ -39,8 +39,8 @@ void get_time_offset(struct timespec *offset) {
         perror("clock_gettime CLOCK_REALTIME");
         exit(EXIT_FAILURE);
     }
-    if (clock_gettime(CLOCK_MONOTONIC_RAW, &monotonic) != 0) {
-        perror("clock_gettime CLOCK_MONOTONIC_RAW");
+    if (clock_gettime(CLOCK_BOOTTIME, &monotonic) != 0) {
+        perror("clock_gettime CLOCK_BOOTTIME");
         exit(EXIT_FAILURE);
     }
     offset->tv_sec = realtime.tv_sec - monotonic.tv_sec;
@@ -53,8 +53,8 @@ void get_time_offset(struct timespec *offset) {
 
 void get_adjusted_time(struct timeval *adjusted, struct timespec *offset) {
     struct timespec now_monotonic;
-    if (clock_gettime(CLOCK_MONOTONIC_RAW, &now_monotonic) != 0) {
-        perror("clock_gettime CLOCK_MONOTONIC_RAW");
+    if (clock_gettime(CLOCK_BOOTTIME, &now_monotonic) != 0) {
+        perror("clock_gettime CLOCK_BOOTTIME");
         exit(EXIT_FAILURE);
     }
 


ribalba:
Ahh, we didn't think about that. 

So I see multiple point here that I don't think we can solve with one clock
1) We want the measurements to not be skewed by NTP or clock changes when the system changes based on Summer/Winter time, etc. As this makes the energy measurements wrong
1) We want to enable systems to sleep while in measurement. This is mainly for the blue angel as a proof not really for benchmarking where we don't actually want this
1) We want the measurement time to match the system time. (This is in contrast to point 1)
1) We want the syscall to be very low overhead

I think the most important thing is that the measurements in energy are correct. I think a drift between time of the system and measurement time is better than calculating a wrong energy value. 

I think that suspend is something people should not use while measuring in real life. I think we should have a feature that shows that this is possible for the blue angel but this is such a small portion of the runs (1 for each blue angel cert) that it is really not crucial and it only needs to show that the software supports this. Which probably 99.9% do as the suspend is handled by the OS.

Something I don't know is how virtualized OSes handle clocks. I know that some Hypervisors do funky stuff there especially when moving live machines. But this is something I would see as out of scope for the GMT.  

(**I think** a lot)

ArneTR:
So if I summarize correctly:
- Leave GMT as it is with the monotonic clock
- Issue a warning when suspend is detected
- Document option to use CLOCK_BOOTIME when wanting to document suspend behaviour, but do not go the extra step and patch metric providers to correctly handle this time skew but rather mark such measurements as invalid?

ribalba:
yes

PR #1236, opened 2025-06-28
Fix error handling for clipboard API is unavailable

davidkopp:
The `copyToClipboard` method had several issues:
- the alert command was unreachable, because it came after the return statement (therefore the warning "unreachable code after return statement" was displayed in the dev console)
- return value is not required, because the event listeners don't use return values for event handling
- the error message was wrong: the clipboard API is actually available locally, `localhost:9142` works, but not `http://metrics.green-coding.internal:9142`
- the clipboard API is async

### DIFF
--- a/frontend/js/helpers/main.js
+++ b/frontend/js/helpers/main.js
@@ -199,14 +199,20 @@ const showNotification = (message_title, message_text, type='warning') => {
     return;
 }
 
-const copyToClipboard = (e) => {
+const copyToClipboard = async (e) => {
   e.preventDefault();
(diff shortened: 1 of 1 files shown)


ArneTR:
Thanks for the PR. Looks cleaner now.

Did you experience any problems though with the implementation? 
To my understanding the effective functionality is the same, or?

davidkopp:
No, there is a difference in the functionality in case the clipboard API is not available. Before, no error was displayed because of the `return false`. Now an error is displayed. I came across this issue because in the browser's dev console there was the warning "unreachable code after return statement".

PR #1247, opened 2025-07-14
204 special handling

ArneTR:
This PR changes the behaviour when no data is present for:
- Old Eco CI repositories / Eco CI URLs with recent dates
- Old Watchlist Items / Timeline URLs with recent dates

Before this is what would show:

Screenshot 2025-07-14 at 1 43 41 PM

Which was hard for the user to understand if a terminal error has happend or what not.

Effectively the API was just returning a HTTP 204 and there might be well data with a different time frame.

Behaviour has now been moved to this:
Screenshot 2025-07-14 at 1 40 58 PM

and
Screenshot 2025-07-14 at 1 40 06 PM

## Alternative implementation explored

We also discussed the implementation of just "showing what we have". This would be for instance the "last 30 measurements"

While this is theoretically possible I faced multiple technical challenges as well as some UX challenges:
- Code for making queries internally is re-used in multiple functions. Date was so far a **mandatory** variable. Making this optional would have been quite an implementation
- Since APIs where expected to return data ascending for a certain time frame it would have made code way more complex and harder to read if there was also an option to sort-inverse => last 30 items => sort-inverse-again. 
- We also make statistics over a defined period of time - If we change that to the last 30 items the user must be aware of what period this is now shown as the aggregate data lacks time information. You could simply take the last 30 runs and then play the period back to the front end. However, this would require a great deal of effort and would always lead to a different result for the user. Typically the view is always 30 days. You then suddenly have MORE than 30 days and so on. Yes, you can again cap the view at 30 days. But the round trips are getting more and more. The question is what do we actually want

This is the solution chosen.

@ribalba : Do you think that solves the user confusion? What would you like to add?

## Greptile Summary

This PR improves the user experience when no data is available for a given time period in the Green Metrics Tool. Previously, when querying old Eco CI repositories or watchlist items with recent dates, users would receive an unclear HTTP 204 response with no indication of why data was missing. The changes implement a more user-friendly approach across multiple frontend files.

Key changes:
1. Added a custom `APIEmptyResponse204` error class in `main.js`
2. Implemented consistent error handling across `timeline.js`, `ci.js`, and related files
3. Added clear UI feedback with red message banners explaining why no data is shown
4. Added consistent CSS classes (`container-no-data`) for better state management

The developer also explored an alternative approach of "showing what we have" (last 30 measurements) but rejected it due to technical complexity and potential UX inconsistencies with time-based aggregations.

## Confidence score: 5/5
1. This is a safe UI improvement that doesn't affect core functionality
2. The changes are well-structured, consistent across files, and improve user experience without introducing new risks
3. Key files to review:
   - frontend/js/helpers/main.js (error handling changes)
   - frontend/timeline.html and frontend/ci.html (UI message changes)
   - frontend/js/timeline.js and frontend/js/ci.js (implementation of new error handling)

<sub>5 files reviewed, 1 comment</sub>
<sub>[Edit PR Review Bot Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=green-metrics-tool_1247)</sub>

### DIFF
--- a/frontend/ci.html
+++ b/frontend/ci.html
@@ -33,7 +33,7 @@
     <gmt-menu></gmt-menu>
     <div class="main ui container" id="main">
         <h1 class="ui header float left"><a id="menu-toggle" class="opened"><i class="bars bordered inverted left icon opened"></i></a> CI Run Info</h1>
-        <div class="ui full-width-card card">
+        <div class="ui full-width-card card container-no-data">
             <div class="content">
                 <div class="header"><a class="ui red ribbon label">
                     <h3>General Info</h3>
@@ -83,7 +83,19 @@
                 </div>
             </div>
         </div><!-- end ui full-width-card card -->
-        <div class="ui icon message blue">
+
+        <div id="message-no-data" class="ui icon message red container-no-data">
+            <i class="info circle icon"></i>
+            <div class="content">
+                <div class="header">
+                    No data for time frame
+                </div>
+                <p>In the selected time frame we have not seen any data :/</p>
+                <p>Try selecting a longer period reaching deeper into the past to see data for this repo.</p>
+            </div>
+        </div>
+
+        <div class="ui icon message blue container-no-data">
             <i class="info circle icon"></i>
             <div class="content">
                 <p> Here you can see the energy measurements of the associated CI/CD pipeline. </p>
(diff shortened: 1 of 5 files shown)


ribalba (approved):
perfect. This resolves the confusion. Good solution

PR #1248, opened 2025-07-14
Allow overwrite of str in usage scenario

ArneTR:
This PR unifies how GMT overwrites included keys from the compose file.

Before basic top level keys where NOT overwritten. For instance 'name'.

However everything that is in *usage_scenario.yml* should have precedence. Which is now the case

## Greptile Summary

This PR modifies how the Green Metrics Tool (GMT) handles configuration overrides between compose files and usage scenarios. The key changes are in the `merge_dicts` function within `scenario_runner.py` and related test updates.

Previously, only nested dictionary values from usage_scenario.yml would override compose file values, while top-level string keys (like 'name', 'author') would retain their compose file values. This created inconsistent behavior where some values could be overridden but others couldn't.

The change ensures that ALL values from usage_scenario.yml take precedence over compose files, making the behavior more consistent and logical. This means users can now reliably override any configuration value by specifying it in their usage scenario file.

The changes also improve the testing framework by moving validation checks earlier in the process (switching from 'setup_services' to 'import_metric_providers' phase) and adding new test cases to verify the string value overwriting behavior.

## Confidence score: 4/5
1. This PR is safe to merge as it maintains backward compatibility while improving consistency
2. The changes are well-tested with clear test cases and the modification is focused on configuration handling rather than core measurement functionality
3. Key files to review:
   - lib/scenario_runner.py: Verify merge_dicts logic
   - tests/test_usage_scenario.py: Ensure test coverage is comprehensive

<sub>2 files reviewed, no comments</sub>
<sub>[Edit PR Review Bot Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=green-metrics-tool_1248)</sub>

### DIFF
--- a/tests/data/usage_scenarios/include_example.yml
+++ b/tests/data/usage_scenarios/include_example.yml
@@ -1,0 +1,4 @@
+---
+name: Name as is
+author: Author as is
+description: Description as is
--- a/tests/data/usage_scenarios/overwrite_string_from_include.yml
+++ b/tests/data/usage_scenarios/overwrite_string_from_include.yml
@@ -1,0 +1,13 @@
+---
+name: Name overwritten
+author: Author overwritten
+
+compose-file: !include include_example.yml
+
+flow:
+  - name: dummy
+    container: test-container-1
+    commands:
+      - type: console
+        command: pwd
+
--- a/tests/data/usage_scenarios/overwrite_string_from_include_even_if_top.yml
+++ b/tests/data/usage_scenarios/overwrite_string_from_include_even_if_top.yml
@@ -1,0 +1,13 @@
+---
+compose-file: !include include_example.yml
+
+name: Name overwritten
+author: Author overwritten
+
+flow:
+  - name: dummy
+    container: test-container-1
+    commands:
+      - type: console
+        command: pwd
+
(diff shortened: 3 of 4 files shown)


lib/scenario_runner.py
```diff
@@ -369,7 +371,7 @@ def merge_dicts(dict1, dict2):
                         else:
                             dict1[k] = v
                     return dict1
-                return dict1
+                return dict2
```
ribalba, line 374: Is this on purpose? Because we set the new keys in `dict1`

ArneTR: I think this is the bug in the code, this is why I pinged you to look at it.

To my understanding this branch of the recursion is a terminal condition and will only be run if a string key is encountered that has no children. In that case you want to return the key from the usage_scenario.yml

In the other branches you are merging data in from the compose.yml in case the key exists in both.

I hope I understood your code correctly. Please give it a detailed look again

ArneTR:
> Hmm, is this the behavior we want? The compose file does not have `name` or `description` as far as I know. The idea was that you import a std compose file and then overwrite the things you need for the compose to work in the GMT. Why do you want name to be overwritten in the compose file?

Compse has a name attribute. This will be conflicting. This was the reason for this PR.

In general I would argue: Whatever is in the usage_scenario.yml should be the final values. It contains the dominant values. Or?

lib/scenario_runner.py
ribalba: Yes. I agree that it looks like an error. Good catch

ribalba (approved):

PR #1263, opened 2025-08-01
More user measurement configurations

ArneTR:
This PR introduces more user configurations for the measurement.

For instance it is very helpful in developing to turn `system_check_threshhold` off on the cluster.

Also setting different baseline times and idle times is very helpful when wanting to have longer reference times.

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -77,12 +77,6 @@ machine:
   base_temperature_feature: False
 
 measurement:
-  system_check_threshold: 3 # Can be 1=INFO, 2=WARN or 3=ERROR
-  pre_test_sleep: 5
-  idle_duration: 10
-  baseline_duration: 5
-  post_test_sleep: 5
-  phase_transition_time: 1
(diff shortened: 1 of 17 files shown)


ArneTR:
I reviewed now with Gemini. This was fine. Happy to see the Codex review still. Merged for now.

This is what Gemini said:

> Here's my overall review of the PR:

>  Positive:

>   * The PR adds a lot of new measurement settings, which
>     gives users more control over the measurement process.
>   * The new settings are exposed in the UI and can be
>     configured by the user.
 >  * The changes are well-tested.

 > Negative:

  > * The runner.py script does not correctly use the user's
>     settings from the database. This is a major issue that
>     needs to be fixed.
  > * The PR is quite large and touches many files. It would
  >   have been better to split it into smaller, more focused
   >  PRs.

ribalba:
Codex can't do it. I have tried now for days and it just can't

PR #1265, opened 2025-08-04
Restructure runner unit tests

### LINKED ISSUE #1264: Migrate unit tests to test_runner.py which are not usage scenario exclusive
as discussed in https://github.com/green-coding-solutions/green-metrics-tool/pull/1176

davidkopp:
Closes #1264

Additionally this PR includes some cleanups.
Reviewing the PR is probably easiest if done commit by commit.


davidkopp:
Yes, the reduced number of tests in on purpose:

1. `test_uri_github_repo` + `test_runner_uses_default_filename` -> `test_uri_github_repo_and_using_default_filename`
   
   Reason: The `test_runner_uses_default_filename` used an URI that is not a Git repository to be able to go to a directory that includes a `usage_scenario.yml` (without the `--dev-no-save` flag it would be flagged as invalid). Because `test_uri_github_repo` already uses the logic of using the default filename, I thought it makes sense to consolidate them.

2. `test_runner_filename_with_remote_uri` was removed, because `test_uri_github_repo_and_using_default_filename` already tested the same functionality.

ArneTR:
👍

PR #1268, opened 2025-08-04
Add direnv for automatic enabling of venv

davidkopp:
I still sometimes forget to execute "source venv/bin/activate". This PR proposes an automatic way of executing it whenever you go into the directory using [direnv](https://direnv.net/).

image

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+source venv/bin/activate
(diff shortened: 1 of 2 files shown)


davidkopp:
Thanks for this custom bash cd function. I will use it instead of direnv. Therefore I close this PR.

Issue #1267, opened 2025-08-04
Cgroup System Metric Providers Cannot Handle Cgroup Names/Paths


davidkopp:
## Summary

The cgroup system metric providers are unable to handle cgroup names or paths as intended, making them unusable for system-level monitoring of services, sessions, and other non-container cgroups.

## Problem Description

4 out of 5 cgroup system providers are symlinked to their container counterparts, causing them to expect Docker container IDs instead of cgroup names/paths. The one with custom implementation (`cpu/time/cgroup/system`) don't accept any argument at the moment (see other issue https://github.com/green-coding-solutions/green-metrics-tool/issues/1266).

### Expected Behavior

According to `config.yml.example` lines 151-157, system cgroup providers should accept cgroup/service names like:
- `"org.gnome.Shell@wayland.service"`  
- `"session-2.scope"`
- `"system.slice/docker-xyz.scope"`

### Actual Behavior

All system providers fail with cgroup names and only work with extracted Docker container IDs.

## Technical Root Cause

All cgroup system providers use `detect_cgroup_path(controller, user_id, container_id)` which:
1. Expects the 3rd parameter to be a Docker container ID (64-char hex string)
2. Constructs Docker-specific paths like `/sys/fs/cgroup/system.slice/docker-{id}.scope/{controller}`
3. Cannot handle arbitrary cgroup paths or systemd service names

### Test Results
```bash
# ❌ Fails with cgroup path (as intended by config examples)
./metric-provider-binary -s "system.slice/docker-2f230ab...scope"
# Error - Could not open container for reading: system.slice/docker-2f230ab...scope

# ✅ Works with container ID (not the intended use case)  
./metric-provider-binary -s "2f230ab321f7d672285deaf9de1e2d97d2a894883ee62d9243528b2b451bd916"
# 1754318223234474 571 2f230ab321f7d672285deaf9de1e2d97d2a894883ee62d9243528b2b451bd916
```

### Affected Providers
- `cpu/utilization/cgroup/system` (symlink → `../container/source.c`)
- `memory/used/cgroup/system` (symlink → `../container/source.c`)  
- `network/io/cgroup/system` (symlink → `../container/source.c`)
- `disk/io/cgroup/system` (symlink → `../container/source.c`)
- `cpu/time/cgroup/system` (custom implementation, neither accepts a container ID nor a cgroup name)

### 1. Create new function in `lib/c/detect_cgroup_path.c`:
```c
[...]

PR #1270, opened 2025-08-05
Detect systemd cgroup path using cgroup name (instead of container id)

davidkopp:
Closes https://github.com/green-coding-solutions/green-metrics-tool/issues/1267

The change was tested successfully:

image

Between the two statements the binary was recompiled with the change introduced in this PR.

This PR includes changes to all Makefiles to add the missing dependency declarations. I had the problem, that re-compiling the metric-provider-binary lead to the message: "'metric-provider-binary' is up to date"
The changed `detect_cgroup_path.o` was not recognized due to the missing dependency declaration. I hope it is ok to include this change here. If not, let me know.

### DIFF
--- a/lib/c/detect_cgroup_path.c
+++ b/lib/c/detect_cgroup_path.c
@@ -85,6 +85,15 @@ char* detect_cgroup_path(const char* controller, int user_id, const char* id) {
         return path;
     }
 
+    // Try cgroups v2 with systemd but non-slice mountpoints and with cgroup name instead of container id (used for debug purposes)
+    snprintf(path, PATH_MAX,
+             "/sys/fs/cgroup/system.slice/%s/%s",
+             id, controller);
+    fd = fopen(path, "r");
+    if (fd != NULL) {
+        fclose(fd);
+        return path;
+    }
 
     // If no valid path is found, free the allocated memory and error
     free(path);
--- a/lib/c/detect_cgroup_path.c
+++ b/lib/c/detect_cgroup_path.c
@@ -85,9 +85,9 @@ char* detect_cgroup_path(const char* controller, int user_id, const char* id) {
         return path;
     }
 
-    // Try cgroups v2 with systemd but non-slice mountpoints and with cgroup name instead of container id (used for debug purposes)
+    // Try cgroups v2 with full cgroup name (typically used for debug purposes)
     snprintf(path, PATH_MAX,
-             "/sys/fs/cgroup/system.slice/%s/%s",
+             "/sys/fs/cgroup/%s/%s",
              id, controller);
     fd = fopen(path, "r");
     if (fd != NULL) {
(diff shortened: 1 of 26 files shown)


metric_providers/cpu/energy/rapl/msr/component/Makefile
```diff
@@ -1,6 +1,6 @@
 CFLAGS = -O3 -Wall -Werror -lm -I../../../../../../lib/c
 
-metric-provider-binary: source.c
+metric-provider-binary: source.c ../../../../../../lib/c/gmt-lib.o
```
ArneTR: To my understanding this  will track the change and rebuild metric-provider-binary if the .o file has changed. But it will also try to rebuild the o-file if  their source has changed, not? I wonder if you can build the source from here ... are all the headers / includes set?

Testing would be that you only change the source of the gmt-lib.c and see if the make of the metric-provider-binary rebuilds fully.

lib/c/detect_cgroup_path.c
```diff
@@ -85,6 +85,15 @@ char* detect_cgroup_path(const char* controller, int user_id, const char* id) {
         return path;
     }
 
+    // Try cgroups v2 with systemd but non-slice mountpoints and with cgroup name instead of container id (used for debug purposes)
+    snprintf(path, PATH_MAX,
+             "/sys/fs/cgroup/system.slice/%s/%s",
```
ArneTR: I thought more that the string should be "%s/%s" to make it fully flexible. This is now restricted to groups in system.slice ?

ArneTR (changes requested):

metric_providers/cpu/energy/rapl/msr/component/Makefile
davidkopp: Correct, the rebuild happens if the source of the dependency has changed. I tested it with gmt-lib.c: If I add a comment, a rebuild happens. However, a change in the header file does not lead to a rebuild. If that should be the case, the header file has to be added to it too as a dependency:

```sh
metric-provider-binary: source.c ../../../../../lib/c/gmt-lib.o ../../../../../lib/c/gmt-lib.h
```

Should I do that?

lib/c/detect_cgroup_path.c
davidkopp: you're right 😊 I changed it now

metric_providers/cpu/energy/rapl/msr/component/Makefile
ArneTR: hmm actually unsure. I have never seen header files as dependencies. @ribalba Do you have insights here?

ribalba: I would include the header files too. What we want is that if the functionality would change you should recompile which can happen if you change the `.h` file. 

Let's say you change some buffer size in the `.h` file and then want to test the metric provider you would expect it to have the new buffer size.

davidkopp: Done. I added the header files as dependencies. I also replaced the long paths to the lib/c directory with a variable.

PR #1273, opened 2025-08-05
Document the usage of cgroup system metric providers

davidkopp:
Corresponding changes in documentation repo: https://github.com/green-coding-solutions/documentation/pull/117

Related to discussion in issue https://github.com/green-coding-solutions/green-metrics-tool/issues/1267

New metric providers in config.yml were tested successfully together with PR https://github.com/green-coding-solutions/green-metrics-tool/pull/1270, with one exception: the cgroup system network provider has an issue (I will create a new PR for that).

PR #1274, opened 2025-08-05
Fix NetworkIoCgroupSystemProvider doesn't create a value column

davidkopp:
Issue: NetworkIoCgroupSystemProvider doesn't implement the `_parse_metrics` method to create a value column, while NetworkIoCgroupContainerProvider does. The base metric processing code expects all metric providers to have a single value column.

Fix: The NetworkIoCgroupSystemProvider uses the same `_parse_metrics` method as the working NetworkIoCgroupContainerProvider. This creates the required value column by combining received and transmitted bytes intervals.

See successful test in https://github.com/green-coding-solutions/green-metrics-tool/pull/1273

### DIFF
--- a/metric_providers/network/io/cgroup/system/provider.py
+++ b/metric_providers/network/io/cgroup/system/provider.py
@@ -1,5 +1,6 @@
 import os
 
+from lib import utils
 from metric_providers.cgroup import CgroupMetricProvider
 
 class NetworkIoCgroupSystemProvider(CgroupMetricProvider):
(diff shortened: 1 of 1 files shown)


davidkopp:
Actually the code did not work on my system with WSL2.

I had to replace

```python
        if (df['received_bytes_intervals'] < 0).any():
            raise ValueError('NetworkIoCgroupContainerProvider data column received_bytes_intervals had negative values.')

        if (df['transmitted_bytes_intervals'] < 0).any():
            raise ValueError('NetworkIoCgroupContainerProvider data column transmitted_bytes_intervals had negative values.')
```

with 

```
        # Handle counter wraparounds by setting negative intervals to 0
        df.loc[df['received_bytes_intervals'] < 0, 'received_bytes_intervals'] = 0
        df.loc[df['transmitted_bytes_intervals'] < 0, 'transmitted_bytes_intervals'] = 0
```

Is this just an environment problem on my side or something the metric provider should handle like the code used here?

PR #1278, opened 2025-08-07
Micro phases

ArneTR:
*Micro Phases* are a concept in GMT to dynamically create phases from *stdout* of running processes.

Example: You have such a usage scenario:

```yaml
...
flow:
    - name: Run Playwright Tests
       container: playwright
       commands:
       - type: console
          command: node playwright.js
...
```

The Playwright browser might run multiple tests but in GMT they would only show up as one phase.
However by reading and parsing the stdout alongside timestamps set there GMT can now *expand* them to separate phases.

This works by using a regex on the *stdout* and capturing three groups:
- phase_name
- start_time
- end_time

A working *stdout* can for instance look something like this:

```log
# node playwright.js
...
Starting (Test No. 1) at 179874192798921 ;
Ending (Test No. 1) at  179874192798929 ;
Starting (Test No. 2) ....
....
```

The flow would then updated with a regex like so:

```flow
```yaml
...
flow:
    - name: Run Playwright Tests
       container: playwright
       commands:
       - type: console
          command: node playwright.js
          log-stdout: true
          sub-phase-expansion-pattern: '^\s*Starting \((?P<phase_name>.+)\) at (?P<start_time>\d+) ;$\n^\s*Ending \([^\)]+\) at (?P<end_time>\d+) ;$'
...
```

The implementation atm requires strong regex knowledge and, if not already present, instrumentation of the stdout to contain the needed timestamps and event names that shall be used as phase names.

## Example of how phase names are then later expanded
Screenshot 2025-08-21 at 8 37 23 AM

## Open Problems

- The newly created phases are sub-phases of exisiting runtime-sub-phases. So technically sub-sub-phases (calling them Micro Phases for now). Which means they must be removed from accounting in phase_stats.py as currently the totals are off (double-counting)
- The display can get quickly garbled if more than 10 sub-sub-phases are expanded. Maybe a new separate tab should be explored here
- The Micro Phases are not started and ended by GMT directly but just mapped. Which means they cannot make use of the phase-padding feature. They will have hard cut-offs and suffer from possible undersampling issues
- Micro Phases can overlap as the parsed stdout is currently not checked against that. It shall be reasoned if that is wanted behaviour or shall be an error


ArneTR:
This PR was also used to evaluate integrating GitHub Actions into GMT. The corresponding repo describing why we believe this is currently only possible with a very high effort integration is at : https://github.com/green-coding-solutions/eco-ci-runner-inside-gmt

PR #1279, opened 2025-08-07
New metric boxes

ribalba:
This is the first version that is not totally complete. Needs testing on machines with more metrics providers enabled. Also I am sure @ArneTR will want some changes :)


ArneTR:
I will fix the failing test in review

frontend/js/helpers/metric-boxes.js
```diff
@@ -1,180 +1,104 @@
...
-            <div class="ui card network-co2">
-                <div class="ui content">
-                    <div class="ui top black attached label overflow-ellipsis">Network Transmission CO2 <span class="si-unit"></span></div>
-                    <div class="description">
-                        <div class="ui fluid mini statistic">
-                            <div class="value">
-                                <i class="burn icon"></i> <span>N/A</span>
+                    <div class="extra content">
+                        <div class="description">
+                            <span class="value">N/A</span>
```
ArneTR: Would prefer the value in bold. Seems a bit thin atm

frontend/js/helpers/metric-boxes.js
```diff
@@ -361,40 +285,82 @@ const calculateCO2 = (phase, total_CO2_in_ug) => {
...
+            else if (isCO2) selector = `.${component}-co2`;
+        }
     }
 
+    if (selector === null) {
+        return;
+    }
 
-    document.querySelector(`div.tab[data-tab='${phase}'] ${selector} .value span`).innerText = `${value} ${std_dev_text}`
+  const cards = document.querySelectorAll(selector);
```
ArneTR: You are significantly changing the selector here. Is that intended?

I am open for performance tweaks, but to my understanding this implimentation breaks the functionality.

When I import demo data (`python3 tools/import_data.py data/demo_data.sql`) and then click on Baseline -> Idle -> Baseline the values are beginnging to be all the same. They do not change anymore.
It seems values are changed once on referesh, but then are frozen in time. I suspect this line is the culprit

frontend/js/helpers/metric-boxes.js
```diff
@@ -361,40 +285,82 @@ const calculateCO2 = (phase, total_CO2_in_ug) => {
...
 
-    document.querySelector(`div.tab[data-tab='${phase}'] ${selector} .value`).setAttribute('title', `${raw_value} [${raw_unit}]`)
+    if (cards.length === 0) {
+        console.warn(`No card found for selector "${selector}" in phase "${phase}"`);
```
ArneTR, line 304: Shouldnt' this be more prominent to the user? Maybe trigger a toast also indicating a major error has happend?

frontend/js/helpers/metric-boxes.js
ArneTR: The same happens when not clicking the phases, but the sub-phases in the runtime phase

frontend/js/helpers/metric-boxes.js
```diff
@@ -1,180 +1,104 @@
...
+                { key: 'disk', name: 'Disk', icon: 'hdd', colour: 'orange', variable:true  },
+                { key: 'machine', name: 'Machine', icon: 'server', colour: 'orange', variable:true },
+        ];
+
+        const extraCards = [
+            { key: 'runtime', name: 'Phase Duration', icon: 'hourglass half', colour: 'green', variable:false },
+            { key: 'network-data', name: 'Data Transferred', icon: 'exchange', colour: 'green', variable:false },
+            { key: 'network', name: 'Network', icon: 'sitemap', colour: 'green', variable:true },
+            { key: 'embodied-carbon', name: 'Embodied Carbon', icon: 'industry', colour: 'green', variable:false },
+            { key: 'operational-carbon', name: 'Operational Carbon', icon: 'fire', colour: 'green', variable:false },
```
ArneTR: (This location of the comment is not really accurate)
THe machine-co2 card does not switch to power. So atm there is no possiblity for me to see the total machine energy.

Since I deem this metric the most important I would argue and it is a direct controllable one I would argue for moving it to the "hardware" row and remove it from the impact row also.

ArneTR (changes requested):

frontend/js/helpers/metric-boxes.js
ArneTR: FYI: This prevented me from fixing the test. Will gladly do so after

ArneTR:
Also one more request: It would be nice if the color line at the bottom of the boxes of the boxes changes when tabbing power -> energy -> co2.
The color I imagine are orange -> blue -> black.

ArneTR:
> @gemini-cli /review

@davidkopp @ribalba This should now trigger gemini PR reviews. Please try it out next time you make a PR. It should also work automatically on new PRs ... untested though.

Currently Google seems overloaded ... so double check the actions if nothing happens (Failed example due to overload: https://github.com/green-coding-solutions/green-metrics-tool/actions/runs/16824036563)

frontend/js/helpers/metric-boxes.js
ribalba: Good catch. I took that out while debugging and forgot to put it back in

frontend/js/helpers/metric-boxes.js
ribalba: I think people will not understand what it means. I would leave it in for us to debug but would also be ok with removing it totally

frontend/js/helpers/metric-boxes.js
ribalba: Fixed. Needed a `&&` instead of an `||`

frontend/js/helpers/metric-boxes.js
```diff
@@ -361,40 +285,82 @@ const calculateCO2 = (phase, total_CO2_in_ug) => {
...
+        else if (metric_name.includes('memory') || metric_name.includes('dram')) component = 'dram';
+        else if (metric_name.includes('gpu')) component = 'gpu';
+        else if (metric_name.includes('disk')) component = 'disk';
+        else if (metric_name.includes('psu') || metric_name.includes('machine')) component = 'machine';
+        else if (metric_name.includes('network')) component = 'network';
+
+        if (component !== null) {
+            if (isPower) selector = `.${component}-power`;
+            else if (isEnergy) selector = `.${component}-energy`;
+            else if (isCO2) selector = `.${component}-co2`;
```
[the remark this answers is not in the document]
ribalba: Arne already saw that

frontend/js/helpers/metric-boxes.js
```diff
@@ -361,40 +285,82 @@ const calculateCO2 = (phase, total_CO2_in_ug) => {
...
 }
 
-const updateKeyMetric = (phase, metric_name, clean_name, detail_name, value, std_dev_text, unit, raw_value, raw_unit, explanation, source) => {
+const updateKeyMetric = (
+    phase, metric_name, clean_name, detail_name,
+    value, std_dev_text, unit, raw_value, raw_unit,
+    explanation, source
+) => {
 
     let selector = null;
```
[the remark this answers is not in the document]
ribalba: good catch

frontend/js/helpers/config.js.example
```diff
@@ -60,6 +60,12 @@ const network_energy_metric_condition = (metric) => {
     return false;
 }
 
+const network_io_metric_condition = (metric) => {
+    if(metric == 'network_io_cgroup_container') return true;
+    return false;
```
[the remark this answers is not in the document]
ribalba: nice, refactored

frontend/js/helpers/metric-boxes.js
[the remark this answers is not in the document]
ribalba: ?

ribalba:
@ArneTR put in all the stuff we talked about. The only thing I didn't come up with a nice way of doing is changing the name to "GPU Power" if there is no power reading for GPU. If this is important let me know and I will write a little for loop that does this at the end. Seemed "unclean" though. 

Also I couldn't test with RAM, GPU and Disk readings. Maybe we should add this to the example data?

ArneTR:
Here a couple of more ideas with moving the unit down to the number and putting the source in a label in the top right
Screenshot 2025-08-12 at 10 16 35 AM
Screenshot 2025-08-12 at 10 16 56 AM
Screenshot 2025-08-12 at 10 17 02 AM

ArneTR:
@gemini-cli /review

davidkopp:
Looks good 👍

Two things:
- The placement of the unit is a little bit strange. The unit follows after the percentage sign of the std.dev.
- What is the meaning of the colors? I don't understand the meaning of orange and green.

ArneTR:
PR should also include fix for: https://github.com/green-coding-solutions/green-metrics-tool/issues/1287

ArneTR:
@davidkopp The colors are arbitrary, but should be consistent:
- Orange: Power
- Blue: Energy
- Black: Carbon
- Purple: Time
- Green: Not set atm ... can be network data for instance (I would rather go for teal TBH and not use green at all)

Are you ok with these arbitrary colors? Or does one trip you off?

@ribalba: can you make them consistent and add in Davids response if any

frontend/js/helpers/metric-boxes.js
```diff
@@ -312,14 +313,24 @@ const updateKeyMetric = (
         const unitNode = card.querySelector('.si-unit');
         if (unitNode) unitNode.innerText = unit;
 
+        const typeNode = card.querySelector('.metric-type');
+
+        if(std_dev_text != ''){
+            if (typeNode) typeNode.innerText = `(AVG + STD.DEV)`;
+        } else {
+            if(String(value).indexOf('%') !== -1) {
+                if (typeNode) typeNod.innerText = `(Diff. in %)`;
```
ArneTR: Here is a typo. Must be typeNode not typeNod

davidkopp:
> The colors are arbitrary, but should be consistent:
> 
>     * Orange: Power
>     * Blue: Energy
>     * Black: Carbon
>     * Purple: Time
>     * Green: Not set atm ... can be network data for instance (I would rather go for teal TBH and not use green at all)
> 
> Are you ok with these arbitrary colors? Or does one trip you off?

I also think, that the green color should not be used. Green implies for me the value is somehow good. But the colors here don't want to give any grading.

To make it more clear what the colors are standing for (at least power, energy and carbon), you may add them to the tab labels.

> Answering also to the % preceding the unit: I think this is how it should be .... Are you used to "20 W%"? I am more used to "20% W"

I'm not used to both 🙈 But I will get used to it 😊

ArneTR:
alright. The fix is a bit slammed in and obfuscates if a value is null now ... which is a bit unwanted, but I guess it will do for now. I will add a comment in the code that a slightly more targeted fix would be to not render the title if a certain compare mode is active.

PR #1280, opened 2025-08-08
Updating echarts to 5.5.1 and using common build

ArneTR:
@ribalba 
@davidkopp 

Can you please if you experience any errors with these charts?

We have no visual tests to verify and furthermore echarts does not throw errors when a feature is missing suprisingly. I used the "simple" variant of the package before and then hover notes and the toolbox in the top right was missing.
This "common" version I am using is now a bit smaller than the one before and seems to have all we need.

To inject demo data into a fresh install run `python3 tools/import_data.py data/demo_data.sql`


ArneTR:
Closed as https://github.com/green-coding-solutions/green-metrics-tool/pull/1282 was merged in favor

PR #1282, opened 2025-08-09
Updated echarts to v.6.0

ArneTR:
Alternative PR to https://github.com/green-coding-solutions/green-metrics-tool/pull/1280

@ribalba 
@davidkopp 

an you please if you experience any errors with these charts?

We have no visual tests to verify and furthermore echarts does not throw errors when a feature is missing suprisingly. I used a custom variant of the package which is commented in the JS file at the top.

To inject demo data into a fresh install run python3 tools/import_data.py data/demo_data.sql

Note: echarts has breaking changes in v6 ... on GitHub it reads like we might not be too affected. Only a different legend might be problematical. Did not spot it thouth so far https://github.com/apache/echarts/releases/tag/6.0.0


davidkopp:
LGTM 👍

I did not experience any errors. Only changes I have seen are visual:

- green color is brighter
  image
- StdDev has now a grey color instead of yellow
  image
  
btw: because you have copied the description from the old PR, the description of this PR includes wrong statements.

ribalba:
works without errors on my Mac with Brave

PR #1289, opened 2025-08-13
Add check if container image architecture is compatible

davidkopp:
Two of my colleagues have already encountered the issue of unintentionally using an ARM image as part of a usage scenario for a GMT measurement. Needless to say, the measurement run failed. However, the error message was confusing:

> Health check of container "kadai-postgres" failed terminally with status "unhealthy" after 0s. Health check errors: {"Status":"unhealthy","FailingStreak":0,"Log":[]}

I think it would make sense to add a compatibility check, so a proper error message can be provided.

With this PR the run would fail early directly after the image pull and the error message would be
> Architecture mismatch for image 'kadai-postgres': Image is built for arm64 but host platform is amd64.

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -803,12 +803,11 @@ class ScenarioRunner:
                         raise OSError(f"Docker pull failed. Is your image name correct and are you connected to the internet: {service['image']}")
 
                 # Check architecture compatibility after pull/verification
-                is_compatible, img_arch, host_arch, error_msg = self._check_image_architecture_compatibility(service['image'])
+                is_compatible, _, _, error_msg = self._check_image_architecture_compatibility(service['image'])
                 if not is_compatible:
                     print(TerminalColors.FAIL, f"ERROR: {error_msg}", TerminalColors.ENDC)
                     raise RuntimeError(error_msg)
 
-                print(f"Image {service['image']} architecture ({img_arch}) is compatible with host ({host_arch})")
 
                 # tagging must be done in pull and local case, so we can get the correct container later
                 subprocess.run(['docker', 'tag', service['image'], tmp_img_name], check=True)
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -274,6 +274,40 @@ class ScenarioRunner:
             self.__warnings.append(warn)
 
 
+    def _check_image_architecture_compatibility(self, image_name):
+        """
+        Check if the Docker image architecture is compatible with the host platform.
+        Returns (is_compatible: bool, image_arch: str, host_arch: str, error_message: str)
+        """
+        try:
+            # Get image architecture
+            result = subprocess.run(
+                ['docker', 'inspect', '--type=image', '--format={{.Architecture}}', image_name],
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='UTF-8', check=True
+            )
+            image_arch = result.stdout.strip()
+
+            # Get host architecture
+            host_arch = platform.machine()
+            # Normalize architecture names
+            arch_mapping = {
+                'x86_64': 'amd64',
+                'aarch64': 'arm64',
+                'armv7l': 'arm',
+            }
+            normalized_host_arch = arch_mapping.get(host_arch, host_arch)
+
+            is_compatible = image_arch == normalized_host_arch
+            if not is_compatible:
+                error_msg = (f"Architecture mismatch for image '{image_name}': "
+                            f"Image is built for {image_arch} but host platform is {normalized_host_arch}.")
+            else:
+                error_msg = ""
+
+            return is_compatible, image_arch, normalized_host_arch, error_msg
+        except subprocess.CalledProcessError as e:
+            return False, "unknown", "unknown", f"Failed to inspect image architecture: {e.stderr}"
+
     def _checkout_repository(self):
         print(TerminalColors.HEADER, '\nChecking out repository', TerminalColors.ENDC)
 
@@ -767,6 +801,14 @@ class ScenarioRunner:
                             raise OSError(f"Docker pull failed. Is your image name correct and are you connected to the internet: {service['image']}")
                     else:
                         raise OSError(f"Docker pull failed. Is your image name correct and are you connected to the internet: {service['image']}")
+
+                # Check architecture compatibility after pull/verification
+                is_compatible, img_arch, host_arch, error_msg = self._check_image_architecture_compatibility(service['image'])
+                if not is_compatible:
+                    print(TerminalColors.FAIL, f"ERROR: {error_msg}", TerminalColors.ENDC)
+                    raise RuntimeError(error_msg)
+
+                print(f"Image {service['image']} architecture ({img_arch}) is compatible with host ({host_arch})")
 
                 # tagging must be done in pull and local case, so we can get the correct container later
                 subprocess.run(['docker', 'tag', service['image'], tmp_img_name], check=True)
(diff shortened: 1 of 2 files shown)


ArneTR:
This is definitely something where the error message should be better.

I wonder though if GMT should have such a check routine and this information can always be provided by the docker cli and / or is always relevant.
What for instance happens if there is just a binary in the container?

To better understand: Can you tell which image was used so I can replay the error on a bare metal host and see why the Docker CLI even continues until the healthcheck without a more prominent error

davidkopp:
In the meantime the image was converted from an ARM-only image to a multi-arch image. Here the commands to get the image with the ARM architecture and check its architecture:
```sh
docker pull --platform linux/arm64 registry.gitlab.com/envite-consulting/sustainable-software-architecture/kadai/kadai-da
tabases:postgres-14.7-100k-tasks
docker image inspect -f "{{.Architecture}}" registry.gitlab.com/envite-consulting/sustainable-software-architecture/kadai/kadai-databases:postgres-14.7-100k-tasks
```

davidkopp:
Test is failing in the CI pipeline:
```
FAILED test_runner.py::test_architecture_compatibility_check_compatible - AssertionError: Expected: Architecture should be compatible, Actual: img_arch: unknown, host_arch: unknown
```

Before I invest time into that, let me know if GMT should include the check at all.

ribalba:
I think it should. This is something I see happening more and more.

ArneTR:
I also agree, such a check should exist.

The docker error message is very opaque and does not really indicate what the underlying problem is.

However I believe a check after downloading the image is needed, but not the most efficient.

In case an image will be downloaded it should be checked if the image even exists for the target architecture beforehand.

`docker manifest inspect` can do that. The JSON output contains all possible architectures. Assumption is then that the docker client will fetch the correct architecture automatically

Still a local check must be done afterwards as a built image might still contain a broken architecture ... very unlikely but to my understanding possible if it is based on a broken architecture already and only minor layers are added.

@davidkopp If you agree please update the PR with the manifest check also. Rest of the PR LGTM

davidkopp:
Executing `docker manifest inspect` takes a lot of time on my system. The time seems to increase proportionally with the number of manifests of the image.

## Examples

1 manifest:

```sh
$ time docker manifest inspect registry.gitlab.com/envite-consulting/sustainable-software-architecture/kadai/kadai-example-spring-boot:kadai-10.1.0
real    0m1.960s
user    0m0.044s
sys     0m0.035s
```

2 manifests:

```sh
$ time docker manifest inspect registry.gitlab.com/envite-consulting/sustainable-software-architecture/kadai/kadai-databases:postgres-14.7-100k-tasks
real    0m4.058s
user    0m0.055s
sys     0m0.081s
```

16 manifests:

```sh
$ time docker manifest inspect postgres:15
real    0m19.516s
user    0m0.249s
sys     0m0.174s
```

## Question

Is this normal? If so, I would vote against the use of `docker manifest inspect`.

Also, [docker manifest inspect](https://docs.docker.com/reference/cli/docker/manifest/inspect/) is an experimental feature.

davidkopp:
The outcome of a meeting with Arne was, that `docker manifest inspect` should not be used at the moment because of its slowness and because of rate limits of DockerHub.

Regarding the failed test: I added mocks to the test to ensure it works on all platforms. So there is no test anymore, that reads the actual CPU architecture from the host.

I think the PR is ready for merge.

ArneTR:
We are currently blocked to investigate this further inside the GitHub VM as our extension to analyse is broken.

We have opened an issue here: https://github.com/lhotari/action-upterm/issues/33


This PR is now waiting for response. If no response comes in we will continue to investigate through different means

ArneTR:
Nice to see the images that can be used. I tried using the index-hash which apparently auto-selects the correct architecture.

However I am still unsure if that would really lead to the error we saw back then.

When I run 
```bash
$ docker run --rm -it -u 0 --entrypoint timeout alpine@sha256:eafc1edb577d2e9b458664a15f23ea1c370214193226069eb22921169fc7e43f
Unable to find image 'alpine@sha256:eafc1edb577d2e9b458664a15f23ea1c370214193226069eb22921169fc7e43f' locally
docker.io/library/alpine@sha256:eafc1edb577d2e9b458664a15f23ea1c370214193226069eb22921169fc7e43f: Pulling from library/alpine
9824c27679d3: Pull complete
Digest: sha256:eafc1edb577d2e9b458664a15f23ea1c370214193226069eb22921169fc7e43f
Status: Downloaded newer image for alpine@sha256:eafc1edb577d2e9b458664a15f23ea1c370214193226069eb22921169fc7e43f
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
exec /usr/bin/timeout: exec format error
```

 I get a proper response from the daemon, that the image cannot be run.

However on my local box, which supports emulation via Docker Desktop, the image runs with only a minor warning.

So I feel two points are still unclear to me:
- Did we catch the error from back then? if so: With what GMT run string can I reproduce it on an old checkout of GMT?
- Should we fail GMT runs if technically the image can be run, but is only emulated? This PR enforces that. But it might be helpful in GMT to benchmark exactly that ...

davidkopp:
## Did we catch the error from back then?

Answer: yes

You can use the following command to reproduce the error when you are using the latest version of GMT on main (commit 949f445)[https://github.com/green-coding-solutions/green-metrics-tool/commit/949f445f69872ddea89239cf632776221bd795f1]).

On an amd64 host:

```sh
python3 runner.py --name "KADAI (using arm64 postgres image)" --uri "https://gitlab.com/envite-consulting/sustainable-software-architecture/kadai/kadai-resource-efficiency/" --branch "gmt-test-image-architecture" --filename "usage_scenario-arm64.yml" --skip-system-checks --dev-no-sleeps --dev-no-save --skip-unsafe
```

On an arm64 host:

```sh
python3 runner.py --name "KADAI (using amd64 postgres image)" --uri "https://gitlab.com/envite-consulting/sustainable-software-architecture/kadai/kadai-resource-efficiency/" --branch "gmt-test-image-architecture" --filename "usage_scenario-amd64.yml" --skip-system-checks --dev-no-sleeps --dev-no-save --skip-unsafe
```

Error:

```plain
Exception_context (NoneType): None
Final_exception (RuntimeError): Health check of container "kadai-postgres" failed terminally with status "unhealthy" after 0s. Health check errors: {"Status":"unhealthy","FailingStreak":0,"Log":[]}
```

Using the current implementation in this branch the error message is improved:

```plain
Final_exception (RuntimeError): Architecture incompatibility detected: Docker image 'registry.gitlab.com/envite-consulting/sustainable-software-architecture/kadai/kadai-databases@sha256:ad5dfc2b4075233e385e92adad9fbeb8e3f874da4e998eb5eca9c9db4556729a' is not available for host architecture 'amd64'. Image architecture is 'arm64'
```

## Should we fail GMT runs if technically the image can be run, but is only emulated?

No. Good point! I haven't thought about that and only tested the implementation with Docker native in WSL2 and not with Docker Desktop that would use the emulation of ARM.

I have implemented the necessary changes to allow emulation in the following PR:
https://github.com/green-coding-solutions/green-metrics-tool/pull/1313

Feel free to merge it into this one, if it makes sense to you. The check after the image pull was completely removed. Instead a delay of 1 second was introduced after the `docker run` to be able to check if the detached container fails immediately after start (e.g. due to an invalid architecture).
If emulation is used, the container will start without issues so there won't be any interruptions by an architecture check. In the tests I added a skip rule if they are run with Docker Desktop.

With the GMT run command from above it will now succeed if you use Docker Desktop (using emulation), but still fail if you use native Docker.

Error message if running an incompatible image architecture on native Docker:

```plain
Final_exception (RuntimeError): Container 'kadai-postgres' failed immediately after start, probably due to architecture incompatibility (exit code: 255). Image architecture is 'arm64' but host architecture is 'amd64'.
```

PR #1290, opened 2025-08-14
Fix output of container logs with the "print-logs" flag

davidkopp:
The argument `--print-logs` did not work anymore. There were multiple issues. This PR fixes them.

[Docs](https://docs.green-coding.io/docs/measuring/runner-switches/):
> `--print-logs` Prints the container and process logs to stdout

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -1893,8 +1893,10 @@ class ScenarioRunner:
             self.__start_measurement = None
             self.__start_measurement_seconds = None
             self.__notes_helper = Notes()
-
-        self.__stdout_logs.clear()
+        else:
+            # Only clear logs when continuing measurement (debugging mode)
+            # For normal completion, preserve logs for --print-logs functionality
+            self.__stdout_logs.clear()
         self.__phases.clear()
         self.__end_measurement = None
         self.__join_default_network = False
--- a/runner.py
+++ b/runner.py
@@ -237,10 +237,13 @@ if __name__ == '__main__':
         error_helpers.log_error('Base exception occured in runner.py', exception_context=e.__context__, final_exception=e, run_id=runner._run_id if runner else None)
     finally:
         if args.print_logs and runner:
-            for container_id_outer, std_out in runner.get_logs().items():
-                print(f"Container logs of '{container_id_outer}':")
-                print(std_out)
-                print('\n-----------------------------\n')
+            logs = runner._get_logs()
+            if logs:
+                print("Container logs:")
+                for log_entry in logs:
+                    print(log_entry)
+                    print('-----------------------------')
+                print()
 
         # Last thing before we exit is to shutdown the DB Pool
         DB().shutdown()
--- a/tests/data/usage_scenarios/capture_logs.yml
+++ b/tests/data/usage_scenarios/capture_logs.yml
@@ -1,0 +1,21 @@
+---
+name: Capture Logs
+author: David Kopp
+description: Capture logs from stdout and sterr
+
+services:
+  test-container:
+    type: container
+    image: alpine
+    log-stdout: true
+    log-stderr: true
+
+flow:
+  - name: Logging
+    container: test-container
+    commands:
+      - type: console
+        command: echo "Test log message" && echo "Test error message" >&2
+        shell: /bin/sh
+        log-stdout: true
+        log-stderr: true
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -497,3 +497,31 @@ def wip_test_verbose_provider_boot():
         diff = (notes[i+1][0] - notes[i][0])/1000000
         assert 9.9 <= diff <= 10.1, \
             Tests.assertion_info('10s apart', f"time difference of notes: {diff}s")
+
+## --print-logs
+def test_print_logs_flag():
+    """Test that --print-logs flag actually prints logs when they exist"""
+    runner = ScenarioRunner(uri=GMT_DIR, uri_type='folder', filename='tests/data/usage_scenarios/capture_logs.yml',
+                          skip_system_checks=True, dev_cache_build=True, dev_no_sleeps=True,
+                          dev_no_metrics=True, dev_no_phase_stats=True, dev_no_save=True)
+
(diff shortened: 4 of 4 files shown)


davidkopp:
As proposed I moved the call to `runner = ScenarioRunner` outside of the loop. 

I also started using a `prepare_run` method that clears the log. However, I finally came to a different solution, using 2 variables for the logs:
- `__current_run_logs` → gets cleared in the `cleanup` method at the end of each run
- `__all_runs_logs` → includes the cumulative logs of all runs and gets used by `--print-logs` at the end of all runs.

I also had to introduce another new variable: `_original_branch`.

Like mentioned in https://github.com/green-coding-solutions/green-metrics-tool/issues/1292 I don't like the current `cleanup` method.

Let me know what you think about the new implementation.

lib/scenario_runner.py
```diff
@@ -319,7 +318,7 @@ def _checkout_repository(self):
             if problematic_symlink := utils.find_outside_symlinks(self._repo_folder):
                 raise RuntimeError(f"Repository contained outside symlink: {problematic_symlink}\nGMT cannot handle this in URL or Cluster mode due to security concerns. Please change or remove the symlink or run GMT locally.")
         else:
-            if self._branch:
+            if self._original_branch is not None:
```
ArneTR, line 321: I see you introduced this new variable, but what is the reason here?

The code logic does not allow for the branch to be specified differently as what the user provided. The only way for the self._branch to be unset and then later be populated differently is through the route of the local directory. 
So my question: Did you run into any issues with the current implementation? If not I do not see the need for this extra var.

lib/scenario_runner.py
```diff
@@ -393,7 +392,6 @@ def recursive_lookup(k, d):
         if '/' in self._original_filename:
             self.__working_folder_rel = self._original_filename.rsplit('/', 1)[0]
             self.__working_folder = usage_scenario_file.rsplit('/', 1)[0]
-            #self.__filename = usage_scenario_file.rsplit('/', 1)[1] # we currently do not use this variable
```
ArneTR, line 396: why remove that? I find comments like this which might indicate possible future implementations quite useful ...

lib/scenario_runner.py
```diff
@@ -176,9 +178,6 @@ def __init__(self,
 
         self._check_all_durations()
 
-        # we currently do not use this variable
```
ArneTR, line 179: why remove that? I find comments like this which might indicate possible future implementations quite useful ...

lib/scenario_runner.py
```diff
@@ -1911,6 +1926,11 @@ def cleanup(self, continue_measurement=False):
 
         print(TerminalColors.OKBLUE, '-Cleanup gracefully completed', TerminalColors.ENDC)
 
+
+    def set_filename(self, filename):
```
ArneTR, line 1930: Although superflous I like the expliciteness of this

tools/kill_gmt.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/env bash
+#!/bin/bash
```
ArneTR: Why this change?

Using /usr/env is tremendously helpful if your bash is not installed in /bin and / or you have an overlayed bash (as I have on macOS which comes through brew that places it in /opt/homebrew/...)

Did you run into errors with this in this particulate PR?

ArneTR (changes requested):

lib/scenario_runner.py
ArneTR: Sorry, just read your comment about it in the conversation directly and now see the need. Is resolved.

ArneTR:
LGTM, just some minor requests. then can be merged

lib/scenario_runner.py
davidkopp: Yes, I run into an issue with the current implementation.

The code

```python
self._branch = subprocess.check_output(['git', 'branch', '--show-current'], cwd=self._repo_folder, encoding='UTF-8').strip()
```

is executed for all types of runs, both local folder and remote repo.
So in the second iteration the branch variable is set. Therefore the RuntimeError "Specified --branch but using local URI. Did you mean to specify a github url?" is raised.

Let me know if you want to have a different workaround than having the new variable "_original_branch".

lib/scenario_runner.py
davidkopp: My thinking was, that having a variable `__filename` is not relevant anymore, because runner.py overrides the filename from the outside via `runner.set_filename(filename)` before each run. So a reset of __filename to the original filename (as one commented out line in the cleanup method indicates), would be counter-intuitive. 

Should I put the 3 commented out lines back into the code?

tools/kill_gmt.sh
davidkopp: Yes, I run into an error while implementing this PR (or more specifically looking into the issue of the time travel mode 3, see https://github.com/green-coding-solutions/green-metrics-tool/issues/1292).
Error:

```sh
$ ./tools/kill_gmt.sh 
bash: ./tools/kill_gmt.sh: cannot execute: required file not found
```

`kill_gmt.sh` is the only shell script in the repo that was using the shebang `#!/usr/env bash`. All the others are using `#!/bin/bash`. Besides fixing the error I also wanted to have more consistency.

lib/scenario_runner.py
davidkopp: This method is used from the outside. So it's not completely superfluous ;)

ArneTR: By superfluous I meant that you also can access _original_filename directly. but as said: I like the expliciteness of a function here

tools/kill_gmt.sh
ArneTR: Upsi. This is however a typo. it should be `/usr/bin/env`

And then I would advocate to make this in all scripts. Not just this.

But please remove this change from this PR and make it fully separate. I feel this is separated concerns

davidkopp: I reverted the shebang change in this PR and added a new one: https://github.com/green-coding-solutions/green-metrics-tool/pull/1299

PR #1297, opened 2025-08-19
Add container dependency collection

davidkopp:
Resolves issue https://github.com/green-coding-solutions/dependency-resolver/issues/5.

## Tasks / Open Questions

Before merging the following tasks / questions should be resolved:

- [x] Publishing of the dependency_resolver to PyPI (issue https://github.com/green-coding-solutions/dependency-resolver/issues/10)
- [x] Refactor the output format of the dependency_resolver (issue https://github.com/green-coding-solutions/dependency-resolver/issues/9)
- [x] Removal of ThreadPool?

## Installation

The used [dependency-resolver](https://github.com/green-coding-solutions/dependency-resolver/) library is not published yet (here on GitHub and on PyPI)!
I installed it locally using

```sh
pip install <path-to-dependency-resolver>
```

## Visualization (Outdated)

**EDIT: The information in this section are outdated. See comments below for screenshots of the updated implementation.**

With the current implementation, only the container images are collected and displayed in a table on the tab "Usage Scenario" in the frontend:

image

No real dependencies resolving happens at the moment.

Other example:

image

Logs from runner.py:

```plain
Waiting for Metric Providers to boot ... 
Stderr check on CpuUtilizationCgroupContainerProvider
Stderr check on MemoryUsedCgroupContainerProvider
 
Collecting dependency information 
Successfully collected dependency information for 3 containers:
  - kadai-postgres: registrygitlabcomenviteconsultingsustainablesoftwarearchitecturekadaikadaidatabasespostgres147100ktasks_gmt_run_tmp:latest (sha256:fe6127e3be15a74cf3b3097c289386e65afc46d41424d2010b1a9580a302c912)
  - kadai-example-app: registrygitlabcomenviteconsultingsustainablesoftwarearchitecturekadaikadaiexamplespringbootkadai1010_gmt_run_tmp:latest (sha256:3b5e7b53918be05c3d17b36cb92f5eb088b1c3732d66d647930a32899133cc2a)
  - k6: k6:1.1.0 (sha256:529bf12eafde3435d413a0a094943abcb2ebffe1aa90e107b5e6093443ff7218)
 
Starting phase [IDLE].
```


ArneTR:
The PR integration is at the right places and I also like the integration of the debugger steps.

I am a bit skeptical about the ThreadPool which makes only sense when we have relevant I/O and not compute. Is that the case ? (Not having read the performance profiling yet, so excuse me if that is already answered somewhere).

Also the exception handling is a bit overheady. You except a lot and then later on the processing still continues even if errors are encountered. Is that beneficial? Would it not be better to fail on first exception? What is the rationale here for me to understand the implementation better.

## What I need
To really evalute this feature it would be needed that we also test an image where NPM packages (for instance) are installed in a non-pulled (but built ) docker image.
can you provide such a sample usage scenario file with needed Docker files?
Also a question here: is the functionality lacking for this currently or just a sample to test?

ArneTR:
@gemini-cli /review

ArneTR:
@gemini-cli /review

davidkopp:
Improved the visualization of the dependencies in the "Usage Scenario" tab:

image

The packages are collapsed by default.
Project and system packages are presented differently:

image

image

ArneTR:
sadly postponed to next week :/

frontend/js/stats.js
```diff
@@ -104,6 +104,172 @@ const fetchAndFillRunData = async (url_params) => {
...
+                    containerInfoDiv.className = 'ui secondary segment';
+                    containerInfoDiv.style.marginLeft = '0px';
+                    containerInfoDiv.innerHTML = `
+                        <strong>Image:</strong> ${image}<br>
+                        <strong>Hash:</strong> <code style="font-family: monospace; word-break: break-all;">${hash}</code>
+                    `;
+                    containerSection.appendChild(containerInfoDiv);
+                    
+                    // Get package managers and group by scope
+                    const packageManagers = Object.keys(containerData).filter(key =>
```
ArneTR: why do both of these very similar keys exist?

frontend/js/stats.js
```diff
@@ -104,6 +104,172 @@ const fetchAndFillRunData = async (url_params) => {
...
+                    // Container info
+                    const image = containerInfo.image || 'N/A';
+                    const hash = containerInfo.hash || 'N/A';
+                    
+                    const containerInfoDiv = document.createElement('div');
+                    containerInfoDiv.className = 'ui secondary segment';
+                    containerInfoDiv.style.marginLeft = '0px';
+                    containerInfoDiv.innerHTML = `
+                        <strong>Image:</strong> ${image}<br>
+                        <strong>Hash:</strong> <code style="font-family: monospace; word-break: break-all;">${hash}</code>
```
ArneTR: This will become a lot of CSS inline styles. Preferably use a class here which can be faster processed

frontend/js/stats.js
```diff
@@ -104,6 +104,172 @@ const fetchAndFillRunData = async (url_params) => {
...
+                            if (!scopeGroups[scope]) {
+                                scopeGroups[scope] = [];
+                            }
+                            scopeGroups[scope].push({ type: pmType, data: pmData });
+                        });
+                        
+                        // Create accordion for scopes
+                        const accordion = document.createElement('div');
+                        accordion.className = 'ui accordion';
+                        accordion.style.marginLeft = '20px';
```
ArneTR: please use an additional class here instead of added css inline styles via JS

lib/scenario_runner.py
```diff
@@ -1466,6 +1470,93 @@ def _save_warnings(self):
...
 
+    def _needs_full_dependency_resolution(self, container_name):
+        """Determine if container needs full dependency resolution based on whether it was built."""
+        services = self._usage_scenario.get('services', {})
+
+        # Find the service that corresponds to this container
+        for service_name, service in services.items():
+            # Check if this service matches the container (by name or container_name)
+            service_container_name = service.get('container_name', service_name)
+            if service_container_name == container_name or service_name in container_name:
```
ArneTR: Can you explain this check?

lib/scenario_runner.py
```diff
@@ -1466,6 +1470,93 @@ def _save_warnings(self):
...
+                # Check if service has build configuration
+                return 'build' in service
+
+        # Default to container-info-only for containers we can't match to services
+        return False
+
+    def _execute_dependency_resolver_for_container(self, container_name):
+        """Execute dependency resolver for a single container using Python package."""
+        try:
+            only_container_info = not self._needs_full_dependency_resolution(container_name)
```
ArneTR: not not? What is happening here?

lib/scenario_runner.py
```diff
@@ -1466,6 +1470,93 @@ def _save_warnings(self):
...
+
+            if not only_container_info:
+                print(f"Built container '{container_name}' requires full dependency resolution")
+                start_time = time.time()
```
ArneTR: maybe use time.perf_counter() here?

lib/scenario_runner.py
```diff
@@ -1466,6 +1470,93 @@ def _save_warnings(self):
...
+                container_identifier=container_name,
+                only_container_info=only_container_info
+            )
+
+            if not only_container_info:
+                duration = time.time() - start_time
+                print(f"Full dependency resolution for container '{container_name}' completed in {duration:.2f} s")
+
+            # Remove the 'name' field from _container-info as we use container name as key
+            if result and '_container-info' in result and 'name' in result['_container-info']:
```
ArneTR: maybe 'name' in result.get('_container-info', '') ?

metric_providers/psu/energy/ac/xgboost/machine/model
ArneTR, line 1: what is this?

tests/data/web-application/testproject/settings.py
```diff
@@ -0,0 +1,67 @@
+import os
+
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+SECRET_KEY = "test-secret-key-for-dependency-resolver-testing"
```
ArneTR, line 6: secret key? What am I missing?

metric_providers/psu/energy/ac/xgboost/machine/model
ArneTR: maybe remove this again with 
`git submodule update --remote`

ArneTR:
Another thing I noticed when taking a container that has python installed is that if no venv does exist the system packages installed are also not listed. Since many additional packages also can come through the OS this would be mandatory.

Example container would be: `greencoding/gcb_playwright:v20`

Here for instance `greenlet` is a non standard packages. And some other typical standard packages are missing, which is also an important info:

```log
arne@mintbook:~ $ docker run --rm -it greencoding/gcb_playwright:v20 bash
root@1b738e8a49d8:/# pip freeze
distlib==0.4.0
filelock==3.18.0
greenlet==3.2.3
platformdirs==4.3.8
playwright==1.54.0
pyee==13.0.0
typing_extensions==4.14.1
virtualenv==20.32.0
```

ArneTR:
I have tested with an additional container where I have installed system packages in a new virtualenv.

This is my Dockerfile:
```Dockerfile
FROM greencoding/gcb_playwright:v20

RUN cd /root && \
    python3 -m virtualenv venv  && \
    /root/venv/bin/pip install psutils

CMD ["tail", "-f", "/dev/null"]
```

`psutils` is not found

frontend/js/stats.js
```diff
@@ -104,6 +104,172 @@ const fetchAndFillRunData = async (url_params) => {
             } else {
                 document.querySelector("#usage-scenario-variables").insertAdjacentHTML('beforeend', `N/A`)
             }
+        } else if(item == 'usage_scenario_dependencies') {
```
ArneTR, line 107: I have made many small remarks also but as discussed this function is due for a full refactor in terms of speed and readability

api/api_helpers.py
```diff
@@ -137,7 +137,7 @@ def get_run_info(user, run_id):
                 (SELECT STRING_AGG(t.name, ', ' ) FROM unnest(runs.categories) as elements
                     LEFT JOIN categories as t on t.id = elements) as categories,
                 filename, start_measurement, end_measurement,
-                measurement_config, machine_specs, machine_id, usage_scenario, usage_scenario_variables,
+                measurement_config, machine_specs, machine_id, usage_scenario, usage_scenario_variables, usage_scenario_dependencies,
```
ArneTR, line 140: adding stuff here will lead to a different API signature in the frontend. Please change the mask accordingly

api/scenario_runner.py
```diff
@@ -273,7 +273,7 @@ async def get_runs(uri: str | None = None, branch: str | None = None, machine_id
     query = '''
             SELECT r.id, r.name, r.uri, r.branch, r.created_at,
             (SELECT COUNT(id) FROM warnings as w WHERE w.run_id = r.id) as invalid_run,
-            r.filename, r.usage_scenario_variables, m.description, r.commit_hash, r.end_measurement, r.failed, r.machine_id
+            r.filename, r.usage_scenario_variables, r.usage_scenario_dependencies, m.description, r.commit_hash, r.end_measurement, r.failed, r.machine_id
```
ArneTR, line 276: same here regarding the mask

ArneTR (changes requested):
as mentioned in the call also: Here are some additional files inthe PR that can be removed like the diff.txt and test.log etc.

tests/data/web-application/testproject/settings.py
davidkopp: The test project was created using the tutorial here: https://github.com/docker/awesome-compose/tree/master/official-documentation-samples/django/

The `SECRET_KEY` seems to be a requirement for Django projects. If I remove the setting it leads to the following exception:
```plain
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.
```

I will add the following comment to make that clear:

```python
# Django requires a SECRET_KEY for cryptographic signing (sessions, CSRF tokens, etc.)
SECRET_KEY = "test-secret-key-for-dependency-resolver-testing"
```

lib/scenario_runner.py
davidkopp: Since now a full dependency resolution is done in every case, this check is not relevant anymore and will be removed.

lib/scenario_runner.py
davidkopp: Since now a full dependency resolution is done in every case, this is not relevant anymore and will be removed.

frontend/js/helpers/runs.js
```diff
@@ -214,7 +214,7 @@ const getRunsTable = async (el, url, include_uri=true, include_button=true, sear
     columns.push({ data: 3, title: '<i class="icon code branch"></i>Branch', render: (el, type, row) => escapeString(el) });
 
     columns.push({
-        data: 9,
+        data: 10,
```
ArneTR, line 217: You have only changed two numbers in the file. However you need to advance all accesses to the data structure by one I think, or?

row[10] and row[11] are for instance accessed earlier in the file. Don't they have to be moved to row[11] and row[12] for instance?

if so: please also check all other array accesses >= 8 in the file

davidkopp:
The output format of the dependency_resolver has changed (see https://github.com/green-coding-solutions/dependency-resolver/issues/9#issuecomment-3309916563). Code in GMT was updated.

In the frontend the packages are separated by their scope: system or project. I'm not sure if we want to stick with this separation. In the beginning I thought it makes sense because someone could be more interested in the direct project dependencies than the system ones. However, if Python packages are installed system-wide (e.g. located in `/usr/local/lib/`) they are now treated as system packages. Does this makes sense?

Here a screenshot of the Django-based web-application scan:
image

The Python packages are listed as system packages:
image

In our last meeting you said, that the container images are missing the repository name. I fixed that. See in the screenshot above `curlimages/curl:8.10.1`. It is missing from the two others, because the GMT generated image name is used. Is that ok?

> I also changed my mind of non-built containers not running through dependency checking. I see no downside in bringing that in. Can you please enable that?

Done. The full dependency resolution is now executing for every container.

> psutils is not found

In your first test case psutils was not found, because the dependency resolver did not found the /root/venv location and only collected the system packages in /usr/local/bin. I have now fixed that. See test here: https://github.com/green-coding-solutions/dependency-resolver/tree/main/tests/specialized/pip_root_venv_detection

The second test case with the removed python3 binary still fails. The dependency-resolver depends on pip and pip does not work if python is missing:

```sh
/usr/local/bin/pip: cannot execute: required file not found
```

As we decided to use the snapshot approach instead of a filesystem scanning, there is no easy solution for this.

> Happy if you do one round of optimization and then we check in to see if that is a reasonable goal, without rabbit holing here.

I added a new optional flag `--skip-hash-collection` so I can see more easily how big the performance impact of collection the hash values are. In the mixed pip performance benchmark scenario (20 pip packages, 471 dpkg packages), it takes with hash collection around 4.5 seconds on my system, without 3.5 seconds.

I also added the new optional flag `select-detectors` to be able to explicitly define which detectors should be used. This has also a performance benefit because there is no need anymore to check if all detectors are available/usable. Using this flag would safe another ~300 ms in the mixed pip performance benchmark scenario.
dpkg detection alone can be quite fast (less than 500 ms). The pip detection however is slow in comparison (> 2 seconds).

```
$ time python3 -m dependency_resolver docker green-coding-gunicorn-container --debug --select-detectors "pip"

Connected to Docker container: green-coding-gunicorn-container
Selected detectors: pip
Checking usability of pip...
Executing docker command: sh -c 'pip --version'
Docker command completed in 0.261s with exit code: 0
pip is usable, extracting dependencies...
Executing docker command: sh -c 'echo $VIRTUAL_ENV'
Docker command completed in 0.089s with exit code: 0
Executing docker command: sh -c 'pip show pip'
Docker command completed in 0.269s with exit code: 0
Executing docker command: sh -c 'pwd'
Docker command completed in 0.092s with exit code: 0
Executing docker command: sh -c 'echo $HOME'
Docker command completed in 0.084s with exit code: 0
Executing docker command: sh -c 'find '/var/www/startup/venv' '/var/www/startup/env' '/var/www/startup/virtualenv' '/root/env' '/root/.env' '/root/.venv' '/root/virtualenv' '/var/www/startup/.env' '/root/venv' '/var/www/startup/.venv' -maxdepth 1 -name 'pyvenv.cfg' -type f 2>/dev/null | head -1'
Docker command completed in 0.089s with exit code: 0
Executing docker command: sh -c 'test -e "/var/www/startup/venv/bin/pip"'
Docker command completed in 0.084s with exit code: 0
Executing docker command: sh -c '/var/www/startup/venv/bin/pip list --format=freeze'
Docker command completed in 0.333s with exit code: 0
Executing docker command: sh -c '/var/www/startup/venv/bin/pip show pip'
Docker command completed in 0.330s with exit code: 0
Executing docker command: sh -c 'cd '/var/www/startup/venv/lib/python3.13/site-packages' && find . -name '__pycache__' -prune -o -name '__editable__*' -prune -o -name 'pip*' -prune -o -name 'setuptools*' -prune -o -name 'pkg_resources' -prune -o -name '*distutils*' -prune -o -path '*/pip/_vendor' -prune -o -not -name '*.pyc' -not -name '*.pyo' -not -name 'INSTALLER' -not -name 'RECORD' \( -type f -o -type l \) -printf '%s %p %l\n' | LC_COLLATE=C sort -n -k1,1 -k2,2'
Docker command completed in 0.139s with exit code: 0
Executing docker command: sh -c '/var/www/startup/venv/bin/pip show pip'
Docker command completed in 0.347s with exit code: 0
Found 66 project packages for pip
{"project": {"package-management": {"pip": {"location": "/var/www/startup/venv/lib/python3.13/site-packages", "hash": "8049515a820684c21d0aa3be8dcbe6570fbec163dba6ddecf24d77a4a265ba27"}}, "packages": [{"name": "annotated-types", "version": "0.7.0", "type": "pip"}, {"name": "anybadge", "version": "1.16.0", "type": "pip"}, {"name": "anyio", "version": "4.10.0", "type": "pip"}], "_excerpt_info": {"total_packages": 66, "shown": 3, "note": "Showing 3 of 66 packages (debug mode excerpt)"}}}

real    0m2.283s
user    0m0.167s
sys     0m0.038
```

I already improved the venv detection approach a little bit, but there is still room for improvement.

davidkopp:
The output format was a bad idea. I reverted it partially. Now the package managers are again the main entity...

image

image

image

ArneTR (changes requested):
I have the PR another review and it feels some functionality got broken again and / or something is unclear to me visually.

## What I did
- Pulled PR
- Reinstalled GMT
- Pulled dependency resolver
- Did "pip install" in the dependency resolver directory

Then I adapted the `basic_stress.yml` in the GMT directory itself to have the following service:
```yml
services:
  test-container:
    type: container
    image: greencoding/gcb_playwright:v20
    setup-commands:
      - shell: bash
        command: |
          cd /root
          python3 -m virtualenv venv
          /root/venv/bin/pip install psutils
          apt update
          apt install nano
```

Then I ran GMT with:
- `runner.py --uri ~/green-coding/green-metrics-tool --filename tests/data/usage_scenarios/basic_stress.yml --dev-no-sleeps --skip-system-checks`

## What is missing

- Although I can see nano getting added, I cannot see psutils
- The view for the container seems broken. I see the hash only in the _container-info ... which is a very cryptic name, above it says Hash N/A

I cleared all browser caches so either this is an error or i am missing a step.

Image

frontend/js/helpers/runs.js
ArneTR: @davidkopp This is also still unresolved. Error still persists. For instance the link to the "History" still uses wrong indexes. But is likely also in other locations

davidkopp: I finally fixed this.

davidkopp:
The energy-dependency-inspector is now able to collect the pip packages from multiple locations:

image

I also added an integration test for this case. 
In this PR there is still the Django web application included used for another integration test. Does this still have any value or should I remove this test and the web-application directory?

After the energy-dependency-inspector is published to PyPI there is only one task left to be done in this PR:
- [ ] Add `energy-dependency-inspector` to `requirements.txt`

ArneTR:
Does this also work for other package managers? 

I tried doing it dynamically for npm, but it seems like it did not find the the package manager at all.

Example service that I used:

```yml
services:
  test-container:
    type: container
    image: mcr.microsoft.com/playwright:v1.55.0-noble
    setup-commands:
      - command: cd /tmp/ && npm init -y && npm install playwright
        shell: bash
```

api/scenario_runner.py
ArneTR, line 276: @davidkopp Hey David, I wanted to check in on your rationale for adding the usage_scenario_dependencies here in the /v2/runs endpoint.

You are adding it to the return, but only the frontend when listing runs to the overview every queries this endpoint.

In the runs.js, which consumes it, you never access this field though.

What was the rationale for adding it in the first place?
Am I missing some usage of it somewhere?

davidkopp: You are totally right. That makes no sense. I don't know, why I added it to this endpoint 🙄
Here a PR that removes it: https://github.com/green-coding-solutions/green-metrics-tool/pull/1403

PR #1299, opened 2025-08-19
Make scripts more portable by change the shebang

davidkopp:
Converting the shebang `#!/bin/bash` to `#!/usr/bin/env bash` as discussed in https://github.com/green-coding-solutions/green-metrics-tool/pull/1290#discussion_r2284557897.

In addition to that I added a shebang to `run-template.sh` and changed the script to make it compatible with `sh` (which should be available everywhere). If you don't like the change and want me to revert the commit let me know.

### DIFF
--- a/.devcontainer/codespace-setup.sh
+++ b/.devcontainer/codespace-setup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euxo pipefail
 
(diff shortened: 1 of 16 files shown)


ArneTR:
I also do not know the rationale of this comment.  It was probably done by me but I would not know why I did it.

If I want to force another shell I can still do `zsh run-template.sh ...` which also works on my Mac.

Please remove the comment and move to bash

davidkopp:
I already replaced the comment. So for me the PR is ready to merge.

btw: If I click "Update branch", the complete pipeline runs again, correct? I think that's unnecessary. Should I do the merge locally and add a `[skip ci]` in the commit description? Or is there any other workaround?

PR #1300, opened 2025-08-20
Improve Time Traveling - Fix mode 2, remove mode 3 and cleanup 'cleanup' method

### LINKED ISSUE #1292: Time Travel Mode 3
During implementation of https://github.com/green-coding-solutions/green-metrics-tool/pull/1290 I discovered that the option 3 of the time travel mode does not work. The cleanup process for it has some flaws:
- `__phases.clear()` is not possible here, because not all phases are restarted, only the flows
- container metric providers are not informed about the new container ids

Is the travel mode option 3 (reload containers and restart flows) even relevant? Is it worth to fix it or can it maybe even removed?

The option 3 of the time travel mode makes the cleanup process using `continue_measurement=True` unnecessarily complex (at least it is confusing). When the time travel mode should be fixed, then I would recommend to do a refactoring of the whole time traveling implementation.

davidkopp:
Resolves #1292

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -1435,13 +1435,6 @@ class ScenarioRunner:
         else:
             self.__notes_helper.add_note( note=f"Ending phase {phase} [UNPADDED]", detail_name='[NOTES]', timestamp=phase_time)
 
-    def _clear_flow_phases_for_timetravel(self, up_to_flow_id):
-        """Clear only flow phases from the phases dictionary, preserving infrastructure phases."""
-        for i in range(up_to_flow_id + 1):
-            if i < len(self._usage_scenario['flow']):
-                flow_name = self._usage_scenario['flow'][i]['name']
-                self.__phases.pop(flow_name, None)
-
     def _run_flows(self):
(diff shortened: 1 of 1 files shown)


ArneTR:
Implementation LGTM but I think I discovered a bug in the code.

IMHO this can be left in and just needs some documentation. Interested in your take on it. here it goes:

Line `self.__ps_to_read.clear() # clear, so we do not read old processes` will clear all processes from the list and thus making GMT unable to read their STDOUT and possible SCI later.

This clearing should actually only happen if a step is retried (or the whole flow) and then only for the processes that happend in these steps / whole flow. This bc also processes from the BOOT phase can be in this list.

The result is that the flows can be time-traveled but all process data gets lost in the proces.

SInce it is a debugging feature I think this can be implemented this way. However unexpected might be that it will be impossible the get SCI scores when using this feature as they are read from STDOUT.

How do you feel about this?

I think an implementation to keep process data will need a more granular data structure where we keep the processes in that enables to identify in which step they have been executed

ArneTR:
LGTM

PR #1307, opened 2025-08-29
Moving away from hostnamectl as we see dbus timeouts in restricted envs

ArneTR:
I am currently setting up the GMT in a restricted environment:

- https://github.com/nestybox/sysbox/tree/master

The sysbox container allows to have a full systemd inside a docker container and thus allow GMT testing with docker in docker.

This enviroment thus does not have a working `dbus` messaging system. Thus the calls to `/usr/bin/hostnamectl` result in timeouts.

Example:
```log
(venv) root@59fe5ff2b021:~/green-metrics-tool# /usr/bin/hostnamectl deployment
Could not get property: Connection timed out
```

If I look at the `dbus` process it looks like a dummy:
```log
(venv) root@59fe5ff2b021:~/green-metrics-tool# ps -ax | grep dbus
    948 ?        Ss     0:00 @dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
  12462 pts/2    S+     0:00 grep --color=auto dbus
```

Thus this PR wants to go away from `/usr/bin/hostnamectl` and use alternative means.

@davidkopp 
@ribalba 

Please test this in your WSL / Fedora systems if results are ok and you see the values also properly displayed in the frontend.


davidkopp:
I asked Claude about the failed detection of the operating system:

# Regex Analysis for /etc/os-release VERSION Pattern

## Problem

The regex pattern on line 28 of `lib/hardware_info.py` has issues:

```python
r'VERSION=["\']\s*(?P<o>.*)["\']'
```

## Issues Identified

1. **Character class issue**: `["\']\s*` means "match a quote OR apostrophe OR whitespace OR asterisk" - the `\s*` is treated as literal characters inside the character class, not as a quantifier.

2. **Greedy matching**: The `.*` will match everything including quotes, so it captures too much.

3. **Inconsistent quote matching**: It doesn't ensure the opening and closing quotes match.

## Current /etc/os-release Format

```
VERSION="24.04.3 LTS (Noble Numbat)"
```

## Solution

The correct regex should be:

```python
r'VERSION=(["\'])(?P<o>.*?)\1'
```

This pattern:
- `(["\'])` captures the opening quote (single or double) in group 1
- `(?P<o>.*?)` captures the content non-greedily 
- `\1` matches the same quote character as captured in group 1

For the current `/etc/os-release` format, this would correctly capture `24.04.3 LTS (Noble Numbat)`.

davidkopp:
Output of `hostnamectl` on my WSL2 system:

```plain
 Static hostname: thinkpad-t14s
       Icon name: computer-container
         Chassis: container ☐
      Machine ID: 9b16d2e...
         Boot ID: 03d8ed6...
  Virtualization: wsl
Operating System: Ubuntu 24.04.3 LTS
          Kernel: Linux 6.6.87.2-microsoft-standard-WSL2
    Architecture: x86-64
```

ArneTR:
@davidkopp @ribalba 

Can you please test again. Incoporated some updates

lib/hardware_info.py
```diff
@@ -73,8 +74,9 @@
     [rpwr, 'Cpu Info', 'sysctl -n machdep.cpu.brand_string', r'(?P<o>.*)'],
     [rpwr, 'Memory Total', 'sysctl -n hw.memsize', r'(?P<o>.*)'],
     [rpwr, 'Mac Version', 'sw_vers -productVersion', r'(?P<o>.*)'],
+    [rpwr, 'Hostname', 'uname -n', r'(?P<o>.*)'],
     [rpwr, 'Build Version', 'sw_vers -buildVersion', r'(?P<o>.*)'],
-    [rpwr, 'Uname', 'uname -a', r'(?P<o>.*)'],
+    [rpwr, 'Kerne Version', 'uname -srv', r'(?P<o>.*)'],
```
davidkopp: I think that line should be removed. Kernel Version is already included in line 27.

lib/hardware_info.py
```diff
@@ -24,16 +24,17 @@
     [platform.system, 'Platform'],
     [rfwr, 'Cpu Info', '/proc/cpuinfo', r'model name.*:\s(?P<o>.*)'],
     [rfwr, 'Memory Total', '/proc/meminfo', r'MemTotal:\s*(?P<o>.*)'],
-    [rpwr, 'Linux Version', '/usr/bin/hostnamectl', r'Kernel:\s*(?P<o>.*)'],
-    [rpwr, 'Operating System', '/usr/bin/hostnamectl', r'Operating System:\s*(?P<o>.*)'],
-    [rpwr, 'Architecture', '/usr/bin/hostnamectl', r'Architecture:\s*(?P<o>.*)'],
+    [rpwr, 'Kernel Version', 'uname -srv', r'(?P<o>.*)'],
```
davidkopp, line 27: "Linux version" was replaced with "Kernel Version":

image

image

works.

lib/hardware_info.py
```diff
@@ -24,16 +24,17 @@
...
+    [rfwr, 'Operating System', '/etc/os-release', r'VERSION=(["\'])(?P<o>.*)\1'],
```
davidkopp: image

"Ubuntu" is not part of the output anymore.
If this should be part, you could consider combining NAME and VERSION.

```sh
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.3 LTS (Noble Numbat)"
VERSION_CODENAME=noble
```

lib/hardware_info.py
```diff
@@ -73,8 +74,9 @@
     [rpwr, 'Cpu Info', 'sysctl -n machdep.cpu.brand_string', r'(?P<o>.*)'],
     [rpwr, 'Memory Total', 'sysctl -n hw.memsize', r'(?P<o>.*)'],
     [rpwr, 'Mac Version', 'sw_vers -productVersion', r'(?P<o>.*)'],
+    [rpwr, 'Hostname', 'uname -n', r'(?P<o>.*)'],
```
davidkopp, line 77: Works

image

lib/hardware_info.py
ArneTR: That is for a different architecture. We separate macOS and linux

lib/hardware_info.py
ArneTR: yes. intentionally. but thanks for the screenshot, that helps me verify

lib/hardware_info.py
ArneTR: thanks for that. according to the standard only name and version_id are to be guaranteed. Will move to that

lib/hardware_info.py
```diff
@@ -24,16 +24,17 @@
     [platform.system, 'Platform'],
     [rfwr, 'Cpu Info', '/proc/cpuinfo', r'model name.*:\s(?P<o>.*)'],
     [rfwr, 'Memory Total', '/proc/meminfo', r'MemTotal:\s*(?P<o>.*)'],
-    [rpwr, 'Linux Version', '/usr/bin/hostnamectl', r'Kernel:\s*(?P<o>.*)'],
-    [rpwr, 'Operating System', '/usr/bin/hostnamectl', r'Operating System:\s*(?P<o>.*)'],
-    [rpwr, 'Architecture', '/usr/bin/hostnamectl', r'Architecture:\s*(?P<o>.*)'],
+    [rpwr, 'Kernel Version', 'uname -srv', r'(?P<o>.*)'],
+    [rfwr, 'Operating System', '/etc/os-release', r'^(?:NAME|VERSION_ID)=(["\'])(?P<o>.+)\1$[\n.]*^(?:NAME|VERSION_ID)=(["\'])(?P<o2>.*)\3$', re.MULTILINE, ('o', 'o2')],
```
davidkopp, line 28: The new result on WSL2:

Image

ArneTR: although not perfect I think this is ok. it trades better generality for slightly less exact infos. Minor versions anyway only differentiate between installed packages which we track

ribalba (approved):

PR #1311, opened 2025-09-02
Saves the existing config files on install

ribalba:
Something that annoyed me all the time is that when I do `install_linux.sh` the install script just overwrites my config files. Especially if I have set up something not trivial and then I switch branches and install to get the right libs everything is gone. Now I just save the config files to a `.backup` so I can copy paste the things I need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 __pycache__
 /config.yml
+/config.yml.backup
 /docker/nginx/api.conf
 /docker/nginx/frontend.conf
 /docker/nginx/block.conf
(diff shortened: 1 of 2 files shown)


ArneTR (changes requested):
I do that always manually, but do agree that this is insider knowledge and nice to have as QoL.

To my understandig these files also need to be in .gitignore:
-     docker/nginx/block.conf
-    docker/nginx/frontend.conf

Can you confirm and if so add. Then please merge from your side

ribalba:
After thinking about this I decided to add `*.backup` as a general gitignore. This could be debated. I think this is cleaner and quite clear. But could also go for the option of naming each file.

ArneTR:
Please explicitely name the files. Makes it also clearer where they are created.

Catch-Alls should at least be directory restricted IMHO

ArneTR:
@ribalba ping

ribalba:
done

PR #1313, opened 2025-09-03
Improve container failure detection

davidkopp:
The reason for this PR is the architecture check implemented in https://github.com/green-coding-solutions/green-metrics-tool/pull/1289.
With this PR the architecture check does not lead to an immediately exit if emulation is used. For example, using Docker Desktop, an ARM-based image can be executed on a AMD-based host system.

EDIT: In general, this PR improves the container failure detection, through:
- architecture compatibility check is done before `docker run` is executed (all the logic for this is part of a new library file called `lib/container_compatibility.py`)
- after the boot phase and after the runtime phase there is now a verification if all the containers are still running (similar to a heartbeat)

### DIFF
--- a/lib/container_compatibility.py
+++ b/lib/container_compatibility.py
@@ -164,9 +164,9 @@ def _get_supported_platforms():
 
     # Final fallback to native platform
     native_platform = get_native_platform()
-    result = [native_platform]
-    _platform_cache[cache_key] = result
-    return result
+    platforms = [native_platform]
+    _platform_cache[cache_key] = platforms
+    return platforms
 
 
 def _can_run_natively(target_platform, host_platform):
@@ -234,7 +234,10 @@ def get_platform_compatibility_status(platform=None):
         
     Returns:
         If platform specified: CompatibilityStatus (NATIVE, EMULATED, or INCOMPATIBLE)
-        If platform is None: Dict with detailed platform compatibility information
+        If platform is None: Dict with detailed platform compatibility information containing:
+            - platform_compatibility: Dict mapping platforms to CompatibilityStatus
+            - native_platform: String of the host's native platform
+            - emulated_platforms: List of platforms that require emulation
     """
     cache_key = 'emulation_support_detailed'
     if cache_key in _platform_cache:
@@ -297,6 +300,8 @@ def check_image_architecture_compatibility(image_name):
             - status: CompatibilityStatus enum (NATIVE, EMULATED, or INCOMPATIBLE)
             - can_run: True if image can execute (natively or via emulation)
             - needs_emulation: True if emulation is required to run
+            - is_native_compatible: True if image runs natively without emulation
+            - error: Error message if compatibility check failed, None otherwise
     """
     try:
         # Get image architecture
@@ -351,6 +356,6 @@ if __name__ == '__main__':
         print(f'  {platform_name}: {status.value}')
 
     if compatibility_status['emulated_platforms']:
-        print('Emulated platforms available:', ', '.join(compatibility_status['emulated_platforms']))
+        print('Emulation supported:', ', '.join(compatibility_status['emulated_platforms']))
     else:
-        print('No emulated platforms available')
+        print('No emulation supported')


lib/scenario_runner.py
```diff
@@ -1304,6 +1314,39 @@ def _setup_services(self):
                 'read-sci-stdout': service.get('read-sci-stdout', False),
             }
 
+            # Check if detached container failed immediately after startup
+            # Docker run -d returns exit code 0 (success) even when containers fail moments later
+            # Common causes: architecture mismatch, missing dependencies, invalid entrypoints
+            time.sleep(1)
+            check_ps = subprocess.run(
+                    ['docker', 'ps', '-q', '-f', f'name={container_name}'],
+                    capture_output=True,
```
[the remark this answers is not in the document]
davidkopp: If the service has a health check implementation, a similar pattern is used to check the health of the dependent services at a later stage of the boot phase.
Therefore I think, your suggestion is not relevant here.

lib/scenario_runner.py
```diff
@@ -257,22 +257,35 @@ def _join_paths(self, path, path2, force_path_as_root=False):
 
         raise FileNotFoundError(f"{path2} in {path} not found")
 
-    def _validate_image_architecture(self, image_name: str) -> None:
-        """Validate that the pulled Docker image architecture is compatible with the host."""
+    def _validate_image_architecture(self, image_name: str, raise_on_mismatch: bool = True) -> tuple[str, str, bool]:
```
ArneTR: Why is there a raise_on_mismatch if it is not used?

lib/scenario_runner.py
```diff
@@ -257,22 +257,35 @@ def _join_paths(self, path, path2, force_path_as_root=False):
...
+
+        Returns:
+            Tuple of (image_arch, host_arch, is_compatible)
+        """
         ps = subprocess.run(
             ['docker', 'image', 'inspect', image_name, '--format', '{{.Architecture}}'],
             stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='UTF-8', check=False
         )
 
         if ps.returncode != 0:
```
ArneTR: Should this not except in any case?

lib/scenario_runner.py
```diff
@@ -257,22 +257,35 @@ def _join_paths(self, path, path2, force_path_as_root=False):
...
+            if raise_on_mismatch:
+                raise RuntimeError(f"Failed to inspect Docker image architecture for '{image_name}': {ps.stderr}")
+            return "", "", False
 
         image_arch = ps.stdout.strip()
         normalized_host_arch = map_host_to_docker_arch()
+        is_compatible = image_arch == normalized_host_arch
 
-        if image_arch != normalized_host_arch:
+        if not is_compatible and raise_on_mismatch:
```
ArneTR: I see that here should be no raise, as the check is done later. But should this line not fully removed if we never set raise_on_mismatch = true?

lib/scenario_runner.py
```diff
@@ -1304,6 +1314,48 @@ def _setup_services(self):
                 'read-sci-stdout': service.get('read-sci-stdout', False),
             }
 
+            # Check if detached container failed immediately after startup
+            # Docker run -d returns exit code 0 (success) even when containers fail moments later
+            # Common causes: architecture mismatch, missing dependencies, invalid entrypoints
+            time.sleep(1)
+            check_ps = subprocess.run(
```
ArneTR: Please do not use text= but rather encoding=  to be consistent

lib/scenario_runner.py
```diff
@@ -1304,6 +1314,48 @@ def _setup_services(self):
...
+                inspect_ps = subprocess.run(
+                    ['docker', 'inspect', '--format={{.State.ExitCode}}', container_name],
+                    capture_output=True,
+                    text=True,
+                    check=False
+                )
+                exit_code = inspect_ps.stdout.strip() if inspect_ps.returncode == 0 else "unknown"
+
+                if exit_code == "0":
+                    # Container exited successfully but immediately - unexpected behavior, log but continue
```
ArneTR: Since the container has been added to the __containers  list just logging will not suffice. GMT will fail later as it cannot attach reporters. I would except here.

lib/scenario_runner.py
```diff
@@ -1304,6 +1314,48 @@ def _setup_services(self):
...
+                                          container_logs_stderr=logs_ps.stderr)
+                else:
+                    # Container failed with non-zero or unknown exit code - this is a real failure
+                    image_arch, host_arch, is_compatible = self._validate_image_architecture(service['image'], raise_on_mismatch=False)
+                    if image_arch and host_arch and not is_compatible:
+                        raise RuntimeError(f"Container '{container_name}' failed immediately after start, probably due to architecture incompatibility (exit code: {exit_code}). Image architecture is '{image_arch}' but host architecture is '{host_arch}'.\nContainer logs:\n{logs_ps.stdout}\n{logs_ps.stderr}")
+                    else:
+                        raise RuntimeError(f"Container '{container_name}' failed immediately after start (exit code: {exit_code}). This indicates startup issues such as missing dependencies, invalid entrypoints, or configuration problems.\nContainer logs:\n{logs_ps.stdout}\n{logs_ps.stderr}")
+
             print('Stdout:', container_id)
```
ArneTR: If this state is reached via emulation I would like to see a warning still. Please just add the arch incompatibility to the '_warnings' data which will later be saved via GMT

ArneTR (changes requested):
Looks quite good. I could reproduce the cases you put in the base PR with Kadai. I think this solution is clean.

Made some remarks.

lib/scenario_runner.py
davidkopp: Leftover from a previous iteration. Removed it now.

lib/scenario_runner.py
davidkopp: yes. Changed it now.

lib/scenario_runner.py
davidkopp: Changed it so it will raise an exception. However, this has one disadvantage: Test cases with containers that exit intermediately are not possible anymore. See test case `test_cmd_entrypoint`. I did not want to extend the duration of this test case. Therefore I introduced a new argument of the ScenarioRunner: `enable_detached_container_start_check`

Not sure, if this is a good solution. Alternative would be to extend the duration of the test case.

lib/scenario_runner.py
davidkopp: Done. Note: Only an arch incompatibility check is done, but no check if emulation is used (e.g. by checking if Docker Desktop is used).

lib/scenario_runner.py
ArneTR: I somehow have a hard time to understand the implementation here. In GMT it should never be possible that containers exit therefore introducing this variable does seem counter-intutive for me. They should always stay alive (unless intentionally disconnected).

Maybe this is better discussed in person on Monday as it can maybe resolved through a design or test-design change.

davidkopp: I've changed my mind. I removed the variable again and instead fixed the failing test `test_depends_on_error_not_running` and disabled the second failing test `test_cmd_entrypoint`. Because the test only discusses how `entrypoint` and `command` can work together, I think it is ok that this test is skipped. If the test should be enabled again, a fix could be to add sleeps of >1s to all commands.

ArneTR:
A good 'ol 10 pm PR on a weekday :). But, they tend to be the best! 
Happy to discuss it tomorrow in the call.

ArneTR:
An adjacent question here: Currently the GMT expects container to be always running.

If containers do not behave like this they must be "hacked" to stay open. Either by putting a shel in the ENTRYPOINT or a `tail -f /dev/null`

Do you have seen that already documented in the docs? Or something similar? Hope you had looked into it before I search. If not: please give it a search

PR #1315, opened 2025-09-05
Implement database reconnection with automatic retry logic

### LINKED ISSUE #1296: GMT reconnect error when running in cluster mode
When running in cluster mode and GMT is effectively running inside an infinite loop in [([client.py](https://github.com/green-coding-solutions/green-metrics-tool/blob/main/cron/client.py))](https://github.com/green-coding-solutions/green-metrics-tool/blob/main/cron/client.py) sometimes the connection to the DB drops or gets reset.

This is the error experienced:

```log

File "/home/gc/green-metrics-tool/cron/client.py", line 195, in
job = Job.get_job('run')
File "/home/gc/green-metrics-tool/lib/job/base.py", line 99, in get_job
cls.clear_old_jobs()
File "/home/gc/green-metrics-tool/lib/job/base.py", line 163, in clear_old_jobs
DB().query(query)
File "/home/gc/green-metrics-tool/lib/db.py", line 76, in query
return self.__query(query, params=params, return_type=None, fetch_mode=fetch_mode)
File "/home/gc/green-metrics-tool/lib/db.py", line 64, in __query
cur.execute(query, params)
File "/home/gc/green-metrics-tool/venv/lib/python3.10/site-packages/psycopg/cursor.py", line 97, in execute
raise ex.with_traceback(None)
psycopg.OperationalError: consuming input failed: server closed the connection unexpectedly
This probably means the server terminated abnormally
before or while processing the request.
server closed the connection unexpectedly
This probably means the server terminated abnormally
before or while processing the request.


```

GMT should not fail but retry the connection. Currently no such means is implemented, but should be possible through the used postgresql adapter of psycopg3 that we use in https://github.com/green-coding-solutions/green-metrics-tool/blob/main/lib/db.py

davidkopp:
Resolves #1296

### DIFF
--- a/lib/db.py
+++ b/lib/db.py
@@ -1,12 +1,73 @@
 #pylint: disable=consider-using-enumerate
 import os
+import time
+import random
+from functools import wraps
 from psycopg_pool import ConnectionPool
 import psycopg.rows
+import psycopg
 import pytest
 from lib.global_config import GlobalConfig
 
 def is_pytest_session():
     return "pytest" in os.environ.get('_', '')
+
+def with_db_retry(func):
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        config = GlobalConfig().config
+        retry_timeout = config.get('postgresql', {}).get('retry_timeout', 300)
+        retry_interval = config.get('postgresql', {}).get('retry_interval', 1)
+
+        start_time = time.time()
+        attempt = 0
+
+        while time.time() - start_time < retry_timeout:
+            attempt += 1
+            try:
+                return func(self, *args, **kwargs)
+            except (psycopg.OperationalError, psycopg.DatabaseError) as e:
+                # Check if this is a connection-related error that we should retry
+                error_str = str(e).lower()
+                retryable_errors = [
+                    'connection', 'closed', 'terminated', 'timeout', 'network',
+                    'server', 'unavailable', 'refused', 'reset', 'broken pipe'
+                ]
+
+                is_retryable = any(keyword in error_str for keyword in retryable_errors)
+
+                if not is_retryable:
+                    # Non-retryable error (e.g., SQL syntax error)
+                    print(f"Database error (non-retryable): {e}")
+                    raise
+
+                time_elapsed = time.time() - start_time
+                if time_elapsed >= retry_timeout:
+                    print(f"Database retry timeout after {attempt} attempts over {time_elapsed:.1f} seconds. Last error: {e}")
+                    raise
+
+                # Exponential backoff with jitter
+                backoff_time = min(retry_interval * (2 ** (attempt - 1)), 30)  # Cap at 30 seconds
+                jitter = random.uniform(0.1, 0.5) * backoff_time
+                sleep_time = backoff_time + jitter
+
+                print(f"Database connection error (attempt {attempt}): {e}. Retrying in {sleep_time:.2f} seconds...")
+
+                # Try to recreate the connection pool if it's corrupted
+                try:
+                    if hasattr(self, '_pool'):
+                        self._pool.close()
+                        del self._pool
+                    self._create_pool()
+                except (psycopg.OperationalError, psycopg.DatabaseError, AttributeError) as pool_error:
+                    print(f"Failed to recreate connection pool: {pool_error}")
+
+                time.sleep(sleep_time)
+
+        # If we get here, we've exhausted all retries
+        raise psycopg.OperationalError(f"Database connection failed after {attempt} attempts over {time.time() - start_time:.1f} seconds")
+
+    return wrapper
 
 class DB:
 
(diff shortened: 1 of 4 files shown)


tests/test_runner.py
```diff
@@ -598,3 +600,105 @@ def test_print_logs_flag_with_iterations():
...
+    T+13.1s - Database restart completes
+    T+16.1s - Database should be available again
+    T+20.0s - REMOVE phase starts (database operations resume with retry logic)
+    T+22.4s - Test completes successfully with database reconnection
+    
+    This timing ensures database restart happens during active measurement phase
+    when database operations are likely occurring for metric storage.
+    """
+
+    def log_with_timestamp(message, prefix="TEST", start_time=None):
```
ArneTR: if this is a helper function and very general I think placing it in test_functions would be more expected

tests/test_runner.py
```diff
@@ -598,3 +600,105 @@ def test_print_logs_flag_with_iterations():
...
+        time.sleep(15)  # Wait for runtime to start but not complete
+        log_with_timestamp("Restarting test-green-coding-postgres-container now...", "DB RESTART", test_start_time)
+        result = subprocess.run(['docker', 'restart', 'test-green-coding-postgres-container'],
+                               check=True, capture_output=True)
+        log_with_timestamp(f"Database restart completed. Docker output: {result.stdout.decode().strip()}", "DB RESTART", test_start_time)
+        time.sleep(3)  # Give DB time to restart
+        log_with_timestamp("Database should be available again after 3s wait", "DB RESTART", test_start_time)
+
+    # Start database restart in background thread
+    restart_thread = threading.Thread(target=restart_database)
```
ArneTR: I wonder if this is the nicest implementation. If the sole reason is that docker restart takes long to return I think a docker compose stop -t 0 and then a docker a docker start should do it. Since there should be no I/O also no thread is needed.

Would that help?

My issue with the thread: Everytime I see the Thread library used my brain switches to scanner mode to find the high complexity part of the code. So it makes code for me harder to read if there is no strong reason for it to use. Maybe also a me issue ...

lib/db.py
```diff
@@ -1,13 +1,74 @@
...
+import psycopg
 import pytest
 from lib.global_config import GlobalConfig
 
 def is_pytest_session():
     return "pytest" in os.environ.get('_', '')
 
+def with_db_retry(func):
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
```
ArneTR, line 17: The function looks good but I think it has a design issue.

If the connection is not on auto-commit you might be re-connection and retrying a part of the query only. This can be tricky problem to debug.

Example code :
```
conn.autocommit = False
query('INSERT 1 INTO table')
## Connection drops here
## connection gets reinstantiated
query('DELETE 1 FROM table') # only the delete query is now executed as first request was dropped
```

In order to find a solution here I believe either the queries need to be buffered ... or we could decide to still fail ... this however would defeat the purpose of this PR entierely i feel.

Currently unclear what the best implementation is without a clear view how complex a query buffering would be. Please investigate / make a concept.

2. Als the `copy()` statement is currently not profiting from the statement. This would need to be covered too.
To extend this question: Are there any other query methods that can be covered we should think of to future-proof?

3. How does the DB behave on keep-alive queries? Will they also lead to a reconnect?
To my understanding the postgres adapter does regular keep-alives. If they fail it will also complain. Can we hook into this?

ArneTR (changes requested):

tests/test_runner.py
davidkopp: The threading isn't about I/O performance here. It's purpose is the concurrent restart execution and timing coordination:
- GMT runner must keep running while database restarts
- Database restart should happen during runtime phase (T+15s)
- Concurrency is required (I don't know how to do this sequentially)

lib/db.py
davidkopp: ## 1. Multi-Statement Queries

The issue you describe is only relevant, if GMT uses multi-statement queries, correct?
I did not find any usage of the multi-statement query feature. Did I overlook something?
If multi-statement queries are really not relevant for GMT, the easiest solution would be to remove the following code block:

```python
if isinstance(query, list) and isinstance(params, list) and len(query) == len(params):
    for i in range(len(query)):
        # In error case the context manager will ROLLBACK the whole transaction
        cur.execute(query[i], params[i])
```

And to consider enabling the `autocommit` feature.

If the multi-statement query feature is needed, I would do a more comprehensive analysis on how to become transactional.

## 2. Cover copy method

Done.

## 3. PostgreSQL keep-alive queries?

There are no automatic keep-alive queries:

> The pool doesn’t actively check the state of the connections held in its state. This means that, if communication with the server is lost, or if a connection is closed for other reasons [...], the application might be served a connection in broken state.
> — https://www.psycopg.org/psycopg3/docs/advanced/pool.html

tests/test_runner.py
davidkopp: Done.

davidkopp:
Thanks for the review. Changes made:
- added test cases for the db methods
- added the `with_db_retry` decorator to the `copy_from` method
- moved `log_with_timestamp` to `test_functions.py`

Compare:
https://github.com/green-coding-solutions/green-metrics-tool/compare/8bb0b4b...16657ec

tests/test_runner.py
davidkopp: Threading is now removed. I implemented it as discussed: Enhanced `run_until` using `yield`. For downward compatibility I added a new method `run_steps` with the yield logic, `run_until` uses it internally but can be used in the same way as before.
The test is now a lot simpler and faster. Great suggestion!

lib/db.py
davidkopp: ## Amendment to the 3. point "PostgreSQL keep-alive queries?"

psycopg3 includes a built-in `ConnectionPool` retry mechanisms. It is disabled by default, but can be enabled with the parameter `check`:

```python
self._pool = ConnectionPool(
    # ...
    check=ConnectionPool.check_connection
)
```

### Comparison

How does it compare to the custom `@with_db_retry` decorator implementation?

#### psycopg3 Built-in Limitations

**Measurement Interference:**
- `check=ConnectionPool.check_connection` executes `conn.execute("")` on every connection borrow
- Continuous background pool maintenance network calls
- Unnecessary network calls during measurement runs should be avoided

**Limited Error Scope:**
- No checks performed during actual query execution (`cur.execute()`)
- Cannot handle connections that fail mid-query
- Generic `PoolTimeout` exceptions with limited context

#### Custom `@with_db_retry` Advantages

**Measurement Integrity:**
- Zero network overhead during normal operation
- Only activates during actual database failures
- No background connection health checks

**Superior Error Handling:**
- Detects and retries failures during `cur.execute()`
- Classifies retryable vs non-retryable errors
- Complete pool recreation ensures clean state
- Detailed retry logging with error context

**Similar Retry Performance:**
Both mechanisms use comparable exponential backoff (1s, 2s, 4s, 8s...) with jitter.

### Conclusion

**Use the custom `@with_db_retry` mechanism** because:
- Preserves measurement accuracy by avoiding unnecessary network calls
- Better error detection scope (covers mid-query failures)
- Superior observability and debugging capabilities
- No compromise in retry performance

To be explicit that we don't want to use the check functionality of psycopg3, we should set it to `None`:

```python
self._pool = ConnectionPool(
    # ...
    check=None
)
```

PR #1319, opened 2025-09-07
Explicit container registry

ArneTR:
Container Registry is a new variable possible mandatory in the configuration.

This enables GMT to force-use alternative container registires for it's use of the Docker CLI, but also for the Kaniko building agent, which needs it to be supplied explicitely


ArneTR:
@gemini-cli /review

ArneTR:
@gemini-cli /review

PR #1329, opened 2025-09-11
Add tests and fixes for PR "Remove html escape"

davidkopp:
This PR improves the PR #1318 by adding XSS detection tests and some fixes.
The pages `runs.html`, `stats.html`, `compare.html`, `watchlist.html`, `timeline.html`, `ci-index.html` and `ci.html` are covered in the tests (see `test_frontend.py`).
EDIT: ~~I did not add a test for `timeline.html`, as it seems to be not used anymore.~~

With the tests I found some more vulnerabilities. 
I also had to remove `escapeString()` from some places where it was not appropriate (using it in URLs or escaping twice).

On the `compare.html` I changed the display of usage scenario variables. Reason: The escaped Python dict string looked ugly, therefore I converted it into a table without ugly characters.

The refactoring of the logs data structure will be done in a separate PR.

### DIFF
--- a/tests/frontend/test_frontend.py
+++ b/tests/frontend/test_frontend.py
@@ -895,7 +895,7 @@ class TestXssSecurity:
         # Test 6: Compare page (usage scenario variables)
         compare_url = f"{base_url}/compare.html?ids={run_id},{run_id2}&force_mode=usage_scenario_variables"
 
-        # Temporarily remove the page error handler for this compare test since there's a JS error I don't know how to fix
+        # Temporarily remove the page error handler for this compare test since there's a JS error due to missing phase_stats data
         page.remove_listener("pageerror", handle_page_error)
 
         page.goto(compare_url)


tests/frontend/test_frontend.py
```diff
@@ -21,33 +25,34 @@
...
+    Set up browser for each test
+    We must close the browser to clear localStorage
+    """
     global page #pylint: disable=global-statement
     global context #pylint: disable=global-statement
     global browser #pylint: disable=global-statement
 
-    browser = playwright.firefox.launch()
+    browser = playwright.firefox.launch(
+        headless=True,
```
ArneTR: did you run into an issue here? headless should be default ...

tests/frontend/test_frontend.py
```diff
@@ -61,95 +66,107 @@ def setup_and_cleanup_test():
...
-
-    value = page.locator("#scenario-runner-count").text_content()
-    assert value== '5'
-
-    page.goto(GlobalConfig().config['cluster']['metrics_url'] + '/index.html')
-    value = page.locator("div.ui.cards.link > div.ui.card:nth-child(2) a.header").text_content()
-
-    assert value== 'Eco CI'
+## Fixture for tests that need demo data
+@pytest.fixture()
```
ArneTR, line 70: nice!

tests/frontend/test_frontend.py
```diff
@@ -191,525 +207,823 @@ def test_eco_ci_adding_data():
...
+            malicious_uri,
+            malicious_branch,
+            malicious_filename,
+            json.dumps(malicious_variables),
+            1,
+            'daily',
+            1
+        ))
+
+        page.evaluate("window.IMG_XSS_EXECUTED = false")
```
ArneTR, line 850: nice, way cleaner than using alert

tests/frontend/test_frontend.py
davidkopp: I only added this to be able to quickly change the setting. For debugging it was helpful for me to see the UI und click around in the browser (using a debugging breakpoint in the test).

frontend/js/timeline.js
```diff
@@ -58,15 +59,19 @@ const populateMachines = async () => {
...
+    repository_uri = url_params['uri']; // Store the unescaped value globally for URL construction later
+    if(repository_uri == null
+        || repository_uri == ''
+        || repository_uri == 'null') {
         showNotification('No uri', 'uri parameter in URL is empty or not present. Did you follow a correct URL?');
         throw "Error";
     }
-    $('input[name="uri"]').val(escapeString(url_params['uri']));
-    $('#uri').text(escapeString(url_params['uri']));
+    if(!repository_uri.startsWith('http') && !repository_uri.startsWith('/')) {
```
davidkopp, line 69: I added  `&& !repository_uri.startsWith('/')` only because of the test case `tests/frontend/test_frontend.py::TestFrontendFunctionality::test_watchlist`. If understand it correctly, the timeline feature can be only used in production with remote URLs, but not with local file paths. However, because in the test database there are only runs with local file path, the test would fail if I only check here for `repository_uri.startsWith('http')`.
Nevertheless, I think it makes sense here to have a `startsWith` check to prevent certain XSS attacks (though, I have not checked if it's realistic here).

tests/frontend/test_frontend.py
```diff
@@ -191,525 +207,832 @@ def test_eco_ci_adding_data():
...
+        page.wait_for_load_state("networkidle")
+        page.wait_for_function("() => document.body.innerText.includes('deadbeef123456789abcdef')", timeout=10000)
+
+        compare_xss_executed = page.evaluate("window.IMG_XSS_EXECUTED")
+        assert compare_xss_executed is not True, "XSS vulnerability detected on compare page: malicious code executed"
+
+        # Test 6: Compare page (usage scenario variables)
+        compare_url = f"{base_url}/compare.html?ids={run_id},{run_id2}&force_mode=usage_scenario_variables"
+
+        # Temporarily remove the page error handler for this compare test since there's a JS error I don't know how to fix
```
davidkopp: fyi: Here the error was `Error('o[t] is undefined`. Probably my custom phase stats data are incomplete.

frontend/js/timeline.js
ArneTR: looks good

tests/frontend/test_frontend.py
ArneTR: When I run locally with test-data imported I also get an error.

Screenshot attached: 

Image

This looks to me as if you need to merge in main. A fix I pushed a while ago seems missing on the branch.
Can you try merging in main? I quickly tried but getting a merge conflict.

davidkopp: I merge main into the branch and fixed the merge conflict. 
The error I mentioned is however related to the test, probably due to missing phase_stats data (I only insert a few and not use your demo data). However, I think for this test regarding the XSS detection this is not an issue.

PR #1334, opened 2025-09-12
Playwright command

ArneTR:
This PR introduces the concept of having playwright commands directly in the usage scenario.

This was made based on an idea of @davidkopp 

To make the commands work we need IPC between playwright and GMT.

To make this low overhead we use native I/O blocking from the linux subsystem by creating two files:
- /tmp/playwright-ipc-ready 
  - This is used to signal to GMT that playwright is ready for the next command
- /tmp/playwright-ipc-commands 
  - This is used for playwright to ingest the command

The functionality is kept hidden from the user only exposing a simple string based YML syntax.

Example can be seen here: 
https://github.com/green-coding-solutions/branch-magazine-energy-tests/blob/2741f206385a5280bb275da1b5bf2db46f148185/usage_scenario_normal.yml#L66

@davidkopp How do you like this interface?

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -1545,28 +1545,28 @@ class ScenarioRunner:
                             timeout=self._measurement_flow_process_duration,
                         )
 
-                        ps_to_read_tmp.append({
-                            'cmd': docker_exec_command,
-                            'ps': ps,
-                            'container_name': flow['container'],
-                            'read-notes-stdout': cmd_obj.get('read-notes-stdout', False),
-                            'ignore-errors': cmd_obj.get('ignore-errors', False),
-                            'read-sci-stdout': cmd_obj.get('read-sci-stdout', False),
-                            'detail_name': flow['container'],
-                            'detach': cmd_obj.get('detach', False),
-                        })
+                    ps_to_read_tmp.append({
+                        'cmd': docker_exec_command,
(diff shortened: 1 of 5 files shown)


davidkopp:
I like the `playwright` command type!

Two questions...

#### Playwright API variant

The interface currently only supports the asynchronous API of Playwright, due to :
```js
await eval(`(async () => { ${data} })()`);
```

Is it sufficient to only support the asynchronous API of Playwright? If so, you might implement a check if the playwright command is missing the `await` keyword and through an error if it is missing.
If both, the synchronous and asynchronous API of Playwright should be supported, the interface needs a modification. 

#### Notes collection

How to use `read-notes-stdout: true` in conjunction with the `playwright` command type? Or would you consider this not relevant, as the flow names are already visible in the graphs?
If `read-notes-stdout: true` should be support with the `playwright` command type, I think there needs to be some modification because of the asynchronous execution (I have not tested it).
The following would not work, as only the asynchronous  (I have not tested it myself)

btw: The synchronous function `logNote(message)` is currently unused in `playwright-ipc.js`. Should it be made `async` so it can be used with the current asynchronous interface?

ArneTR:
Good input!

- In async loops sync functions can still be executed. Executing logNote should not be an issue. The function is there for general compatibility in other projects
- Jap, it needs documentation  :)

PR #1341, opened 2025-09-15
Refactor logs structure to json

### LINKED ISSUE #1335: JSON structure for logs
Currently the Logs of GMT are just a text dump.

They are lightly structured by putting words like "STDOUT" or "STDERR" at the line beginning.

This shall be changed to a JSON structured log where we still leave the output from the `exec` and `run` commands of *docker* as is, but categorize them in a JSON object structure for parsing in the fronted.

@davidkopp Please amend this issue with the structure we discussed in the call

davidkopp:
Resolves #1335 

## New log structure

Example JSON in the successful case:

```json
{
    "kadai-example-app": [
        {
            "type": "container_execution",
            "id": "131377540003888",
            "cmd": "docker run -it -d --name kadai-example-app ...",
            "phase": "[BOOT]",
            "stdout": "..."
        }
    ],
    "k6": [
        {
            "type": "flow_command",
            "id": "131377540004704",
            "cmd": "docker exec k6 k6 run ...",
            "phase": "[RUNTIME]",
            "stdout": "...",
            "flow": "Standard Usage Scenario"
        }
    ]
}
```

Example JSON in the error case:

```json
{
    "[SYSTEM]": [
        {
            "type": "exception",
            "id": "137534737070496",
            "cmd": "run_scenario",
            "phase": "[RUNTIME]",
            "stderr": "Process '['docker', 'exec', 'gcb-alpine-stres', 'stress-ng', '-c', '1', '-t', '5', '-q']' had bad returncode: 1. Stderr: ...",
            "class": "RuntimeError"
        }
    ],
    "gcb-alpine-stres": [
        {
            "type": "flow_command",
            "id": "137534735368160",
            "cmd": "docker exec gcb-alpine-stres stress-ng -c 1 -t 5 -q",
            "phase": "[RUNTIME]",
            "stderr": "Error response from daemon: No such container: gcb-alpine-stres\n",
            "flow": "Stress"
        }
    ]
}
```

## New UI

By default all are collapsed:
image

Logs of a container execution:
image

Logs of a flow command:

image

Logs of a flow command from stderr:

image

System logs (exceptions):

image

## New output using CLI flag --print-logs

Command:

```sh
python3 runner.py --name "Shell commands (--print-logs)" --uri "https://github.com/green-coding-solutions/example-applications/" --filename "shell-commands/usage_scenario.yml" --dev-no-sleeps --dev-no-save --iterations 2 --print-logs
```

Output:

```plain
Container logs:
=== Run 1: shell-commands/usage_scenario.yml (iteration 1) ===
--- Container: gcb-alpine ---
STDOUT (setup_command):
asd
123

STDOUT (flow_command):
123
xyz

STDOUT (flow_command):
/tmp:
repo

STDERR (flow_command):
ls: &&: No such file or directory
ls: echo: No such file or directory
ls: This will fail: No such file or directory
ls: &&: No such file or directory
ls: echo: No such file or directory
ls: But we force continue with ignore-errors: No such file or directory

-----------------------------
=== Run 2: shell-commands/usage_scenario.yml (iteration 2) ===
--- Container: gcb-alpine ---
STDOUT (setup_command):
asd
123

STDOUT (flow_command):
123
xyz

STDOUT (flow_command):
/tmp:
repo

STDERR (flow_command):
ls: &&: No such file or directory
ls: echo: No such file or directory
ls: This will fail: No such file or directory
ls: &&: No such file or directory
ls: echo: No such file or directory
ls: But we force continue with ignore-errors: No such file or directory

-----------------------------
```

## Open Questions

The current implementation saves the whole docker run string as part of the new JSON structure and displays it in the UI. Does this make sense?

Is the phase label of any help? In the context of the container execution it could be confusing. The label is "[BOOT]", but the logs are collected during boot, idle and runtime phase.
EDIT: Was changed to "[MULTIPLE]"

The new JSON structure is still stored in the database as TEXT for backward compatibility and not as JSONB. If we want to improve the efficiency, I can think of two options:
- use a new column 'logs_json' and leave the 'logs'  column with the old logs as they are
- provide a migration script that transforms the old text-only logs into the new JSON structure as good as possible

### DIFF
--- a/frontend/js/stats.js
+++ b/frontend/js/stats.js
@@ -109,7 +109,7 @@ const fetchAndFillRunData = async (url_params) => {
             const logsData = run_data[item];
             if (logsData === null) {
                 // Display simple message indicating no output was produced
-                document.querySelector("#logs").innerHTML = '<pre>run did not produce any output</pre>';
+                document.querySelector("#logs").innerHTML = '<pre>Run did not produce any logs to be captured</pre>';
             } else if (typeof logsData === 'object' && logsData !== null) {
                 // Handle JSON structure logs
                 // Check first if any logs have type 'legacy' - if so, render as simple text instead of structured interface


lib/scenario_runner.py
```diff
@@ -43,6 +44,12 @@
 from lib import container_compatibility
 from lib.container_compatibility import CompatibilityStatus
 
+class LogType(Enum):
```
davidkopp: I introduced an enum type here. Probably more the Java than the Python way 😉
Let me know, if I should remove it.

ArneTR: fine with that. we have that actually in multiple occurences and I personally find code more readable with that

frontend/js/stats.js
```diff
@@ -106,8 +106,22 @@ const fetchAndFillRunData = async (url_params) => {
             }
 
         } else if(item == 'logs' && run_data?.[item] != null) {
-            // textContent does escaping for us
-            document.querySelector("#logs").textContent = run_data[item];
+            const isJsonString = typeof run_data[item] === 'string' &&
```
ArneTR: note for myself: This will be reworked once refactored to JSON type

frontend/js/stats.js
```diff
@@ -106,8 +106,22 @@ const fetchAndFillRunData = async (url_params) => {
...
+                                 (run_data[item].startsWith('{') || run_data[item].startsWith('['));
+            if (isJsonString) {
+                try {
+                    const logsData = JSON.parse(run_data[item]);
+                    renderLogsInterface(logsData);
+                } catch (e) {
+                    // Fallback to original text if JSON parsing fails
+                    document.querySelector("#logs").textContent = run_data[item];
+                    document.querySelector("#logs").classList.add("plain-text");
```
ArneTR: if it is always plaintext, why is it not in the HTML code already?

frontend/js/stats.js
```diff
@@ -179,6 +193,146 @@ const fillRunTab = async (selector, data, parent = '') => {
...
+    const logsElement = document.querySelector("#logs");
+    let accordionHTML = '<div class="ui styled accordion">';
+
+    Object.keys(logsData).forEach(containerName => {
+        const containerLogs = logsData[containerName];
+
+        let contentHTML = '';
+        containerLogs.forEach(logEntry => {
+            const typeIcon = logEntry.type === 'container_execution' ? 'cog' : 'play';
+            const typeTitle = escapeString(logEntry.type.replace('_', ' ').replace(/\b\w/g, l => l.toUpperCase()));
```
ArneTR: is this supposed to only replace the first underscore? Can you please add a comment WHY you are writing this line to help understand the code in the future

frontend/js/stats.js
```diff
@@ -179,6 +193,146 @@ const fillRunTab = async (selector, data, parent = '') => {
...
+            // For exceptions, don't show the command section since it's now a label
+            const commandContent = logEntry.type === 'exception' ? '' :
+                `<div class="content">
+                    <h5 class="ui header"><i class="terminal icon"></i> Command</h5>
+                    <div class="ui segment">
+                        <code>${escapeString(logEntry.cmd)}</code>
+                    </div>
+                </div>`;
+
+            contentHTML += logCardTemplate
```
ArneTR: can any of the replacements be there more than once every?

frontend/js/stats.js
```diff
@@ -179,6 +193,146 @@ const fillRunTab = async (selector, data, parent = '') => {
...
+                .replace('{{phaseLabel}}', phaseLabel)
+                .replace('{{operationLabel}}', operationLabel)
+                .replace('{{idLabel}}', idLabel)
+                .replace('{{commandContent}}', commandContent)
+                .replace('{{stdoutContent}}', stdoutContent)
+                .replace('{{stderrContent}}', stderrContent);
+        });
+
+        accordionHTML += containerTemplate
+            .replace('{{containerName}}', escapeString(containerName))
```
ArneTR, line 397: can any of the replacements be there more than once every?

lib/scenario_runner.py
```diff
@@ -1440,22 +1448,58 @@ def _setup_services(self):
...
+            stdout: Standard output (optional)
+            stderr: Standard error (optional)
+            flow: Flow name for flow_command type (optional)
+            exception_class: Exception class name for exception type (optional)
+        """
+        if container_name not in self.__current_run_logs:
+            self.__current_run_logs[container_name] = []
+
+        log_entry = {
+            "type": log_type.value,
```
ArneTR: please adhere to our contribution.md style guide that all non-interpolated strings are single quoted. Unless  they contain single quotes themselves

ArneTR:
Also note: Tests are failing. Seems like only a stdout text matching change. However we also had other failing pipelines as docker hub pull limit was exhausted. So keep in mind it could have also been a transient error. Re-Do tests in case of doubt

frontend/js/stats.js
davidkopp: It is not always plaintext. If the new JSON structure is provided, the class is now used. But I will change it and add the class into the stats.html file and let JavaScript remove the class if the JSON was inserted successfully.

frontend/js/stats.js
davidkopp: Not sure if I understand the question fully.
The replacements are inside the `containerLogs.forEach` loop. In each container there can be multiple log entries: one `container_execution` log, multiple `setup_commands` logs, multiple `flow_command` logs and an `exception` log. So some of the values may be replaced multiple times, e.g. if there are multiple `flow_command` log entries the flow label would be replaced multiple times in the same way. But there is no replacement that is independent of the log entry in the loop.

frontend/js/stats.js
davidkopp: This is inside the `Object.keys(logsData).forEach` loop. Each loop iteration handles a different container. So the `containerName` is always different and also the `logCount`, `logPlural` and the `content` may be different.

frontend/js/stats.js
ArneTR: my comment is bc .replace() only replaces the first occurence. I just wanted to double check if that is what you wanted (aka you only expect one instance)

lib/scenario_runner.py
davidkopp: The file is full of double-quotes. Maybe this should be something for a separate PR to clean this up.
btw: In the `CONTRIBUTING.md` there is only a style guide for Python, not for JavaScript 😉

ArneTR: I also only meant that you adhere to this style guide in the python files. sorry for being not clear about this. our JS files are wild-west :)

frontend/js/stats.js
davidkopp: ah, now I understand. Yes, I only expect one instance. I double checked it and it is correct.

frontend/js/stats.js
davidkopp: double-checked and is correct.

frontend/js/stats.js
davidkopp: it is now refactored to JSON type.

frontend/js/stats.js
davidkopp: changed the implementation again.  the css class does not exist anymore. in the legacy case the text is just displayed in a `<pre>` html element.

frontend/js/stats.js
davidkopp: I added the following comment:
```js
// Make the type name more visually appealing by replacing underscores with spaces and capitalising the first letter of each word
```

lib/scenario_runner.py
davidkopp: My mistake. I changed the double-quotes I have introduced in the Python code. There are some more wrong usages, but I did not want to add changes to completely different places in the file.

lib/scenario_runner.py
```diff
@@ -43,6 +44,13 @@
 from lib import container_compatibility
 from lib.container_compatibility import CompatibilityStatus
 
+class LogType(Enum):
+    CONTAINER_EXECUTION = 'container_execution'
+    SETUP_COMMAND = 'setup_command'
+    FLOW_COMMAND = 'flow_command'
+    EXCEPTION = 'exception'
+    NETWORK_STATS = 'network_stats'  # Used in metric_importer.py (avoid circular import)
```
davidkopp: Note: this enum value is not referenced. The value is hardcoded in `metric_importer.py`, because an import of the enum in `metric_importer.py` was not possible to avoid circular imports.

Maybe the logging functionality should be extracted into an own library file.

ArneTR: why not put the enum in a separate file? we have the same logic for terminal_colors.py

davidkopp: makes sense. Done.

ArneTR:
I have now tested locally and have two remarks:

1. When I have an exception in the code that has no text the output looks wierd:
Screenshot 2025-09-17 at 11 05 39 AM

This happens for instance if I press ctrl+c in the runner during exection. Or alternative way to test is put a `raise NotImplementedError()` somewhere.

The DB has the information: 
```log
{"[SYSTEM]":[{"id":"4426307904","cmd":"run_scenario","type":"exception","class":"KeyboardInterrupt","phase":"[RUNTIME]","stderr":""}]}
```

So I would argue the exception class should be specfied 

2. When I have this flow:
```
      - type: console
        command: sleep 1
        note: Starting Stress
        log-stdout: true

```

The output will be  this:
Screenshot 2025-09-17 at 11 04 36 AM

This is I would say more a "feels suprising" than a bug. But I think it would be nice to have there a box that says "stdout:" and then just the empty box.

The current implementation gives me the feeling that GMT has not even "tried" to collect logs. Not that it tried but they where empty. How do you feel?

PR #1344, opened 2025-09-17
Add create backup script

davidkopp:
There was already an `import_backup.sh` script in the folder. To be able to quickly create database backups and import them I think it would make sense to have also a `create_backup.sh` script.

### DIFF
--- a/tools/create_backup.sh
+++ b/tools/create_backup.sh
@@ -8,15 +8,96 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 mkdir -p "$SCRIPT_DIR/backup"
 
 TIMESTAMP=$(date +"%Y%m%d_%H%M%S")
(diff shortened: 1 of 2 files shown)


ArneTR:
having a backup script like this is sure nice.

However the implementation will only have limited use in production like this.

ATM we are running backups on our cluster daily but export per table.

The issue being is that some sql files become really large. At some point we had 80GB text blobs which were impossible to handle.
Can you change the script to create one file per table. This would also mean you have to create a table list beforehand.

davidkopp:
I enhanced the scripts to support also a table-based backup and import. The import script still supports the file-based import as before.

ArneTR:
@davidkopp Are you still in need of this functionality?

If not and you are ok I would close this PR as it has become a bit stale. 

Our use cases internally have not changed and we use different means to backup. If you feel this is still relevant I would still be open to bring in a minimal version though.

davidkopp:
I do not need it at the moment as I do not make code changes with db schema modifications anymore.
Therefore I will close this PR now.

PR #1345, opened 2025-09-17
Adds better error codes

ribalba:
It has happened quite a few times that I got a link to a run and it gave me the error that the backend returned an empty response. This made me think that something with the GMT is broken. Most of the cases it was, that I didn't have access to the run. As 99% of the time I am on my local machine where I am superuser I don't think of this initially. 

This PR isn't finished it is just a hook to start a discussion if we want better error codes. I am in favour of giving the user the best information possible. I also understand that we don't want the one DB query more but would argue that it is so small that it shouldn't matter as it works on indexed data. 

Also I changed the
```
 raise RequestValidationError('Run ID is not a valid UUID or empty')
```
as you should always raise a HTTP exception

https://fastapi.tiangolo.com/tutorial/handling-errors/
https://fastapi.tiangolo.com/reference/exceptions/

### DIFF
--- a/api/api_helpers.py
+++ b/api/api_helpers.py
@@ -131,6 +131,14 @@ def get_machine_list():
     return DB().fetch_all(query)
 
 def get_run_info(user, run_id):
+
+    run_exists = DB().fetch_one(
+        "SELECT 1 FROM runs WHERE id = %s",
+        params=(run_id,)
+    )
+    if not run_exists:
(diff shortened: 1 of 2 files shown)


ribalba:
Ahhh. I wasn’t finished with this pr. As currently the errors are not really
displayed in the frontend. 



Sent from my mobile


-------- Original Message --------

PR #1356, opened 2025-09-19
Fix note 'pulling image' in charts

davidkopp:
This PR fixes 3 issues regarding the display of the note `pulling {image}` in the charts:
1. Remove unnecessary double-escaping of notes (was accidentally introduced by https://github.com/green-coding-solutions/green-metrics-tool/pull/1329)
   image
2. Fix show actual pulled image name in notes
   image
3. Add automatic text wrapping for long notes
    image

Final result:
image

### DIFF
--- a/frontend/js/stats.js
+++ b/frontend/js/stats.js
@@ -475,6 +475,35 @@ const buildTimelineChartData = async (measurements_data) => {
     return metrics;
 }
 
+const wrapNoteText = (text, maxLength = 80) => {
+    if (text.length <= maxLength) return text;
+
+    const lines = [];
+    let currentLine = '';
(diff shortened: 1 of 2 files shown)


ArneTR (approved):
Thanks for this!

I was actually suprised that this is such a complex problem in code.

I thought there is a function just for that somewhere in JS, but it turns out breaking a string properly is a really complex problem.

I looked in the Google Chrome Source Code and they spent 3000 lines on it: https://chromium.googlesource.com/chromium/src/+/163f573980212f25a361ff78c0e7bdcdbc16e70e/third_party/blink/renderer/core/layout/ng/inline/ng_line_breaker.cc

So it feels that 30 Lines are well invested here :)

PR #1359, opened 2025-09-20
Fix null byte handling in logs to prevent PostgreSQL errors

davidkopp:
If there were null bytes in the stdout or stderr, it came to the following database error:

```plain
psycopg.errors.UntranslatableCharacter: unsupported Unicode escape sequence
DETAIL: \u0000 cannot be converted to text.
CONTEXT: JSON data, line 1: ...........('127.0.0.1', 53764) unrecognised: \u0000...
unnamed portal parameter $1 = '...'
```

With this PR null bytes from stdout/stderr are cleaned when added to the internal log structure.

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -1472,9 +1472,9 @@ class ScenarioRunner:
         }
 
         if stdout is not None:
-            log_entry['stdout'] = stdout
+            log_entry['stdout'] = stdout.replace('\x00', '')  # Remove null bytes - PostgreSQL can't handle \u0000 in JSON
         if stderr is not None:
-            log_entry['stderr'] = stderr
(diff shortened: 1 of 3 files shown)


ArneTR:
Thanks for the very detailed explanations! ❤️

- I had to remove some comments which I deemd not necessary :)
- I made the representation of the the NUL byte a bit clearer in what was really replaced
- I found another case where a NUL byte can slip in and do damage: When an error gets reported that AGAIN contains the NUL byte. Thus the centralised error handler now also replaces it

davidkopp:
I'm wondering why you put the docstring of `_get_all_run_logs` above the function name. Is there a specific reason?

> A docstring is a string literal that occurs as the **first statement** in a module, function, class, or method definition.
> https://peps.python.org/pep-0257/

The rest is fine for me 😊

PR #1364, opened 2025-09-22
Add option to set timezone

Lisi4ka59:


ArneTR:
Hey @Lisi4ka59

I also wanted to chip in here with a feedback as I believe @ribalba has mixed up some feedback here. Frontend timezones are already shown in local time of the user.

---

Since you did not put any description in the PR my feedback includes some assumptions :)

I have linked an issue that we had open for this functionality. I believe your PR is related to that?
If so: How you feel about adding timezone information also to the other containers?

Or have you created the PR for a different functionality that you needed in the tool?
If so: Could you elaborate a bit what kind of features where needed or what kind of issues you ran into?

tests/setup-test-env.py
```diff
@@ -32,6 +32,7 @@
 test_frontend_config_path = os.path.join(current_dir, TEST_FRONTEND_CONFIG_NAME)
 
 DB_PW = 'testpw'
+TZ_VALUE = 'Europe/Berlin'
```
ArneTR: The timezone is now fixed to Europe/Berlin for tests.

What is the rationale here?

IMHO I would say that having no timezone information, which will default then to UTC, is better than having it set to an arbitrary location. If you run the tool in Brazil you will be quite confused.

Why not make the same inference with default settings as it is done in the install script?

tests/setup-test-env.py
```diff
@@ -128,6 +129,7 @@ def edit_compose_file():
             new_env = []
             for env in compose['services'][service]['environment']:
                 env = env.replace('PLEASE_CHANGE_THIS', DB_PW)
+                env = env.replace('__TZ__', TZ_VALUE)
```
ArneTR: is this not redundant to the replacement for all services in Line 143 ?

ArneTR (changes requested):
Thanks for the update of the PR!

I left some comments in the review where i had some remarks about the code. Happy if you give it a look

tests/setup-test-env.py
Lisi4ka59: Yes, I just forgot to delete this

tests/setup-test-env.py
Lisi4ka59: The rationale was that testing environment must be uniform, but now I see that timezone of the local system may appear elsewhere, so it is really better to set it to local systems timezone.

PR #1366, opened 2025-09-25
Fix checkboxes reset in user settings

davidkopp:
Minor fix: The checkboxes in the user settings were not reset to false after a reload if they were checked previously without saving.

### DIFF
--- a/frontend/js/settings.js
+++ b/frontend/js/settings.js
@@ -23,10 +23,13 @@ const getSettings = async () => {
     try {
         const data = await makeAPICall('/v1/user/settings');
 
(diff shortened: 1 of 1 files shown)


ArneTR:
Although I see the fix as non breaking and far cleaner than the previous code I cannot reproduce that bug.

Can you tell me what browser you are using?

Does this happen locally or also on metrics.green-coding.io?

Which checkbox are you looking at in particular?

davidkopp:
I had the issue on Firefox. Steps to reproduce the error there ...

1. Open https://metrics.green-coding.io/settings.html
2. Go to the Measurement tab
3. Add a check to an option that is currently disabled (e.g. No optimizations after run)
4. Press F5
5. Go to the Measurement tab
6. The check is still there, despite it was not saved

Testing this on Brave, the issue did not occur.

PR #1367, opened 2025-09-25
Dynamic grid carbon intensity

### LINKED ISSUE #1330: Dynamic Carbon Intensity in GMT
GMT shall be extended by a dynamic carbon intensity value and so alter it's handling of the carbon grid intensity (CGI) value as it is currently.

The carbon intensity currently comes into the GMT through a `config.yml` selectable variable.

Here the value shall in a new version take two forms:
- Fixed grid intensity value (current behaviour)
- A string selecting a pre-configured service (working name Elephant), where the grid intensity can be fetched from

The service for the last part is a sidecar container we will develop separately in this issue: https://github.com/green-coding-solutions/green-metrics-tool/issues/1333

A run in the GMT shall then calculate the operational carbon values in two was:
- Either as usual by just applying the static CGI in `phase_stats.py`
- Or by calling Elephant in thet `phase_stats.py` and asking for a time series of CGI data. The time series has the start and end times of the run itself. Then effectively mapping every data point of the energy data to the same time frame of the CGI data and multiplying them making the carbon operational value effectively "dynamic".

davidkopp:
Resolves #1330

Documentation: https://github.com/green-coding-solutions/documentation/pull/126

Open questions:
- Is the new way how carbon metrics are calculated ok? (after the normal phase stats were calculated)
- Is it ok to store the carbon intensity as integer and with the unit gram in the database? ElectricityMaps returns also only integer values, however, Carbon-Aware-Computing.com returns also float values (e.g. 270.9). With the current implementation 270.9 would be stored as 270 in the database. Should I leave it, add `round()` or store it as mg in the database (and show it as mg in the frontend)?
- Is it ok that I also store the static carbon intensity from the config as a measurement_metric with multiple values (at least one value per phase) in the database?
- With the current approach it is ensured that there are always at least one data point per phase and in addition one data point for the measurement start and end point. So at least 7 data points always exist, even if there is only one constant value. Is that ok? The reasoning was, that with this approach it aligns with the general phase stats approach and I always can use the same calculation approach, independent of the number of data points given from the API provider.
- what should be displayed as the "scope" of the new metric in the frontend? `system`, `global`, `container`, etc. don't make sense here I think. Maybe `grid` or `zone` for the dynamic value, but for static? The metric name for the dynamic carbon intensity is currently `grid_carbon_intensity_dynamic`, therefore "dynamic" is displayed as the scope.
- the new metric should be displayed as key metric, correct? which one should be replaced with it?
- EDIT: Is it ok that the grid carbon intensity metrics are only stored in database if phase stats are calculated (`dev-no-phase-stats` is disabled)?

## Screenshots

The following two screenshots were taken from the synthetic test case `test_phase_stats_dynamic_grid_carbon_intensity`.

image

image


davidkopp:
I removed the docstrings and some comments. 

> Have you thought about making it a proper metrics provider? Personally I think it might still get merged in like this but will soon be refactored into that form. So I wanted to explore if you faced any issues there. What I could see are two metric providers, one static and one dynamic, which are mutually excluside. They produce the same value though, a "grid_carbon_intensity". This will then later be used in phase_stats for furhther processing. All API calls needed can happen in the provider.py of the metrics reporter. It has however no actual binary that runs and would thus be similar to the SDIA reporter for instance.

No, I have not thought about making it a proper metrics provider. It seems that I have interpreted the following design guideline incorrectly:

> Grid Intensity is not a classic metric provider. Nor is it really a phase stat. What we is that the CGI shall be collected at the end of the run. This can be done inthe provider.py, because we need the data available in the phase_stats.py as proper metric series
> — https://github.com/green-coding-solutions/green-metrics-tool/issues/1330#issuecomment-3291032233

If you want to have one metric provider that produces a metric called "grid_carbon_intensity" then you might have issues with the display name in the frontend (see `frontend/js/helpers/config.js`), at least if you want to differentiate between static and dynamic what is currently the case.

> A round would be nice I guess though.

Done.

> How about 'grid_carbon_intensity_location'. Naming is typically done at the moment as "origin" -> "metric" -> "source" -> "scope" . Scope would be then the location it is derived for. One could argue to separate between "grid_carbon_intensity_api_location" and "grid_carbon_intensity_static_location" to include the source. In the future we could then also add "grid_carbon_intensity_api_market" to have market based values or something. How do you feel about this?

I changed the metric names to `grid_carbon_intensity_config_location` and `grid_carbon_intensity_api_location`. I went with `config_location` instead of `static_location` as config is a noun like api.

davidkopp:
The test `test_network_providers` failed again in the pipeline. 10887248 Bytes (~ 10 MB) were transferred instead of the expected 5-6 MB. The test uses the static carbon value given from the config, so there are no additional network calls introduced by this PR. So I assume the failing of the test is independent of the changes in this PR.

PR #1370, opened 2025-09-28
Stdout and Stderr of subprocesses are now displayed better in logs

ArneTR:
stdout and stderr are often merged in output from system level execptions.

For many exceptions this PR resolves this and directly transports stdout and stderr.

It will then for instance look like this for a docker build that failed:

Screenshot 2025-09-27 at 8 52 44 PM

### DIFF
--- a/tests/test_usage_scenario.py
+++ b/tests/test_usage_scenario.py
@@ -761,7 +761,7 @@ def test_invalid_container_name_2():
         runner.run()
 
     exc_str = str(e.value)
-    assert exc_str.startswith("Command: '['docker', 'run', '-it', '-d', '--name', 'highload-api-:cont'")
+    assert exc_str.startswith("Command '['docker', 'run', '-it', '-d', '--name', '8zhfiuw:-3tjfuehuis'")
     assert exc_str.endswith('returned non-zero exit status 125.')
 
     assert e.value.stderr == "docker: Error response from daemon: Invalid container name (8zhfiuw:-3tjfuehuis), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed\n\nRun 'docker run --help' for more information\n"


ArneTR:
test fixed

lib/scenario_runner.py
```diff
@@ -818,50 +818,62 @@ def _build_docker_images(self):
                     ps = subprocess.run(docker_build_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='UTF-8', errors='replace', check=False)
 
                 if ps.returncode != 0:
-                    print(f"Error: {ps.stderr} \n {ps.stdout}")
-                    raise OSError(f"Docker build failed\nStderr: {ps.stderr}\nStdout: {ps.stdout}")
+                    raise subprocess.CalledProcessError(ps.returncode, 'Docker build failed', output=ps.stdout, stderr=ps.stderr)
```
ribalba, line 821: Why are you calling one `output` and one `stderr`. Wouldn't either `stdout` or `error` be the consistent wording?

ribalba (commented):
I would probably make the stderr/stdout title a lot larger as this is not really visible in the output if you don't know what you are looking for. Is it possible to have something like:
```
-----------------------------------------------------------
|                        STDERR                            |
-----------------------------------------------------------
```

If you want to keep it like this I would add a `\n`

Screenshot 2025-09-29 at 09 50 04

ArneTR:
I already created the block. Have you seen that?

This is the relevant code part
```python3
\n\n========== Stdout ==========\n{logs_ps.stdout}\n\n========== Stderr ==========\n{logs_ps.stderr}")
```

2. The signature of subprocess.CalledProcessError is set. It is not defined by me. I thought it would be overhead to subclass the exception for identical functioanlity but just a parameter rename

Issue #1360, opened 2025-09-22
Sometimes the test "test_docker_run_multi_arch_image_with_arm64_digest_on_amd64_host_fails" fails in GitHub Actions


davidkopp:
Here two runs where the test `test_docker_run_multi_arch_image_with_arm64_digest_on_amd64_host_fails` did not raise an exception as expected and thus failed:
- https://github.com/green-coding-solutions/green-metrics-tool/actions/runs/17879739592
- https://github.com/green-coding-solutions/green-metrics-tool/actions/runs/17792358683

In most cases the test succeeds. I don't know what could cause that sometimes the test fails on GitHub Actions.
There seems to be an issue in the container compatibility detection logic, but I have no clue at the moment what it could be.

Are there differences in the servers used for GitHub Actions in regards of the docker orchestration?

ArneTR:
To my knowledge the docker CLI used in GitHub Actions is the official one from Docker. So no `nerdctl` or similar replacement.

The machines should also be plain x86_64.

The only quirk I know from GitHub Actions is that the python cache gets polluted very often (which should not be the case here) and also that it often has network hick-ups (could that be the case? ... or should that not also lead to an error?)

So the only way to debug this properly IMHO would be to add more print() debug statements into the runner flow that can be optionally turned on. We then activate that for the pipeline and also instruct pytest to capture the STDOUT.

Can you introduce the statments that would help you debug the feature further to main?


I also tried to reproduce the error on my local linux machine but test seems to work fine ...

davidkopp:
Interestingly, the arm64 image was successfully started on the GitHub Action machine.
From the logs:
```
Creating container
Checking image architecture compatibility...
Warning: Container will use emulation (image: arm64, host: amd64)
Running docker run with: docker run -it -d --name test_service -v /home/runner/work/green-metrics-tool/green-metrics-tool:/tmp/repo:ro --net GMT_default_tmp_network_2406265 alpinesha2564562b419adf48c5f3c763995d6014c123b3ce1d2e0ef2613b189779caa787192_gmt_run_tmp
Stdout: 9277c094b9d00f00be70f1c613a1a54736ce22bbd7ad15900c08ff14e8cba65b
Running commands
Started containers:  ['test_service']
```

I now have two questions I don't understand:
1. How is it possible that GitHub Action is sometimes actually able to run an ARM64 image using emulation?
[...]

PR #1371, opened 2025-09-28
Add debug outputs to investigate failing multi arch image test case

davidkopp:
Add debug outputs to investigate issue #1360 

In addition to that it also introduces two new optional input variables for the CI pipeline:
- `test-path` -> let us specify a specific test that should be executed in the pipeline instead of all
- `enable-ssh` -> let us enable a ssh connection for debugging purposes

I also added the ability to run the PR workflow manually (`workflow_dispatch`).

### DIFF
--- a/.github/actions/gmt-pytest/action.yml
+++ b/.github/actions/gmt-pytest/action.yml
@@ -16,7 +16,7 @@ inputs:
   test-path:
     description: 'Specific test path to run'
     required: false
-    default: 'test_runner.py::test_docker_run_multi_arch_image_with_arm64_digest_on_amd64_host_fails_github_actions' # Only for debugging
+    default: ''
   github-token:
     description: 'pass in your secrets.GITHUB_TOKEN'
     required: true
--- a/.github/actions/gmt-pytest/action.yml
+++ b/.github/actions/gmt-pytest/action.yml
@@ -10,9 +10,13 @@ inputs:
     required: false
     default: '.'
   tests-command:
-    description: 'The command to run the tests'
+    description: 'Base pytest command'
     required: false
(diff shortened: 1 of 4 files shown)


ArneTR:
Seems like it failed due to a permission issue of accessing the python cache .... GH Actions 😡 

- have cleared the cache
- have triggered re-run

davidkopp:
I issue was not the Python cache, it was a mistake by me (changed default value of `tests-command`). I will fix it

davidkopp:
I updated the PR description. Let me know if the changes make sense to you.

ArneTR:
The changes are reasonable and helpful, however it is unclear to me why they are in the -main.yml AND the -pr.yml.

Having it in the -main.yml would suffice, not? I can trigger any branch from there.

Furthermore: Did you already find the problem? Or what is the intent now with this PR? 
We can run it now with this setup as often as we like until we have found the error, right?
Alternative would be to bring it into main and run alongside the code "organically" until the error pops up next time, right? This is what this PR is for?

If so: Happy to merge it in

davidkopp:
> Alternative would be to bring it into main and run alongside the code "organically" until the error pops up next time, right? This is what this PR is for?

Exactly, that is the main purpose of this PR.

> Did you already find the problem?

No.

> The changes are reasonable and helpful, however it is unclear to me why they are in the -main.yml AND the -pr.yml.
> Having it in the -main.yml would suffice, not? I can trigger any branch from there.

Having it only in the main.yml is not sufficient, as it only allows to run a workflow file from a different branch, but not the code from a different branch:

image

The main.yml always makes a checkout of main. But when I want to test something from a different branch, the pr.yml would be the way to go. This is not directly related to the debugging of the failed test, but could be helpful for other issues in the future.
Does this makes sense to you now?

PR #1372, opened 2025-09-28
Unique constraint of measurement values for time is now done in code instead of Index in DB

ArneTR:
Reason being as that the index is exactly as big as the table itself due to max cardinality.

This is very costly for large datasets and checking on insert one time is cheaper here.

This also means that we can later fail more gracefully instead of only failing on insert


ArneTR:
This is waiting for rollout to our machines as it must happen synchronously with deleting the index on the main DB

Issue #1373, opened 2025-09-28
Double escaping in comparison view when comparing repositories


davidkopp:
Screenshot:

<img width="1675" height="303" alt="Image" src="https://github.com/user-attachments/assets/0873496d-38ec-4eef-a7ad-44be5a6a126c" />

This bug was probably introduced by me in #1329

PR #1376, opened 2025-09-29
Fix double escaping in JS [skip ci]

ArneTR:
Should fix https://github.com/green-coding-solutions/green-metrics-tool/issues/1373


davidkopp (approved):
LGTM 👍
Thanks for fixing my mistakes!

PR #1379, opened 2025-10-03
Full Docker Prune whitelist implemented

ArneTR:
This PR makes the --full-docker-prune flag more dynamic by allowing a whitelist of images that will not be pruned.

Helpful for cluster installations where certain images shall be retained that are not security critical but heavy in download size


ArneTR:
@greptileai

ArneTR:
@greptile

PR #1393, opened 2025-10-13
Fixes that variables where not subsituted in included files.

ribalba:
And adds docker build args.

Keeping this open as there might be more hidden bugs that I will discover once I dig deeper

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -414,7 +414,7 @@ class ScenarioRunner:
                 nodes = loader.get_constructed_nodes(node)
 
                 if not re.fullmatch(r'gmt-playwright-(?:with-cache-)?(v\d+.\d+.\d+)\.yml', nodes[0]):
-                    raise ValueError(f"You tried include unallowed files with !include-gmt-helper function. Included files must conform to regex gmt-playwright-(?:with-cache-)?(v\d+.\d+.\d+)\.yml but actually is {nodes[0]}")
+                    raise ValueError(f"You tried include unallowed files with !include-gmt-helper function. Included files must conform to regex gmt-playwright-(?:with-cache-)?(v\\d+.\\d+.\\d+)\\.yml but actually is {nodes[0]}")
 
                 filename = runner_join_paths(f"{GMT_ROOT_DIR}/templates/partials/", nodes[0], force_path_as_root=True, force_path_in_repo=False)
 


ribalba:
Still needs tests

lib/scenario_runner.py
```diff
@@ -55,14 +55,23 @@ def validate_usage_scenario_variables(usage_scenario_variables):
             raise ValueError(f"Usage Scenario variable ({key}) has invalid name. Format must be __GMT_VAR_[\\w]+__ - Example: __GMT_VAR_EXAMPLE__")
     return usage_scenario_variables
 
-def replace_usage_scenario_variables(usage_scenario, usage_scenario_variables):
+uc_replaced = tuple()
+
+def uc_string_replace(uc_string, old, new):
```
ArneTR: I think this can be done nicer with a simple list from which you delete items every time a replace has happend.

I imagine a new __usage_scenario_variables_unused = self._usage_scenario_variables.keys()

Then you can make it available in the Loader at the start of the class. Then just remove the element each time a replace has happened.
If after your two stages, include and parse, the list is non empty make an error.

I think this is cleaner than this global helper function and more readable.

lib/schema_checker.py
```diff
@@ -107,7 +107,23 @@ def check_usage_scenario(self, usage_scenario):
                     Optional('init'): bool,
                     Optional('type'): Use(self.valid_service_types),
                     Optional('image'): And(str, Use(self.not_empty)),
-                    Optional('build'): Or(Or({And(str, Use(self.not_empty)):And(str, Use(self.not_empty))},list),And(str, Use(self.not_empty))),
+                    Optional('build'): Or(
+                        And(str, Use(self.not_empty)),
+                        {
+                            Optional('context'): And(str, Use(self.not_empty)),
+                            Optional('dockerfile'): And(str, Use(self.not_empty)),
+                            Optional('target'): And(str, Use(self.not_empty)),
```
ArneTR: I like this new implementation. It is clearer about what we supprort and does not use the catch all list. Ignoring can be done with the new ignore-unknown-compose flag.

However why do you allow target? I don't see it used anywhere in our code ... ?

ribalba: Good catch. This was a left over from my experiments to add some sort of caching.

lib/scenario_runner.py
ribalba: Refactored

PR #1395, opened 2025-10-14
Calculation errors

ArneTR:
This PR fixes two calculation errors in GMT in some rarely used values:

- Max values of cumulative values in the runtime phase were calculated on wrong basis. This is not really an uncorrect value, but more a non-existent value as runtime would show a non existent max value, although there is not really one. MAX(energy) makes no sense and thus should never be shown. However also for runtime we want to show the max of the max values, which effectively happens indirectly as all runtime sub-phases are considered. So the value, where it was sensical, was never incorrect. It was showing non-null values for cumulative values where it should not
- power and energy slices were calculated wrongly. 
  - Power values where malformed when looking at the runtime phase, as reconstruction was doing a Mean, where it should have done a total
  - energy surplus from baseline phase was 3 orders of magnitude too low.

## Summary

- Correct runtime maximum and minimum reconstruction from sub-phase extrema.
- Correct runtime power aggregation to use totals.
- Correct baseline-phase surplus energy scaling.
- Set PSU power slice and container metrics to `MEAN`; retain PSU energy metrics as `TOTAL`.
- Add regression tests for runtime extrema and PSU calculations.


ArneTR:
When looking at this PR please also look at the AI suggestion to less approximate the runtime phase reconstruction:

Second update query in `reconstruct_runtime_phase()`

```python3
 # Step 3: compute weighted MEAN correctly
    DB().query('''
        WITH durations AS (
            SELECT
                run_id,
                phase,
                value AS duration_us
            FROM phase_stats
            WHERE run_id = %s
              AND metric = 'phase_time_syscall_system'
              AND detail_name = '[SYSTEM]'
              AND unit = 'us'
              AND type = 'TOTAL'
              AND hidden IS FALSE
        ),
        weighted AS (
            SELECT
                p.metric,
                p.detail_name,
                p.unit,
                SUM(p.value * d.duration_us) / NULLIF(SUM(d.duration_us), 0) AS weighted_mean,
                MAX(p.max_value) AS max_value,
                MIN(p.min_value) AS min_value
            FROM phase_stats p
            JOIN durations d
              ON d.run_id = p.run_id
             AND d.phase = p.phase
            WHERE p.run_id = %s
              AND p.phase = %s
              AND p.hidden IS FALSE
              AND p.type = 'TOTAL'
            GROUP BY p.metric, p.detail_name, p.unit
        )
        UPDATE phase_stats t
        SET value = w.weighted_mean
        FROM weighted w
        WHERE t.run_id = %s
          AND t.phase = %s
          AND t.metric = w.metric
          AND t.detail_name = w.detail_name
          AND t.unit = w.unit
          AND t.type = 'MEAN'
    ''', params=(run_id, run_id, runtime_phase, run_id, runtime_phase))
```

ArneTR:
@coderabbitai review

PR #1400, opened 2025-10-15
Playwright enhanced

ArneTR:
The playwright functionality for GMT was added one minor version ago.

However it was quite verbose to use with many setup and teardown routines.

This PR hides the complexity by:
- Creating an new magic include
- Creating a new magic volume mount in GMT directly
- Creating a new magic container name ("gmt-playwright-nodejs") [only used if magic include was used to prevent overwriting]

The resulting usage_scenario.yml for a minimal case looks like this:

```yml
---
name: "Website Test 
author: Arne Tarara <arne@green-coding.io>
description: "Opens page uncached and waits for full load"

compose-file: !include-gmt-helper gmt-playwright-v1.0.0.yml

flow:
  - name: Visit page
    container: gmt-playwright-nodejs
    commands:
      - type: playwright
        command: await page.goto("https://www.green-coding.io");
```

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -401,7 +401,10 @@ class ScenarioRunner:
             def include_gmt_helper(self, node):
                 nodes = self.get_constructed_nodes(node)
 
-                filename = runner_join_paths(GMT_ROOT_DIR, f"templates/partials/{nodes[0]}", force_path_as_root=True, force_path_in_repo=False)
+                if not re.fullmatch(r'gmt-playwright-(?:with-cache-)?(v\d+.\d+.\d+)\.yml', nodes[0]):
+                    raise ValueError(f"You tried include unallowed files with !include-gmt-helper function. Included files must conform to regex gmt-playwright-(?:with-cache-)?(v\d+.\d+.\d+)\.yml but actually is {nodes[0]}")
+
+                filename = runner_join_paths(f"{GMT_ROOT_DIR}/templates/partials/", nodes[0], force_path_as_root=True, force_path_in_repo=False)
                 return self.process_include(filename, nodes)
 
             def include(self, node):


ArneTR:
@greptile

lib/scenario_runner.py
```diff
@@ -226,7 +227,7 @@ def get_optimizations_ignore(self):
     # path with `..`, symbolic links or similar.
     # We always return the same error message including the path and file parameter, never `filename` as
     # otherwise we might disclose if certain files exist or not.
-    def _join_paths(self, path, path2, force_path_as_root=False):
+    def _join_paths(self, path, path2, force_path_as_root=False, force_path_in_repo=True):
```
ribalba, line 230: One Problem I see with this approach is that you can do things like
```
compose-file: !include-gmt-helper ../../config.yml
or
compose-file: !include-gmt-helper ../../../../../../etc/passwd
```

While this will fail with a `schema.SchemaError` it might leak passwords if we ever output the file. Another problem I see is that you can find out if certain files exists. As you will get a different error if it can't load the file. Same goes with what software is installed. 

I would write a new function that checks that the file is in the GMT sub directory

lib/scenario_runner.py
```diff
@@ -414,6 +431,9 @@ def recursive_lookup(k, d):
             usage_scenario = f.read()
             usage_scenario = replace_usage_scenario_variables(usage_scenario, self._usage_scenario_variables)
 
+            if match := re.search(r'!include-gmt-helper gmt-playwright-(?:with-cache-)?(v\d+.\d+.\d+)\.yml', usage_scenario):
```
ribalba, line 437: You could pull this check before the load and fail if there is a different file included.

lib/scenario_runner.py
```diff
@@ -414,6 +431,9 @@ def recursive_lookup(k, d):
...
+                self.__include_playwright_ipc = match[1]
```
ribalba: very small note. I would probably rename the var its it is the version and not the full ipc. Makes it easier to read below when you use the var

templates/partials/gmt-playwright-v1.0.0.yml
```diff
@@ -0,0 +1,26 @@
+services:
+    gmt-playwright-nodejs:
+        image: mcr.microsoft.com/playwright:v1.56.0-noble
+#        volumes:
+#            - /tmp/.X11-unix:/tmp/.X11-unix # for debugging in non-headless mode
+        environment:
+            DISPLAY: ":0" # for debugging in non-headless mode
```
ribalba: Do we need this if you don't mount in the unix socket?

templates/partials/gmt-playwright-with-cache-v1.0.0.yml
```diff
@@ -0,0 +1,37 @@
...
+#        ports:
+#           - 3128:3128
+
+flow-prepend:
+  - name: Starting browser IPC
+    container: gmt-playwright-nodejs
+    hidden: true
+    commands:
+      - type: console
+        command: node /tmp/gmt-utils/gmt-playwright-ipc-v1.0.0.js --headless true --browser firefox --proxy http://squid:3128
```
ribalba, line 31: Would be cool if there would be a way to also use chrome. But I can't think of anything easy and I wouldn't put more work into it. GMT Var defaults comes to mind.

lib/scenario_runner.py
ArneTR: hmm, are you commenting on my use of the function of of the functionality in general?

My addition to the signature makes the path checker also usable outside of the checked out repository and now it can be used on any path in the filesystem.

So getting to ../../../../../../etc/passwd is currently not possible as I limit the root node to GMT_ROOT_DIR.

However nicely spotted the root node should better be GMT_ROOT_DIR/templates/partials to not escape that particular directory. I missed that we a have a sensitive file in the base directory (config.yml).

Would you feel it is save enough then?

templates/partials/gmt-playwright-v1.0.0.yml
ArneTR: no need, but also does not hurt. Commented it out for now

lib/scenario_runner.py
ArneTR: I made a regex check. give it a look

lib/scenario_runner.py
ArneTR: changed the base dir

PR #1449, opened 2025-11-24
Removing repo ACLs in case /tmp has them set to inherig

ArneTR:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

Adds ACL removal logic to fix permission issues when `/tmp` has inherited ACLs.

The changes address a specific edge case where `/tmp` has ACL inheritance settings that could prevent the cloned repository folder from being executable. The fix:
- Removes ACL inheritance with `setfacl -k`
- Clears existing ACLs with `setfacl -b`  
- Explicitly sets execute permissions on the repo folder

**Issues Found:**
- Missing error handling: `subprocess.check_output()` will crash if `setfacl` is unavailable or fails
- The `acl` package is not included in `install_linux.sh` dependencies, which may cause failures on fresh installations
- Minor typo: "inhereted" should be "inherited"

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -13,6 +13,7 @@ import subprocess
 import json
 import os
 import time
+import stat
 import importlib
 import re
 from pathlib import Path
@@ -283,6 +284,13 @@ class ScenarioRunner:
         print(TerminalColors.HEADER, '\nChecking out repository', TerminalColors.ENDC)
 
         if self._uri_type == 'URL':
+            Path(self._repo_folder).mkdir(parents=True, exist_ok=True)
+            subprocess.check_output(['setfacl', '-k', self._repo_folder]) # no inheritance in subfolders of ACLs
+            subprocess.check_output(['setfacl', '-b', self._repo_folder]) # remove set ACLs
+            # Make /tmp/repo executable as this might not be the case due to inhereted ACLs in place from /tmp
+            current_mode = os.stat(self._repo_folder).st_mode
+            os.chmod(self._repo_folder, current_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+
             # always remove the folder if URL provided, cause -v directory binding always creates it
             # no check cause might fail when directory might be missing due to manual delete
             if self._branch:


lib/scenario_runner.py
```diff
@@ -283,6 +284,13 @@ def _checkout_repository(self):
         print(TerminalColors.HEADER, '\nChecking out repository', TerminalColors.ENDC)
 
         if self._uri_type == 'URL':
+            Path(self._repo_folder).mkdir(parents=True, exist_ok=True)
+            subprocess.check_output(['setfacl', '-k', self._repo_folder]) # no inheritance in subfolders of ACLs
+            subprocess.check_output(['setfacl', '-b', self._repo_folder]) # remove set ACLs
```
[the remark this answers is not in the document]
ArneTR: Good argument. We need to include acl as a requirement in the installation then

lib/scenario_runner.py
```diff
@@ -283,6 +284,13 @@ def _checkout_repository(self):
...
+            # Make /tmp/repo executable as this might not be the case due to inhereted ACLs in place from /tmp
```
davidkopp, line 290: the path is `/tmp/green-metrics-tool/repo`, not `tmp/repo`

lib/scenario_runner.py
davidkopp, line 289: why not combine the two commands into one?
```
subprocess.check_output(['setfacl', '-b', '-k', self._repo_folder]) # remove ACLs and inheritance
```

lib/scenario_runner.py
```diff
@@ -283,6 +284,13 @@ def _checkout_repository(self):
...
+            current_mode = os.stat(self._repo_folder).st_mode
+            os.chmod(self._repo_folder, current_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
```
davidkopp, line 292: Why not explicetly set the permissions we would like to have? That would make it easier to read, at least for me.

```
os.chmod(self._repo_folder, 0o755)  # rwxr-xr-x
```

ribalba:
So without testing this my 2 cents.

I am not a big fan of changing the ACLs on a dir that "other" parties might interact with. I don't think this is relevant in this case but I have used ACLs to lock certain people out of certain files in dirs. Just removing the ACL could potentially open up the checked out code to other users on a shared system. Why would the ACL be there in the first place? We don't know hot the VM/ container works so this could potentially leak code. Albei very very unlikely. 

A solution that would be "cleaner" is not checking out the repo in `/tmp` but in `~/.cache/gmt/repo`. The problem that we have here is that this is then probably not a memory filesystem. But as we don't know how the VM is set up and we don't do real measurements runs it is probably fine in this case.

PR #1455, opened 2025-11-28
Enforcing CPU and memory limits

ArneTR:
This PR introduces auto-set CPU and memory limits to prevent the host system from either running into OOM or getting timer issues with the Metric Providers.

## CPU
Currently the GMT puts all metric providers on Core 0.
This PR uses *cpuset* directives of docker to keep core 0 free.

This also introduces the minium system requirement of at least having two cores available.

## Memory
The host running GMT could easily OOM if the containers request too much memory.

Before memory limits where optional and had no limits. With this PR GMT will now auto-assign memory limits to the extent that 1 GB is always kept unavailable to the containers.
This shall save the host from getting OOM

This PR is subject to change. Discussion points:
- Is 1 GB a good limit? Typically GMT with the python runtime takes ~800 MB to run. So this would mean 200 MB left for the OS. Enough ...?
- The set configuration is very opaque. Where to display the info best? Re-paste it to the usage scenario? Show it somewhere else? Maybe a limits/utilization tab?

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR introduces automatic CPU and memory limits for Docker containers to prevent OOM and timing issues. The implementation reserves 1GB for GMT overhead, assigns remaining memory fairly to containers, and keeps CPU core 0 free for metric providers using `cpuset`.

**Key changes:**
- Adds `_populate_cpu_and_memory_limits()` function that runs before container startup
- Auto-assigns memory limits by dividing available memory (total - 1GB) equally among containers without user-specified limits
- Moves memory limits from `deploy.resources.limits.memory` to `mem_limit` for consistent handling
- Configures cpuset to exclude core 0 from container assignment
- Disables swap completely by setting `--memory-swap` equal to `--memory`
- Updates system checks to require minimum 2 CPU cores and 2GB free memory
- Refactors memory parsing into shared `utils.docker_memory_to_bytes()` function

**Issues found:**
- **Critical**: The utilization reporter at `optimization_providers/resources/utilization.py:20` only checks `deploy.resources.limits.memory`, but this is deleted and moved to `mem_limit` during processing. This will cause false warnings about missing memory limits.

**Architecture notes:**
- The same CPU count validation is performed twice (lines 626 and 1110), suggesting these could be deduplicated by storing the value as an instance variable
- The PR correctly handles edge cases like tight memory scenarios with appropriate warnings

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -646,9 +646,13 @@ class ScenarioRunner:
             if 'deploy' in service and service['deploy'] is not None and (cpus := service['deploy'].get('resources', {}).get('limits', {}).get('cpus', None)) is not None:
                 del service['deploy']['resources']['limits']['cpus']
                 service['cpus'] = cpus
-            if service.get('cpus', 0) == 0:
+            REQUESTED_CPUS = float(service.get('cpus', 0))
+            if REQUESTED_CPUS == 0:
                 # we do not want to auto enforce CPU limits. So we re-map the limit spec here to the host system for transparency and for comparing with other runs
                 service['cpus'] = SYSTEM_ASSIGNABLE_CPU_COUNT
+            elif REQUESTED_CPUS > SYSTEM_ASSIGNABLE_CPU_COUNT:
+                raise ValueError(f"You are trying to assign more cpus to service {service_name} than is available host system. Requested CPUs: {REQUESTED_CPUS}. Available CPUs: {SYSTEM_ASSIGNABLE_CPU_COUNT}")
+
 
 
         service_count = len(to_be_assigned_services)
--- a/tests/data/usage_scenarios/basic_stress_with_limits.yml
+++ b/tests/data/usage_scenarios/basic_stress_with_limits.yml
@@ -1,0 +1,37 @@
+---
+name: Test Stress
+author: Dan Mateas
+description: test
+
+services:
+  test-container:
+    type: container
+    image: gcb_stress
+    mem_limit: 100MB
+    cpus: 2
+    build:
+      context: ../stress-application
+
+flow:
+
+  - name: I am a hidden phase
+    container: test-container
+    hidden: true
+    commands:
+        - type: console
+          command: echo 1
+
+  - name: Stress
+    container: test-container
+    commands:
+      - type: console
+        command: stress-ng -c 1 -t 1 -q
+        note: Starting Stress
+
+  - name: I am hidden too
+    container: test-container
+    hidden: true
+    commands:
+      - type: console
+        command: echo 1
+        note: echo
--- a/tests/data/usage_scenarios/resource_limits_good.yml
+++ b/tests/data/usage_scenarios/resource_limits_good.yml
@@ -71,6 +71,11 @@ services:
     image: alpine
     deploy: # allowed to be None
 
+  test-container-limits-none:
+    type: container
+    image: alpine
+    # no deploy key at all
+
   test-container-cpu-and-memory-in-both:
     type: container
     image: alpine
--- a/tests/data/usage_scenarios/resource_limits_too_high.yml
+++ b/tests/data/usage_scenarios/resource_limits_too_high.yml
@@ -1,24 +1,14 @@
 ---
 name: Testing SHM
 author: Arne Tarara
-description: Testing SHM
+description: Testing Too high limits
 
 services:
   test-container:
     type: container
     image: alpine
-    cpus: "1.2"
-    mem_limit: "10MB"
+    cpus: 400
 
-    shm_size: "30MB"
-
-  test-container-2:
-    type: container
-    image: alpine
-    cpus: "1.2"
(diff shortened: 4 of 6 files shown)


ArneTR:
@greptileai

ArneTR:
@greptileai

ArneTR:
@greptileai

lib/scenario_runner.py
```diff
@@ -613,6 +614,56 @@ def _populate_image_names(self):
...
+        to_be_assigned_services = []
+        for service_name, service in services.items():
+            # wildly the docker compose spec allows deploy to be None
+            # ... thus we need to check and cannot .get()
+            if 'deploy' in service and service['deploy'] is not None and (memory := service['deploy'].get('resources', {}).get('limits', {}).get('memory', None)) is not None:
+                del service['deploy']['resources']['limits']['memory']
+                service['mem_limit'] = memory
+            if service.get('mem_limit', 0) == 0:
+                to_be_assigned_services.append(service_name)
+                self.__warnings.append(f"Service '{service_name}' had no memory limit set. GMT does not allow unbounded memory limits and auto value was applied.")
```
davidkopp, line 639: This warning message will be displayed for a lot of runs (as I assume that the majority of the GMT users haven't set resource constraints yet, or is this already the case?). Maybe it would make sense to add a link to a documentation page, e.g. https://docs.docker.com/reference/compose-file/deploy/#resources

lib/scenario_runner.py
```diff
@@ -1060,6 +1111,10 @@ def _setup_services(self):
         # This use case is when you have running containers on your host and want to benchmark some code running in them
         services = self._usage_scenario.get('services', {})
 
+        SYSTEM_ASSIGNABLE_CPU_COUNT = int(subprocess.check_output(['docker', 'info', '--format', '{{.NCPU}}'], encoding='UTF-8', errors='replace').strip()) -1
+        if SYSTEM_ASSIGNABLE_CPU_COUNT <= 0:
+            raise RuntimeError(f"Cannot assign docker containers to any CPU as no CPUs are available to Docker. Available CPU count: {SYSTEM_ASSIGNABLE_CPU_COUNT}")
```
davidkopp, line 1116: this code is duplicated. It is already part of `_populate_cpu_and_memory_limits`.
The variable `SYSTEM_ASSIGNABLE_CPU_COUNT` is required later, but there is also a bug in the usage (see next comment).

lib/scenario_runner.py
```diff
@@ -1275,24 +1330,17 @@ def _setup_services(self):
...
-                docker_run_string.append(str(cpus))
-                print('Applying CPU Limit from deploy')
-            elif cpus := service.get('cpus', None): # we only need to get resources or cpus. they must align anyway
-                docker_run_string.append('--cpus')
-                docker_run_string.append(str(cpus)) # value in (fractional) cores
-                print('Applying CPU Limit from services')
+            # apply cpuset but keep one core for GMT and metric providers free
+            # This cannot be configured via user as no knowledge of machine shall be required
+            docker_run_string.append('--cpuset-cpus')
+            docker_run_string.append(','.join(map(str, range(1,SYSTEM_ASSIGNABLE_CPU_COUNT+1)))) # range inclusive as we do not assign to 0
```
davidkopp, line 1336: I think there is bug. Based on my understanding, the `+1` is wrong here and it should be

```
docker_run_string.append(','.join(map(str, range(1,SYSTEM_ASSIGNABLE_CPU_COUNT))))
```

If there are in total 4 cores, the variable `SYSTEM_ASSIGNABLE_CPU_COUNT` has the value 3. `--cpuset-cpus` is expected to be set to `1-3`, but with the current implementation it is set to `1-4`.

davidkopp (commented):
out of interest I looked through the implementation. In general it looks good to me 😊 but I think there is a bug in the code.

ribalba:
The code looks ok from scanning over it. The bigger problem I am seeing is that we are introducing another hurdle for people to measure with the GMT. Or warning on another thing. When benchmarking a tool I don't know how much memory it will take and to be honest I don't really care.

Where I care is if the server goes down because of OOM and here I would want something in place stopping the GMT process making the server unresponsive. So I would make this feature a config var that we can enable on the servers but no warn on the desktop. 

I can see the argument that people should always specify the constraints, and I agree totally personally, but I don't see it happening in real life.

lib/scenario_runner.py
ArneTR: jap, will remember it

lib/scenario_runner.py
ArneTR: Did you take into account the maybe unexpected range function behaviour in python? The right boundary is exclusive, not inclusive.

```
>>> list(range(1,4))
[1, 2, 3]
```

Please follow up if I misunderstood the bug report

ArneTR:
Will merge this for now and apply the new changes like a container tab in new PR

PR #1459, opened 2025-12-01
Adds variables to the submit form

ribalba:
This PR adds the functionality to add 0..n variables to a run. Tried to make it look nice and not just a textbox.

Screenshot 2025-12-01 at 22 23 32


ribalba:
I had to run the tests 2 times so that it would pass. No idea what is happening here.

PR #1468, opened 2025-12-04
Adds the option to watch a different repo

ribalba:
As I can see the case quite often, that we want to monitor a repo and then execute a usage scenario from another repo here is a first idea how to do this. 

@ArneTR what do you think about this implementation as it is your code. 

I still need to write the migration and do tests. But didn't want to spend too much time on this as I am assuming a change.

### DIFF
--- a/api/object_specifications.py
+++ b/api/object_specifications.py
@@ -23,6 +23,7 @@ class Software(BaseModel):
     machine_id: int
     schedule_mode: str
     usage_scenario_variables: Optional[Dict[str, str]] = None
+    repo_to_watch_url: Optional[str] = None
 
     model_config = ConfigDict(extra='forbid')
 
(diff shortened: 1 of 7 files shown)


ArneTR:
Thanks for checking in early here.

I feel this is not something that makes any sense in GMT as it incresases the complexity of how to execute runs even more.

Acually I am not even really happy with the way that GMT currently can monitor for tag / commit changes.
It is somethingthat the tool should have no knowledge of IMHO.

I would argue this is best done in an external solution and not in GMT itself.

Having code that checks a repository, which then has a linked repository, which then needs to be checked ... that feels like too much.
Also I see edge cases not covered. Your code does now NOT execute when the nextcloud-runner repo gets changed. but only when the linked repo does. But is not what we want that if ANY gets changed we want to monitor. But then I can think of users that do want to trigger a RUN only when the BASE repo gets changed, but not the linked one and so on .... I think this goes to far.

Also it would be nice if this "external service" understands GitHub Webhooks directly and can emit a GMT run from there ...

Let's phone tomorrow about this

PR #1471, opened 2025-12-06
Backfill carbon intensity

ArneTR:
With CarbonDB now open source we now introduce also the backfill mechanism to the open source code, which was previously premium members only.

- Table columns can now be null
- Will be backfilled when CarbonDB cronjobs run.

atm it might be confusiong that carbonDB is backfilling eco-ci .... might split out later in two separate cronjobs

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR introduces carbon intensity backfilling from CarbonDB, allowing NULL values in `carbon_intensity_g` column that will be populated by cron jobs.

**Key changes:**
- Migration to allow NULL in `carbon_intensity_g` column
- Backfill logic added to populate missing carbon intensity values from German grid data
- API now returns NULL instead of -1 when carbon intensity unavailable
- Power HOG data now passes actual grid intensity instead of placeholder -1

**Critical issues found:**
- Line 67 references non-existent `carbon_ug` column (should be `carbon_kg`)
- `carbon_kg` column needs to allow NULL since code sets it to None when carbon intensity unavailable
- Migration incomplete - missing `carbon_kg` NULL constraint removal
- `validate_table_constraints()` will fail after migration since it expects columns to be NOT NULL that are now nullable

### DIFF
--- a/cron/carbondb_copy_over_and_remove_duplicates.py
+++ b/cron/carbondb_copy_over_and_remove_duplicates.py
@@ -165,20 +165,23 @@ def copy_over_scenario_runner(interval=30):
 
 def validate_table_constraints():
     data = DB().fetch_all('''
-        SELECT
-            column_name,
-            is_nullable
+        SELECT id
         FROM
-            information_schema.columns
+            carbondb_data_raw
         WHERE
-            table_name = 'carbondb_data_raw'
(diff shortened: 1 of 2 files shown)


ArneTR:
@greptileai

PR #1481, opened 2025-12-09
Removed legacy custom key type

ArneTR:
This PR aims to remove a key we introduced a while ago to the service definition in the compose file:
- `type`

Currently there never was and there currently  is no use for this key and thus it shall be removed as it poses rather a compatibility risk than any gain.

In case it can get reintroduced.

This PR will stay open for one month as RFC. If no complaings happen it will be removed.

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR removes the legacy `type` field from service definitions in usage scenario YAML files. The field was defined in the schema from the beginning but was never actually used by the system. The only valid value was `container`, which provided no functional benefit.

**Key changes:**
- Removed `valid_service_types()` validation method from `lib/schema_checker.py:57-62`
- Removed `Optional('type'): Use(self.valid_service_types)` from the service schema in `lib/schema_checker.py:102`
- Removed `type: container` lines from 57 test and example usage scenario files

**Assessment:**
The removal is clean and complete. No runtime code references the `type` field, and the schema validation properly rejects it as an unsupported key after this change. Since the field was never used functionally, this is a low-risk cleanup that simplifies the schema.

### DIFF
--- a/tests/data/usage_scenarios/basic_stress_with_limits.yml
+++ b/tests/data/usage_scenarios/basic_stress_with_limits.yml
@@ -5,7 +5,6 @@ description: test
 
 services:
   test-container:
-    type: container
     image: gcb_stress
     mem_limit: 100MB
     cpus: 2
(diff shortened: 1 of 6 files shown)


ArneTR:
@greptileai

PR #1467, opened 2025-12-04
Archive and Note functionality added to GMT

ArneTR:
This PR adds the functionality to:
- Archive runs. Effectively hiding them from the default search view
- Add Notes to runs
Screenshot 2025-12-04 at 2 37 08 PM
Screenshot 2025-12-04 at 12 45 12 PM
Screenshot 2025-12-05 at 9 32 25 AM

Screenshot 2025-12-05 at 9 32 29 AM

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR adds archive, note, and public functionality to runs, allowing users to hide runs from default search, add notes, and make runs publicly viewable.

**Critical Issues Found:**
- SQL syntax error in `api/api_helpers.py:255` - references undefined table alias `r.public` in `get_comparison_details()`
- Migration schema mismatch - missing `NOT NULL` constraints on `archived` and `public` columns compared to `structure.sql`

**Implementation Notes:**
- The `show_archived` parameter is correctly handled by FastAPI's boolean conversion ("0" → False, "1" → True)
- The update endpoint silently succeeds when updating non-existent or unauthorized runs (affects 0 rows), which differs from the `update_job` pattern but may be intentional
- Note field properly escapes content using `.value` and `.innerText` to prevent XSS

PR #1482, opened 2025-12-09
Adds a resbmit button

ribalba:
Adds a simple button to resubmit jobs if they have failed. This is useful when you have a timeout in playwright and the tests should actually work. Is only relevant for the cluster but didn't add a logic to hide on dev machines.


ArneTR:
Oh, yes! Absolute lovely!!

I was actually longing for a feature like this but didn't get my hand around where to implement it best.

I like the implementation. I also see use in the feature for non-failed runs. It could then be integrated in the "Actions" tab that I created in this PR: https://github.com/green-coding-solutions/green-metrics-tool/pull/1467

Will put that in the backlog once the other PR is merged.

This one here is good to merge IMHO. No need to wait for the tests

PR #1488, opened 2025-12-12
No backfilling of usage scenario. Rather using internal copy and freezing original

ArneTR:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR refactors usage scenario handling by maintaining an immutable frozen copy (`_usage_scenario_original`) for storage while using a separate internal working copy (`__usage_scenario`) for runtime modifications. The containers data structure is also changed from an array to a dictionary (keyed by container name) for more efficient lookups.

**Major Changes:**
- Introduced `FrozenDict` class and `freeze_dict()` function to create immutable copies of the usage scenario
- Split usage scenario into `_usage_scenario_original` (frozen, stored in DB) and `__usage_scenario` (mutable, used internally)
- Changed containers storage format from array to dictionary for better performance
- Updated frontend, tests, and optimization providers to work with new dict-based container structure

**Issues Found:**
- Test file treats containers as list when it's actually stored as dict (line 55-56 in `tests/test_resource_limits.py`)
- Schema checker incorrectly validates the original copy instead of the working copy, potentially modifying data that should remain pristine (line 526 in `lib/scenario_runner.py`)
- Error messages in duplicated `FrozenDict` class reference "GlobalConfig" instead of generic "FrozenDict" (lines 15-29 in `lib/utils.py`)

### DIFF
--- a/tests/test_resource_limits.py
+++ b/tests/test_resource_limits.py
@@ -52,8 +52,7 @@ def test_resource_limits_good():
 
     with open(f'{GMT_DIR}/tests/data/usage_scenarios/resource_limits_good.yml', 'r', encoding='utf-8') as f:
         usage_scenario_contents = yaml.safe_load(f)
-    container_list = utils.get_run_data(run_name)['containers']
-    container_dict = { ctr['name'] : ctr for ctr in container_list }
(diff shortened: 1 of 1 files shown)


ArneTR:
@greptileai

PR #1509, opened 2025-12-30
Repository relations

ArneTR:
Screenshot 2025-12-30 at 7 53 39 PM

<h3>Greptile Summary</h3>

Adds repository relations feature allowing measurement runs to checkout and mount additional repositories into containers. This enables multi-repository testing scenarios.

**Key Changes:**
- Added `relations` jsonb column to `runs` table to store relation metadata
- New `_checkout_relations()` method clones related repositories to `/tmp/relations/` and mounts them into containers at `/tmp/relations/{relation_key}:ro`
- Schema validation added for relations with support for URL, optional branch, and optional commit_hash
- Frontend displays relation information with clickable links to repository URLs
- Removed `-it` flags from docker run commands (now only uses `-d`)

**Issues Found:**
- **Critical XSS vulnerability**: `commit_hash` field not escaped in frontend HTML insertion (stats.js:162)
- Outdated comment about `-it` flag that was removed
- Missing URL protocol validation in schema checker - should restrict to git protocols only

### DIFF
--- a/lib/scenario_runner.py
+++ b/lib/scenario_runner.py
@@ -61,14 +61,19 @@ class ScenarioRunner:
     def __init__(self,
         *, uri, uri_type, name=None, filename='usage_scenario.yml', branch=None,
         debug_mode=False, allow_unsafe=False,  skip_system_checks=False,
-        skip_unsafe=False, verbose_provider_boot=False, full_docker_prune=False,
-        dev_no_sleeps=False, dev_cache_build=False, dev_no_metrics=False,
-        dev_flow_timetravel=False, dev_no_optimizations=False, docker_prune=False, job_id=None,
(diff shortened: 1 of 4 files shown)


lib/scenario_runner.py
```diff
@@ -349,6 +335,60 @@ def _checkout_repository(self):
...
+            if 'commit_hash' in relation:
+                subprocess.run(
+                    ['git', 'checkout', relation['commit_hash']],
+                    check=True,
+                    capture_output=True,
+                    encoding='UTF-8',
+                    cwd=relation_path,
+                )
+
+            if problematic_symlink := utils.find_outside_symlinks(relation_path):
```
ribalba, line 390: Why do you do this after the checkout runs?

ArneTR: The repository can contain symlinks. e.g. it possible to have a dangling symlink to /etc/passwd in the repo

PR #1530, opened 2026-01-31
Added name as filter parameter and search is now case insensitive

ArneTR:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

Added a new `name` filter parameter to the runs endpoint and converted all search queries from case-sensitive (`LIKE`) to case-insensitive (`ILIKE`). The implementation is clean with proper parameterized queries in the backend and correct URL encoding in the frontend.


ArneTR:
@greptileai

PR #1531, opened 2026-01-31
Webnrg scroll

ArneTR:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

Fixed network interface display truncation and added scroll-to-bottom functionality with a 40-second timeout.

**Changes made:**
- Removed `grep` filters from network interface commands in `lib/hardware_info.py` to display complete output including IP addresses
- Added automated scroll-to-bottom functionality in the website usage scenario with a 200-tick limit (40 seconds total at 200ms per tick)

**Notes:**
- The infinite scroll loop issue from the previous review thread has been addressed with the `max_loop_ticks` limiter
- The commit message "Limiting loop to 40 ticks" is misleading—the code actually uses 200 ticks, but the 40 seconds calculation is correct (200 × 200ms = 40s)

### DIFF
--- a/templates/website/usage_scenario.yml
+++ b/templates/website/usage_scenario.yml
@@ -22,7 +22,10 @@ flow:
     commands:
       - type: playwright
         command: |
-          while (true) {
+          let loop_ticks = 0;
+          const max_loop_ticks = 200;
+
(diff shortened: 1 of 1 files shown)


ArneTR:
@greptileai

PR #1538, opened 2026-02-04
Adds the new simulation site with elephant backend

ribalba:
This is a functional complete implementation that I have't cleaned up to date. Still some debug output there. But I wanted to get the functional aspect running first. 

@ArneTR can you please test and see if there is anything more/ different you want. 

For this you need to checkout the https://github.com/green-coding-solutions/elephant service and the do `docker compose up --build ` 

Every run then gets a new button under the Actions tab that opens the simulation page where you can play around. 

You will also need to rerun the installer in the GMT repo as it needs to set the elephant URL. You will be asked to add it when installing.

Edit: Not quite clear why the tests are failing. Seems to be not related to this PR.


lib/install_shared.sh
```diff
@@ -651,6 +658,17 @@ if [[ -z $metrics_url ]] ; then
     metrics_url=${metrics_url:-"http://metrics.green-coding.internal:9142"}
 fi
 
+if [[ -z $elephant_url ]] ; then
```
ArneTR: This will fail in the automated pipeline, as it requires stdin. You need to supply an argument to either force-install or force-ignore it. See the lines ff where to append: https://github.com/green-coding-solutions/green-metrics-tool/blob/0ae98ba2558461e1c6dde2060a5aed70222faf45/.github/actions/gmt-pytest/action.yml#L54

frontend/js/stats.js
```diff
@@ -72,6 +72,13 @@ class CO2Tangible extends HTMLElement {
 
 customElements.define('co2-tangible', CO2Tangible);
 
+const getElephantServiceUrl = () => {
+    if (typeof ELEPHANT_URL === 'string' && ELEPHANT_URL.trim() !== '') {
```
ArneTR: Maybe just reduce to:
- return typeof ELEPHANT_URL === 'string' ? ELEPHANT_URL.trim() : ''

frontend/js/simulation.js
```diff
@@ -0,0 +1,1426 @@
...
+const MINI_CHART_COLORS = {
+    provider: '#4a7a9c',
+    metric: '#5c9d6e'
+};
+const PROVIDER_HISTORY_PAST_HOURS = 24;
+const PROVIDER_HISTORY_FUTURE_HOURS = 12;
+const METRIC_SHIFT_STEP_MS = 10 * 60 * 1000;
+const METRIC_SIMULATION_SHIFT_STEP_MS = 1 * 60 * 1000;
+
+const SIMULATION_DEBUG = true;
```
ArneTR: ?

frontend/js/simulation.js
```diff
@@ -0,0 +1,1426 @@
+
+const MINI_CHART_COLORS = {
+    provider: '#4a7a9c',
+    metric: '#5c9d6e'
+};
+const PROVIDER_HISTORY_PAST_HOURS = 24;
+const PROVIDER_HISTORY_FUTURE_HOURS = 12;
+const METRIC_SHIFT_STEP_MS = 10 * 60 * 1000;
```
ArneTR: I dont quite get the variable name. This shifts by 600 ms? Or by 600.000 ms ?

frontend/js/simulation.js
```diff
@@ -0,0 +1,1426 @@
...
+const METRIC_SIMULATION_SHIFT_STEP_MS = 1 * 60 * 1000;
+
+const SIMULATION_DEBUG = true;
+const query = (selector) => document.querySelector(selector);
+const setTextContent = (selector, value) => {
+    const element = query(selector);
+    if (element) {
```
ArneTR: I think this still of shadowing the effect will be hard to debug. You get no feedback that the select has failed and it silently discards the event ...

Sane for next two functions.

frontend/js/simulation.js
```diff
@@ -0,0 +1,1426 @@
...
+
+const renderRunDetails = (runData, runTimes) => {
+    setTextContent('#run-id', runData?.id ?? '-');
+    setTextContent('#run-start', runTimes ? new Date(runTimes.startMs).toLocaleString() : '-');
+    setTextContent('#run-end', runTimes ? new Date(runTimes.endMs).toLocaleString() : '-');
+};
+
+const fetchRunData = async (runId) => {
+    const run = await makeAPICall(`/v2/run/${runId}`);
+    return run?.data;
```
ArneTR, line 535: why is the return approach here different than in 477 ?

frontend/js/simulation.js
```diff
@@ -0,0 +1,1426 @@
...
+    if (metrics.length === 0){
+        return [];
+    }
+
+    const psuMetrics = metrics.filter((metric) => metric.startsWith('psu_energy_'));
+    if (psuMetrics.length > 0) {
+        return psuMetrics;
+    }
+
+    return metrics.filter((metric) => metric.includes('energy'));
```
ArneTR: you should limit to _energy_ to be more precise. Otherwise technicaly a provider named energyapi or so could be returned. _X_ encapsulates it better

frontend/js/simulation.js
```diff
@@ -0,0 +1,1426 @@
...
+const buildEnergyData = (measurements, metric) => {
+    if (!Array.isArray(measurements) || measurements.length === 0) return null;
+    if (!metric) return null;
+
+    const series = {};
+    let unit = null;
+
+    measurements.forEach((entry) => {
+        if (entry?.[2] !== metric) return;
+        const detail = entry?.[0] || '[MACHINE]';
```
ArneTR: Is that ever missing?

lib/install_shared.sh
ribalba: Added a new flag `--no-elephant` to avoid this and added it to the GitHub Actions

frontend/js/stats.js
ribalba: Fair point. Fixed

frontend/js/simulation.js
ribalba: This is for me for debugging. But I set it to false now.

frontend/js/simulation.js
ribalba: Added a comment

frontend/js/simulation.js
ribalba: Good point. I added some error logging.

frontend/js/simulation.js
ribalba: Not quite sure what you mean here. Do you mean that I return an `[]` if no measurements are found?

frontend/js/simulation.js
ribalba: You mean it should start with energy? What do you mean with ` X encapsulates it better`

ribalba: Ahhh. You meant `_energy_`. Ok fixed

frontend/js/simulation.js
ribalba: It must have been. But I can not reproduce. Took it out till we see the error again

ribalba:
@ArneTR 

You don't need the elephant url in the `config.yml` as it is only set in the `frontend/js/helpers/config.js`

I can not reproduce your error with the `/y` maybe you entered it in the url while running the install. Can you please rerun the install and see if the error persists?

frontend/js/simulation.js
ArneTR: yes

frontend/js/simulation.js
ArneTR: if you need this why not have it permanently but dynamic?

Checking for window.SIMULATION_DEBUG can do the trick. This one you can set from your Chrome Developer Tools

frontend/js/simulation.js
ArneTR: You still have this behaviour throughout your code and it is not clear to me why.

Some more examples:

https://github.com/green-coding-solutions/green-metrics-tool/blob/274b4f9c56108585392b9438aac08b23c29a9ea5/frontend/js/simulation.js#L43

https://github.com/green-coding-solutions/green-metrics-tool/blob/274b4f9c56108585392b9438aac08b23c29a9ea5/frontend/js/simulation.js#L69

and at least 3 more

Now you also added a logging on top. Why?

Why not just crash the code? Should this really ever happen that you supply a selector or a button which then is null? I feel this guard clause is not for an actual dynamic interface but for a potential extenral developer who uses the functions wrongly ... or am I missing something?

My implementation would be to remove all these guard clauses effectively exposing the errors which then can be fixed.

ribalba: As discussed I removed everything

frontend/js/simulation.js
ribalba: I like the idea with window. Changed

ribalba:
@ArneTR all fixed

PR #1546, opened 2026-02-09
Hardening the subit by making the file also unwriteable by group

ArneTR:
Technically not needed as the subit already pops when the file is modified anyway.

But there is no need to have the file group writeable, so locking it down

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR hardens several setuid metric-provider build targets by changing the post-build ownership step from `sudo chown root $@` to `sudo chown root:root $@`, ensuring the resulting binaries are owned by group `root` rather than inheriting a potentially writable group.

This fits into the codebase’s pattern of compiling small metric-provider binaries and then setting ownership/permissions (including setuid) so they can access privileged system interfaces at runtime.

Issues found:
- Several Makefiles now lack a trailing newline, producing `\ No newline at end of file` diffs and potentially tripping strict text-file checks.


ArneTR:
@greptileai

PR #1560, opened 2026-02-12
Logical draft of how the metric processors shall work

ArneTR:
This PR outlines the concept of introducing two new stages to the GMT:

- Post Metric Processors
- Post Phases Stat Processors

They run in the pipeline as follows:
Metric Providers -> Post Metric Processors -> Phase Stats -> Post Phase Stats Processors

The already commited `config.yml.example` outlines how they are defined in the config.

## Simplifications and Domain Logic

This PR simplifies the mental load on where processing happens in the GMT and makes it more explicit and more modular.

- The `phase_stats.py` will then really only create avg and aggregate values of metric data
- SCI will not be all over the place but in a separate stage. Making it also more easy to get SCI values per phase
- Overloaded Providers like XGBoost and SDIA are then in their separate stage. Not having the need to be processed last in the metric provider stage and loading in "hopefully" generated values by other providers ... which is currently only by accident :)

## When we implement this approach the following refactorings need to be done:

- Rip out the `calculate_co2_intensity` from the `phase_stats.py` and bring it to a Post Metric Processor (https://github.com/green-coding-solutions/green-metrics-tool/blob/d0a621556f7e9c7b188b7dd43514b77106bbb04d/lib/phase_stats.py#L81)
- Migrate the XGBoost and SDIA providers to be Post Metric Processors
- Migrate the SCI generation code from the `phase_stats` and bring it to the Post Phases Stats Processors 
- Look at https://if.greensoftware.foundation/pipelines/ to see if we can take some good ideas from how IF thinks about pipelines. We want it less complex. But still might find some good work there

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -204,61 +204,86 @@ measurement:
     common:
 #      network.connections.proxy.container.provider.NetworkConnectionsProxyContainerProvider:
 ##        host_ip: 192.168.1.2 # This only needs to be enabled if automatic detection fails
-    #--- Model based - These providers estimate rather than measure. Helpful where measuring is not possible, like in VMs
-#      psu.energy.ac.sdia.machine.provider.PsuEnergyAcSdiaMachineProvider:
-      #-- This is a default configuration. Please change this to your system!
-#        CPUChips: 1
-#        TDP: 65
-#      psu.energy.ac.xgboost.machine.provider.PsuEnergyAcXgboostMachineProvider:
-      #-- This is a default configuration. Please change this to your system!
-#        CPUChips: 1
-#        HW_CPUFreq: 3200
-#        CPUCores: 4
-#        CPUThreads: 4
-#        TDP: 65
-######### The value for memory must be in GB not in GiB
-#        HW_MemAmountGB: 16
-#        Hardware_Availability_Year: 2011
-######### vhost_ratio is the virtualization degree of the machine. For Bare Metal this is 1. For 1 out of 4 VMs this would be 0.25 etc.
-#        VHost_Ratio: 1
-#
-###### DEBUG
-#      network.connections.tcpdump.system.provider.NetworkConnectionsTcpdumpSystemProvider:
+##      network.connections.tcpdump.system.provider.NetworkConnectionsTcpdumpSystemProvider:
 #        split_ports: True
 #--- END
 
+# Processors
+  # These processors run after the metric_providers stage
+  # Typical uses cases are converting raw metric provider data into another, synthetic raw metric provider. For instance
+  # taking a psu_energy_* provider and a grid_intensity_* provider and chaining it into a psu_carbon_* metric
+  # Another alternative is to take a cpu_utilization_* value and converting into a psu_energy_* value like Cloud Energy
+  post_metric_provider_processors:
+    SDIAPostMetricProviderProcessor:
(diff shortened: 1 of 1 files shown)


config.yml.example
```diff
@@ -204,61 +204,86 @@ measurement:
...
-#--- END
+#      VHost_Ratio: 1
+    Energy2CarbonPostMetricProviderProcessor:
+      # Processor outputs for instance cpu.energy.rapl.msr.component.provider -> cpu.carbon.rapl.msr.component.provider
+      # effectively translating an energy value to a carbon value
+      input:
+        - cpu.energy.rapl.msr.component.provider.CpuEnergyRaplMsrComponentProvider:
+        - psu.energy.ac.mcp.machine.provider.PsuEnergyAcMcpMachineProvider
+        # - ...
+      carbon_input: grid.intensity.elephant.api.global.GridIntensityElephantApiGlobalProvider
```
ribalba, line 248: Why are you creating a new key `carbon_input`? Wouldn't this just be another input?

config.yml.example
```diff
@@ -204,61 +204,86 @@ measurement:
...
+#--- END
+
+# Processors
+  # These processors run after the metric_providers stage
+  # Typical uses cases are converting raw metric provider data into another, synthetic raw metric provider. For instance
+  # taking a psu_energy_* provider and a grid_intensity_* provider and chaining it into a psu_carbon_* metric
+  # Another alternative is to take a cpu_utilization_* value and converting into a psu_energy_* value like Cloud Energy
+  post_metric_provider_processors:
+    SDIAPostMetricProviderProcessor:
+      # Processor outputs psu_energy_ac_xgboost_machine
```
ribalba, line 218: Doesn't the SDIA output a linear model and not XGBoost?

config.yml.example
```diff
@@ -204,61 +204,86 @@ measurement:
...
+        - psu.energy.ac.mcp.machine.provider.PsuEnergyAcMcpMachineProvider
+        # - ...
+      carbon_input: grid.intensity.elephant.api.global.GridIntensityElephantApiGlobalProvider
 
 
-sci:
+
+  post_phase_stats_processors:
     # https://github.com/Green-Software-Foundation/sci/blob/main/Software_Carbon_Intensity/Software_Carbon_Intensity_Specification.md
+    SCI:
```
ribalba, line 254: I would still add a processor script here that then has the SCI values as parameters. Like with metric providers.

config.yml.example
```diff
@@ -204,61 +204,86 @@ measurement:
...
 #      network.connections.proxy.container.provider.NetworkConnectionsProxyContainerProvider:
 ##        host_ip: 192.168.1.2 # This only needs to be enabled if automatic detection fails
-    #--- Model based - These providers estimate rather than measure. Helpful where measuring is not possible, like in VMs
-#      psu.energy.ac.sdia.machine.provider.PsuEnergyAcSdiaMachineProvider:
+##      network.connections.tcpdump.system.provider.NetworkConnectionsTcpdumpSystemProvider:
+#        split_ports: True
+#--- END
+
+# Processors
+  # These processors run after the metric_providers stage
```
ribalba, line 212: I would make clear that they are called in order they are defined here. This is important if a "carbon" processor relies on an "energy" processor.

ribalba (commented):
Perfect. Is what we discussed. The more I think about this the more I like how clean it separates concerns.

PR #1585, opened 2026-02-25
Shell command

ribalba:
Something to discuss :) I think this is really really useful but see the arguments against it. 

Maybe just run it 
```
./shell.py sleep 5
```

## Summary

- Added `shell.py` to run commands through GMT measurement on the host.
- Added host execution support with capability checks and container metric-provider filtering.
- Added `--print-phase-stats-table` for aligned phase metric output.
- Added shell argument handling, error reporting, cleanup, and exit-status mapping.
- Added schema validation, migrations, seed-data updates, and test coverage for host execution and shell mode.


ribalba:
So right now I am using it for the benchmarking of the new X tooling. I need a comparison of the overhead of the docker containers. I am aware, that the measurements will not be 100% but I want to get an idea of degrees of. I also want to use the GMT compare functionality to easily look at how things differ. I can run the job 5 times and then get a good idea how big the jitter is (Not very large on my Laptop if I kill most processes. Under 2%) I discard runs when some cron script started or there is one job which is out of line. 

Then I very often have something and I just want to get an idea of how much resources it uses but don't want to go the full docker way. 

CodeCarbon is great but I am used to the GMT tooling and I want the dashboard so I can compare and dig deeper. Not all things need to be 100%, especially if need something fast to give me an idea if it is worth investing my time. 

To be merged this needs some work, I agree. This is only something I am using privately and I wanted to share as I finally had the time to clean it up a little. I didn't want to touch anything in the "core" GMT as if this is not merged I will have to maintain it off record and I don't want merge conflicts. That is why a lot of the logic is in the shells file and not in the "core" code.

It is always the discussion: ease of use vs correct measurements.

ArneTR:
This will be kept open and re-worked when we need it again.

Effectively when we think this bigger:
- This can a tool to measure the host with GMT. By completely neglecting docker containers but still starting metric providers
- This can be a general replacement for run-template.sh
- The host mode might be a feature we need when we want to enable application measurement on windows where it is unlikely that we can encapsulate the to-be-measured applications into a docker container - See: https://github.com/green-coding-solutions/dbu-caso-code/issues/11

ribalba:
@ArneTR so this PR has changed a little in that the GMT now also supports host mode execution. It also keeps the shell command which is now just a wrapper that creates a usage scenario with the command wrapped. 
I also added a new command `--print-phase-stats-table` which prints out a nice table on the shell so you can see the values right away without needing to go to the DB. 

Ready for initial review

lib/scenario_runner.py
```diff
@@ -817,6 +820,12 @@ def _initial_parse(self):
         if self._allow_unsafe:
             print(TerminalColors.WARNING, arrows('Warning: Runner is running in unsafe mode'), TerminalColors.ENDC)
 
+        host_flows = [flow['name'] for flow in self.__usage_scenario.get('flow', []) if flow_runs_on_host(flow)]
```
ArneTR: The maintainability would be better with a any() and not User(...) if clause. Using first a list comprehension and then a 2-level if clouds this intent for me as a reader

lib/scenario_runner.py
```diff
@@ -2201,6 +2210,10 @@ def _start_metric_providers(self, allow_container=True, allow_other=True):
             if not metric_provider._metric_name.endswith('_container') and metric_provider._metric_name != 'network_connections_tcpdump_system' and not allow_other:
                 continue
 
+            if metric_provider._metric_name.endswith('_container') and not self.__containers:
```
ArneTR: maybe also a warning that you see in the UI later? Or do you feel that would get to verbose?

lib/scenario_runner.py
```diff
@@ -2222,6 +2235,9 @@ def _start_metric_providers(self, allow_container=True, allow_other=True):
             if not metric_provider._metric_name.endswith('_container') and metric_provider._metric_name != 'network_connections_tcpdump_system' and not allow_other:
                 continue
 
+            if metric_provider._metric_name.endswith('_container') and not self.__containers: # was skipped and never booted
```
ArneTR: Thinking out loud: May moving the providers dynamically to the "Disabled Providers" array? We have that already as manually configurable but could also use it for this use case ...?

lib/scenario_runner.py
```diff
@@ -2285,20 +2301,28 @@ def _run_flows(self):
...
 
                 for cmd_obj in flow['commands']:
                     self._check_total_runtime_exceeded()
 
                     if 'note' in cmd_obj:
-                        self.__notes_helper.add_note( note=cmd_obj['note'], detail_name=flow['container'], timestamp=int(time.time_ns() / 1_000))
+                        self.__notes_helper.add_note( note=cmd_obj['note'], detail_name=flow_detail_name, timestamp=int(time.time_ns() / 1_000))
 
-                    print(TerminalColors.HEADER, '\nExecuting ', cmd_obj['type'], 'command on container', flow['container'], TerminalColors.ENDC)
+                    if run_on_host:
```
ArneTR: Simplify to print(TerminalColors.HEADER, '\nExecuting ', cmd_obj['type'], 'command on', flow_detail_name, TerminalColors.ENDC)

lib/scenario_runner.py
```diff
@@ -2285,20 +2301,28 @@ def _run_flows(self):
 
             print(TerminalColors.HEADER, '\nRunning flow: ', flow['name'], TerminalColors.ENDC)
 
+            run_on_host = flow_runs_on_host(flow)
```
ArneTR: why not call it runs_on_host with s just like the function name?

lib/scenario_runner.py
```diff
@@ -2313,19 +2337,27 @@ def _run_flows(self):
                         print(TerminalColors.WARNING, arrows('Process output is streamed. Please note that this disallows capturing of errors and build outputs in logs and error messages.'), TerminalColors.ENDC)
 
                     if cmd_obj['type'] == 'playwright':
-                        docker_exec_command.append(cmd_obj.get('shell', 'sh'))
-                        docker_exec_command.append('-ec')
+                        if run_on_host: # should already be caught by the schema checker. Last line of defense
```
ArneTR: debateable. I would rather prefer a test for that. 

I know we have that in multiple places these double guards, but in the end they pollute the code

lib/schema_checker.py
```diff
@@ -11,6 +11,10 @@
 VALID_CHARS_SPACE = VALID_CHARS.copy()
 VALID_CHARS_SPACE.add(' ')
 
+def flow_runs_on_host(flow):
```
ArneTR, line 14: I would not go for python objects here but stick to the YAML syntax.

Either the value is omitted or it is null without quotes. that produces an actual python None

lib/schema_checker.py
```diff
@@ -297,7 +302,12 @@ def check_usage_scenario(self, usage_scenario):
                 raise SchemaError(f"The 'name' field in 'flow' must be unique. '{flow['name']}' was already used.")
             known_flow_names.append(flow['name'])
 
+            runs_on_host = flow_runs_on_host(flow)
+
             for command in flow['commands']:
+                if runs_on_host and command['type'] != 'console':
```
ArneTR, line 315: I would also rather do a test that schema checker works as expected here than an extra guard in code

ArneTR (changes requested):
Code looks good. Did not check the tests or the shell script though. 

Now firing up AI

ArneTR:
@coderabbitai review

ArneTR:
The tests are OK. The errors shown seem to be related to an test-data issue for a commit I pushed yesterday. WIll investigate separately

ArneTR:
I made a merge commit with weaving in the changes I did in the big PR about test refactoring. Should be up to date now and functionally unchanged

ArneTR:
Sadly the merge did not cut it ... somehow the tests seem to look for an outdated plain container name without the parallel-worker suffix ...

I cannot see why atm ... maybe you see more on monday with fresh eyes ...

ribalba:
Let's try those tests again :)

ribalba:
@ArneTR so tests seem to work now

ArneTR:
@coderabbitai review

lib/scenario_runner.py
```diff
@@ -1176,6 +1185,14 @@ def _import_metric_providers(self):
             print(TerminalColors.WARNING, arrows('No metric providers were configured in config.yml. Was this intentional?'), TerminalColors.ENDC)
             return
 
+        # A usage scenario without services will never have containers. Instead of guarding all
+        # call sites we move the container bound metric providers to the disabled providers.
+        if not self.__usage_scenario.get('services'):
+            for metric_provider in metric_providers:
+                if (metric_provider.endswith('_container') or metric_provider == 'network_connections_tcpdump_system') and metric_provider not in self._disabled_metric_providers:
```
ArneTR: why also the network_connections_tcpdump_system ?

lib/host_platform.py
```diff
@@ -66,6 +66,18 @@ def clear_file_system_caches():
     )
 
 
+def shell_command_argv(shell, command, shell_options=()):
+    # cmd and powershell do not understand the POSIX '-c' convention, so we must map to their native invocation syntax.
+    # They also have no equivalent for the POSIX shell options GMT sets (see process_helpers.DEFAULT_SHELL_OPTIONS),
+    # so 'shell-options' from the usage_scenario only ever reaches a POSIX shell.
+    shell_binary = Path(shell).name.lower().removesuffix('.exe')
+    if shell_binary == 'cmd':
```
ArneTR, line 80: Why are shell options simply discarded?

 I understand that the default shell options from linux do not make sense here, but then the default shell options should be derived differently when the given shell is cmd / powershell / pwsh in process_helpers.get_shell_options ... otherwise GMT will silently skip this and it will feel very confusing to use

ArneTR (changes requested):
I had some remarks on implementation especially for windows.

 Will also comment on the requests from Coderabbit which seem to be even an extended scope than this PR only

lib/scenario_runner.py
```diff
@@ -869,6 +872,12 @@ def _initial_parse(self):
         if self._allow_unsafe:
             print(TerminalColors.WARNING, arrows('Warning: Runner is running in unsafe mode'), TerminalColors.ENDC)
 
+        has_host_flows = any(flow_runs_on_host(flow) for flow in self.__usage_scenario.get('flow', []))
+        if has_host_flows and not User(self._user_id).can_use_orchestrator('host'):
+            raise PermissionError(f"The usage scenario contains flows that are configured to run directly on the host (container: null), but host execution is not permitted for your user (id: {self._user_id}). The user needs the 'host' orchestrator capability (measurement.orchestrators.host).")
+        if has_host_flows:
+            self._append_and_print_warning('Some flows will execute directly on the host system (container: null) instead of inside a container. These commands are not sandboxed and the measurement data is not directly comparable to fully containerized runs.')
+
```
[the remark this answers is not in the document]
ArneTR: in a follow up PR. This is acceptable for now

lib/scenario_runner.py
```diff
@@ -2375,10 +2392,14 @@ def _run_flows(self):
...
 
-            resolved_flow_container = utils.container_name(flow['container'])
-
             print(TerminalColors.HEADER, '\nRunning flow: ', flow['name'], TerminalColors.ENDC)
 
+            # utils.container_name() appends this worker's xdist suffix under test runs, exactly as
+            # _setup_services() did when it created the container. Every docker command below must
+            # target that same suffixed name, or it 404s the moment a worker suffix is in play.
+            runs_on_host = flow_runs_on_host(flow)
+            resolved_flow_container = HOST_EXECUTION_DETAIL_NAME if runs_on_host else utils.container_name(flow['container'])
```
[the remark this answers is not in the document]
ArneTR: This suggestion is plain wrong. flow['container'] is always correct

ArneTR:
Ok, I just merged in all the changes that Coderabbit commented on, which where outside the diff range in other files.
Mostly stuff from filter_sensitive_values and some stuff with clearing them.

Log: 
- https://github.com/green-coding-solutions/green-metrics-tool/commit/dc9ef0764c3579339fe6f02aac57523ea7388a60
- https://github.com/green-coding-solutions/green-metrics-tool/commit/7c95bf5a123bdbe32a0cba4f100d7e2b2fd809cf
- https://github.com/green-coding-solutions/green-metrics-tool/commit/15f45d0f89985f95a7b293f2fc9586a6321239b3
- https://github.com/green-coding-solutions/green-metrics-tool/commit/415a0929b8e2c8c872262deceb747f0ecfd5f0e8


What is left-over are changes that were attributed to shell.py, but somehow could not be posted.

Here again for you to fix please ( you can also find them above under the caret after the coderabbit comment)

---

shell.py (5)

## 136-142: 🎯 Functional Correctness | 🟡 Minor | ⚡ Quick win

Preserve command argument boundaries.

" ".join(command_parts) is not verbatim. For example, ./shell.py -- python -c 'print("hello world")' becomes python -c print("hello world"), which can change the script passed to the inner shell. Define one unambiguous command-input form and preserve quoting boundaries for arguments containing spaces.

Also applies to: 151-151

168-171: 🎯 Functional Correctness | 🟡 Minor | ⚡ Quick win

Route config override failures through the CLI error handler.

A missing, unreadable, malformed, or incorrectly suffixed config file raises before the try block. The CLI then prints an uncaught traceback instead of using _report_shell_error and the normal exit-code path. Move config validation and loading inside the try block, or handle these errors explicitly.

## 200-203: 🩺 Stability & Availability | 🟡 Minor | ⚡ Quick win

Clean temporary files on failure as well as success.

ScenarioRunner.run() cleans runner resources in its finally block but does not remove runner._tmp_folder. The cleanup in shell.py runs only after runner.run() returns, so failures, timeouts, and interrupts can leave the temporary folder behind. Move the guarded removal into finally, after error reporting.

## 71-72: 🔒 Security & Privacy | 🟠 Major | ⚡ Quick win

Sensitive Data Exposure (CWE-312): Cleartext Storage of Sensitive Information

Reachability: External

Do not persist the raw shell command in run metadata.

_arguments is serialized into runs.runner_arguments, and filter_sensitive_data() does not redact arbitrary command contents. Remove shell_command from _arguments and use a constant or redacted value for persisted usage-scenario metadata.

## 120-121: 🔒 Security & Privacy | 🟠 Major | 🏗️ Heavy lift

Authorization Bypass (CWE-639): Authorization Bypass Through User-Controlled Key (IDOR)

Reachability: External · Exploitability: Moderate

Do not use --user-id as the authorization subject.

args.user_id flows directly into ScenarioRunner._user_id, and _initial_parse() grants host execution when User(self._user_id).can_use_orchestrator('host') succeeds. A caller can select user ID 1, the default local user with the host capability, and bypass the intended capability boundary. Bind authorization to the authenticated caller, or prevent --user-id from selecting the authorization subject.

lib/host_platform.py
ribalba: I now error when this happens and handle powershell as a seperate shell

lib/scenario_runner.py
ribalba: Because this would then dump the whole system which I felt was very spammy. Removed it now so that it is also added

ribalba:
@ArneTR feedack added

lib/host_platform.py
```diff
@@ -66,18 +67,48 @@ def clear_file_system_caches():
...
+        return 'powershell'
+    return 'posix'
+
+
+def shell_command_argv(shell, command, shell_options=()):
+    family = shell_family(shell)
+
+    if family == 'cmd':
+        # cmd cannot express the strictness the POSIX options express. process_helpers.get_shell_options()
+        # therefore never returns options for cmd and raises if the user set some explicitly, so that they
```
ArneTR: Your comments says you raise if shell options are supplied. But you don't ...?

ArneTR: I see now you do that at a different path at the code. But would it not here be the most expected ... because here you silently ignore it just to pick it up somehwere else? 

In get_shell_options I think it is fine. cmd returns no shell options. But when applying them there the error should happen ... as applying something non empty is forbidden with cmd.

Maybe you can move the runtime check over here?

lib/scenario_runner.py
```diff
@@ -1195,7 +1195,7 @@ def _import_metric_providers(self):
         # call sites we move the container bound metric providers to the disabled providers.
         if not self.__usage_scenario.get('services'):
             for metric_provider in metric_providers:
-                if (metric_provider.endswith('_container') or metric_provider == 'network_connections_tcpdump_system') and metric_provider not in self._disabled_metric_providers:
+                if metric_provider.endswith('_container') and metric_provider not in self._disabled_metric_providers:
```
ArneTR, line 1198: here the removal of tcpdump is OK. as it works on the system level. No need to remove it. 
=> Nothing to change. this is just an FYI comment

lib/scenario_runner.py
```diff
@@ -2310,10 +2310,10 @@ def _start_metric_providers(self, allow_container=True, allow_other=True):
         # This includes tcpdump, which is only for debugging of the containers itself
         # If debugging of the tool itself is wanted tcpdump should be started adjacent to the tool and not inline
         for metric_provider in self.__metric_providers:
-            if (metric_provider._metric_name.endswith('_container') or metric_provider._metric_name == 'network_connections_tcpdump_system' ) and not allow_container:
+            if metric_provider._metric_name.endswith('_container') and not allow_container:
```
ArneTR: Here tcpdump should not be removed. as now the provider will not start anymore anywhere.

The reason why we have this separation is that tcpdump creates A Lot of debug data. And typically we only want it to run when we actually execute commands (which was so far ONLY when containers are up).

But during shell command on the host execution we ALSO want to trace with tcpdump. So code should stay as is.

=> I will create a PR for a new TCP Dump provider that is really system only and one that is scoped to the containers only. 
=> Please change code on this path (start_metric_providers) back to normal

lib/scenario_runner.py
```diff
@@ -2331,10 +2331,10 @@ def _start_metric_providers(self, allow_container=True, allow_other=True):
         self._custom_sleep(2)
 
         for metric_provider in self.__metric_providers:
-            if (metric_provider._metric_name.endswith('_container') or metric_provider._metric_name == 'network_connections_tcpdump_system' ) and not allow_container:
+            if metric_provider._metric_name.endswith('_container') and not allow_container:
```
ArneTR: => Please change code on this path (start_metric_providers) back to normal

lib/scenario_runner.py
```diff
@@ -2310,10 +2310,10 @@ def _start_metric_providers(self, allow_container=True, allow_other=True):
         # This includes tcpdump, which is only for debugging of the containers itself
         # If debugging of the tool itself is wanted tcpdump should be started adjacent to the tool and not inline
         for metric_provider in self.__metric_providers:
-            if (metric_provider._metric_name.endswith('_container') or metric_provider._metric_name == 'network_connections_tcpdump_system' ) and not allow_container:
+            if metric_provider._metric_name.endswith('_container') and not allow_container:
                 continue
 
-            if not metric_provider._metric_name.endswith('_container') and metric_provider._metric_name != 'network_connections_tcpdump_system' and not allow_other:
+            if not metric_provider._metric_name.endswith('_container') and not allow_other:
```
ArneTR: => Please change code on this path (start_metric_providers) back to normal

lib/scenario_runner.py
```diff
@@ -2331,10 +2331,10 @@ def _start_metric_providers(self, allow_container=True, allow_other=True):
         self._custom_sleep(2)
 
         for metric_provider in self.__metric_providers:
-            if (metric_provider._metric_name.endswith('_container') or metric_provider._metric_name == 'network_connections_tcpdump_system' ) and not allow_container:
+            if metric_provider._metric_name.endswith('_container') and not allow_container:
                 continue
 
-            if not metric_provider._metric_name.endswith('_container') and metric_provider._metric_name != 'network_connections_tcpdump_system' and not allow_other:
+            if not metric_provider._metric_name.endswith('_container') and not allow_other:
```
ArneTR: => Please change code on this path (start_metric_providers) back to normal

shell.py
```diff
@@ -24,6 +24,7 @@
 from lib.terminal_colors import TerminalColors
 
 SHELL_FLOW_NAME = 'Shell Command'
+LOCAL_USER_ID = 1
```
ArneTR: I know coderabbit recommended this, but I disagree.

I typically want to run shell comands NOT on our public user, but on a hidden user in the clsuter (say user 22 or so).

So fixing this breaks usability for me.

I think that user_id=1 is a sane default though as for many people user 1 will be the only user. so having that as the argv default i feel is good

shell.py
```diff
@@ -148,7 +146,11 @@ def parse_args():
     if args.shell_executable is None:
         args.shell_executable = 'powershell' if host_platform.is_windows() else 'bash'
 
-    args.command = " ".join(command_parts)
+    if len(command_parts) == 1:
```
ArneTR, line 151: Why this fix? What does it address?

shell.py
```diff
@@ -68,8 +69,6 @@ def __init__(self, *, shell_command, shell_executable, **kwargs):
         )
         self.__shell_command = shell_command
         self.__shell_executable = shell_executable
-        self._arguments['shell_command'] = shell_command
```
ArneTR: Why this removal? What use case does it address?

shell.py
```diff
@@ -165,23 +167,28 @@ def main():
...
     runner = None
 
-    if args.config_override is not None:
-        if not args.config_override.endswith(".yml"):
-            raise ValueError("Config override file must be a yml file")
-        GlobalConfig(config_location=args.config_override)
-
     print(TerminalColors.WARNING, "\n####################################################################################")
     print("Please use the docker version for exact measurements on the cluster!")
+    print("The command is stored with the run. Use --dev-no-save if it contains secrets!")
```
ArneTR, line 174: Why not filter it? Can it not be filtered?

shell.py
```diff
@@ -165,23 +167,28 @@ def main():
...
     print("####################################################################################\n", TerminalColors.ENDC)
 
 
+    if args.config_override is not None:
```
ArneTR, line 178: Why this move and expansion? What does it address? Without a comment or at least a **COMMIT MESSAGE** !!!!!!  :)  it is unclear to me what you are doing here

ArneTR (commented):
After our call now the try for another review.

First my general remarks:
- I pasted the Coderabbit comments as is. I typically decide if they really should come in or not. I did not add my opinion on it as it was not my PR and for some of the remarks I was not sure what the AI wanted. Now seeing your fixes though I have different opinions on them and would not have implemented some of them and plain rejected them. Feel free to also do that in the future. I find CodeRabbit remarks good, but not always helpful. My details you find annotated  on the code lines in the Review.
- The PR got so long that in many instances I am unsure which use case / change for request you are addressing. Please write shorter commits and more clear commit messages 🥺

shell.py
ArneTR: It says outdated, but actually for a fresh removal. Weird that it shows that if I comment on the left side of a diff ...

lib/scenario_runner.py
ribalba: https://github.com/green-coding-solutions/green-metrics-tool/pull/1585/changes/7694b2b73535d9845cceb7b690d4ed958970ae5e

shell.py
ribalba: Reverted https://github.com/green-coding-solutions/green-metrics-tool/pull/1585/changes/c76ce68cb3ce159749598511841903c421c940be

lib/host_platform.py
ribalba: https://github.com/green-coding-solutions/green-metrics-tool/pull/1585/changes/d86a65c48e0dba201f023306a3a8ae670e700055

done

shell.py
ribalba: This fixes:

136-142: 🎯 Functional Correctness | 🟡 Minor | ⚡ Quick win

Preserve command argument boundaries.

" ".join(command_parts) is not verbatim. For example, ./shell.py -- python -c 'print("hello world")' becomes python -c print("hello world"), which can change the script passed to the inner shell. Define one unambiguous command-input form and preserve quoting boundaries for arguments containing spaces.

-----
Is the case that this fixes
```
>>> shlex.join(['echo "Hello" && sleep 10'])
'\'echo "Hello" && sleep 10\''
>>> 
```

shell.py
ribalba: How do we know what a secret is? For example
```
$ ./login_script PASSWORD
```
we have no idea that password is what should filter?

shell.py
ribalba: This fixes this:

168-171: 🎯 Functional Correctness | 🟡 Minor | ⚡ Quick win

Route config override failures through the CLI error handler.

A missing, unreadable, malformed, or incorrectly suffixed config file raises before the try block. The CLI then prints an uncaught traceback instead of using _report_shell_error and the normal exit-code path. Move config validation and loading inside the try block, or handle these errors explicitly.

shell.py
ribalba: This was because of:

71-72: 🔒 Security & Privacy | 🟠 Major | ⚡ Quick win

Sensitive Data Exposure (CWE-312): Cleartext Storage of Sensitive Information

Reachability: External

Do not persist the raw shell command in run metadata.

_arguments is serialized into runs.runner_arguments, and filter_sensitive_data() does not redact arbitrary command contents. Remove shell_command from _arguments and use a constant or redacted value for persisted usage-scenario metadata.

but now that I have thought about this in more detail I would actually want to leave it in and accept the leak

ribalba:
@ArneTR ready for the next round.

shell.py
ArneTR: I am also OK with this. If the user wants to supply something secret is has to go through the usage_scenario_variables route. _arguments are passed on saving through filter_sensitive_data() by default currently So this is enough I feel

shell.py
ArneTR: see above. It should be filtered when supplied as usage_scenario_variable

ArneTR:
Alright, I think this PR can now be merged in 😊

In the comments I made a remark, that I feel it would be helpful to have the feature to use secrets in the command, which are used there as usage_scenario_variables and then later replaced.

Example: shell.py --var GMT_SECRET_VAR_SUPER__=TOTALLY_SECRET -- python3 do_something.py -n 1 -x __GMT_SECRET_VAR_SUPER__

This way we could simply skip saving the shell_arguments as they anywhere appear in the usage scenario later pasted in. And then handle the secret replacement in the usage scenario as usual.

Having said that: I believe this is an additional feature for a separate PR. If you see this the same way please merge this in!

2. I checked the tests. They are fine. The error is from the still open bug related to https://github.com/green-coding-solutions/green-metrics-tool/pull/1825

PR #1604, opened 2026-03-19
Makes some tests run in parrallel

ribalba:
I played around a little with ChatGPT to speed up the tests. It's not great but shaves off a little. Also it doesn't add to much to the complexity. The problem I see is that we need to add new test files to the mapping. 

@Arne what do you think?


ArneTR:
Closing this as it will be replaced with new implementation

PR #1618, opened 2026-03-27
Add Windows RAPL metric provider via ScaphandreDrv and cmd.exe bridge

MaximilianJahns:
i create the folder structur /metric_providers/cpu/energy/rapl/scaphandre/component and insert all the requested files.
i modified the config.example with the proivder. if the solution works i guess it would be nice to put the provider inside a windows section and out of the common / component section.

## Summary

This PR adds a Windows RAPL metric provider for measuring CPU energy consumption via the ScaphandreDrv kernel driver. The implementation uses a cmd.exe bridge to invoke the RAPL reader from WSL2.

## Changes

**Configuration**
- Updated `config.yml.example` with commented configuration for `CpuEnergyRaplScaphandreComponentProvider` including sampling rate and RAPL executable path

**Core Implementation**
- `provider.py`: New `CpuEnergyRaplScaphandreComponentProvider` class that:
  - Manages configurable sampling rate, rapl_reader_exe path, and disabled domains
  - Implements `check_system()` to verify driver availability
  - Extends metric invocation with executable path and domain parameters
  - Reports energy in microjoules (uJ)

**Windows C Binary**
- `rapl_reader_cli.c`: Windows native CLI tool that:
  - Reads RAPL MSR energy registers directly via ScaphandreDrv IOCTL
  - Supports domain filtering (CPU, DRAM, GPU, PSYS) with auto-detection
  - Uses monotonic performance counter timestamps with microsecond precision
  - Computes per-interval energy deltas with 32-bit wrap handling
  - Outputs timestamped energy data in GMT format

**Build & Bridge Scripts**
- `build_and_deploy.bat`: Windows batch script to compile and deploy `rapl_reader.exe` to `C:\rapl\`
- `metric-provider-binary`: Bash wrapper script that invokes the Windows RAPL reader via `cmd.exe` with options for sampling rate, domain exclusion, and driver checks

**Tests**
- Added `tests/usage_the_test.yml`: CPU and RAM stress test scenario using `stress-ng`

## Notes

Per PR discussion, the metric-provider-binary script was initially missing from the commit but has been added. The author acknowledged that calling bash directly via `subprocess.run` would be preferable and indicated intent to implement that improvement. The folder structure placement in the Windows-specific section (rather than common/component) was noted as a potential future refinement.


metric_providers/cpu/energy/rapl/scaphandre/component/__init__.py
ArneTR, line 1: Do we really need this? An empty __init__.py is only needed in Python < 3.3 to my undertstanding, or?

metric_providers/cpu/energy/rapl/scaphandre/component/build_and_deploy.bat
```diff
@@ -0,0 +1,65 @@
...
+    mkdir "C:\rapl\"
+    echo [OK] Created C:\rapl\
+) else (
+    echo [OK] C:\rapl\ already exists.
+)
+echo.
+
+:: ── Deploy ────────────────────────────────────────────────────
+echo [3/3] Copying rapl_reader.exe to C:\rapl\ ...
+copy /Y rapl_reader.exe "C:\rapl\rapl_reader.exe" >nul
```
ArneTR, line 48: Would it not make more sense to have the rapl_reader, which is very custom to the GMT, to be in this directory and not in C:/rapl ?

This would also save the user from some configuration overhead where the filename has to be supplied

metric_providers/cpu/energy/rapl/scaphandre/component/provider.py
```diff
@@ -0,0 +1,53 @@
+import os
+import subprocess
+from metric_providers.base import BaseMetricProvider, MetricProviderConfigurationError
+
+
+class CpuEnergyRaplMsrWindowsProvider(BaseMetricProvider):
+
+    def __init__(self, sampling_rate, folder, skip_check=False, rapl_reader_exe=None):
+
+        if rapl_reader_exe is None:
```
ArneTR: See my note above about fixing the path to a defined location

metric_providers/cpu/energy/rapl/scaphandre/component/provider.py
```diff
@@ -0,0 +1,53 @@
...
+            sampling_rate=sampling_rate,
+            unit='uJ',
+            current_dir=os.path.dirname(os.path.abspath(__file__)),
+            skip_check=skip_check,
+            folder=folder,
+        )
+
+    def check_system(self, check_command=None, check_error_message=None, check_parallel_provider=False):
+        # Check that rapl_reader.exe exists and driver is accessible
+        binary = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'metric-provider-binary')
```
ArneTR: what is the metric-provider-binary here?

metric_providers/cpu/energy/rapl/scaphandre/component/provider.py
```diff
@@ -0,0 +1,53 @@
...
+                f"  sc.exe start ScaphandreDrv"
+            )
+        return True
+
+    def _add_extra_switches(self, call_string):
+        # Pass exe path to the shell script via -e flag
+        return f"{call_string} -e \"{self._rapl_reader_exe}\""
+
+    def _parse_metrics(self, df):
+        df['detail_name'] = df['detail_name'].fillna('cpu_package')
```
ArneTR: why fillna? would you not need to totally replace the column?

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
ArneTR, line 1: Would it be possible to add PSYS domain here? Does your laptop have this? Or do you only have access to a desktop machine?

config.yml.example
```diff
@@ -202,7 +202,13 @@ measurement:
         sampling_rate: 99
     #--- Architecture - Common
     common:
-#      network.connections.proxy.container.provider.NetworkConnectionsProxyContainerProvider:
+#--- Windows RAPL - Only enable on Windows machines with WSL2
+#      cpu.energy.rapl.scaphandre.component.provider.CpuEnergyRaplScaphandreComponentProvider:
+#        sampling_rate: 100
```
ArneTR: why 100 if 99 is the default for all our sampling_rate entries?

Maybe background why we use 99 ms: To avoid lockstep sampling. Many tools have internal tick rates of 100 ms. To not sample on their internal step (where some interesting logic & thus energy spikes might occur) we have this small drift to always sample at non matching intervals

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -0,0 +1,139 @@
...
+        uint64_t ts = now_us();
+
+        if (curr.valid && !first) {
+            long long pkg_uj  = delta_uj(curr.pkg_energy_j,  prev.pkg_energy_j,  curr.energy_unit);
+            long long pp0_uj  = delta_uj(curr.pp0_energy_j,  prev.pp0_energy_j,  curr.energy_unit);
+            long long pp1_uj  = delta_uj(curr.pp1_energy_j,  prev.pp1_energy_j,  curr.energy_unit);
+            long long dram_uj = delta_uj(curr.dram_energy_j, prev.dram_energy_j, curr.energy_unit);
+
+            /* Skip values <= 1 to avoid GMT resolution underflow and dont use same timestamp */              /* cpu_gpu immer ausgeben, auch kleine Werte */
+            if (pkg_uj  > 1) printf("%llu %lld cpu_package\n", (unsigned long long)ts,     pkg_uj);
```
ArneTR: why is 1 or 0 not an acceptable value?

In the code for our other RAPL providers we only skip negative entries which result from overflow. see: https://github.com/green-coding-solutions/green-metrics-tool/blob/6a8c3c3e37fbdd79f1aaf591233e7c4b627dee59/metric_providers/cpu/energy/rapl/msr/component/source.c#L478

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -0,0 +1,139 @@
...
+static long long delta_uj(double curr, double prev, double eu) {
+    double d = curr - prev;
+    if (d < 0) d += (double)(1ULL << 32) * eu;
+    return (long long)(d * 1000000.0);
+}
+
+int main(int argc, char *argv[]) {
+    int interval_ms = 100, check_mode = 0;
+    for (int i = 1; i < argc; i++) {
+        if (!strcmp(argv[i], "-i") && i+1 < argc) interval_ms = atoi(argv[++i]);
```
ArneTR: we do not use atoi anymore because of a lot of edge cases. Have a look at https://github.com/green-coding-solutions/green-metrics-tool/blob/6a8c3c3e37fbdd79f1aaf591233e7c4b627dee59/lib/c/gmt-lib.c#L29

Maybe you can include the lib also in your project. If the linker complains you need to copy the function over to the windows code

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -0,0 +1,139 @@
...
+            if (pkg_uj  > 1) printf("%llu %lld cpu_package\n", (unsigned long long)ts,     pkg_uj);
+            if (pp0_uj  > 1) printf("%llu %lld cpu_cores\n",   (unsigned long long)ts + 1, pp0_uj);
+            printf("%llu %lld cpu_gpu\n", (unsigned long long)ts + 2, pp1_uj > 0 ? pp1_uj : 2);
+            if (dram_uj > 1) printf("%llu %lld dram\n",        (unsigned long long)ts + 3, dram_uj);
+        }
+
+        if (curr.valid) { prev = curr; first = 0; }
+
+        DWORD elapsed = GetTickCount() - t_start;
+        if ((DWORD)interval_ms > elapsed) Sleep(interval_ms - elapsed);
```
ArneTR: that is different from our implementation of other sleeps where we just do fixed sleeps. Why did you choose this?

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -0,0 +1,139 @@
...
+    DWORD bytes = 0;
+    BOOL ok = DeviceIoControl(hDev, IOCTL_READ_MSR, &req, sizeof(req),
+                               out, sizeof(*out), &bytes, NULL);
+    return (ok && bytes >= sizeof(uint64_t)) ? 0 : -1;
+}
+
+static rapl_sample_t read_rapl(HANDLE hDev, uint32_t cpu) {
+    rapl_sample_t s = {0};
+    uint64_t raw = 0;
+    if (read_msr(hDev, MSR_RAPL_POWER_UNIT, cpu, &raw) != 0)
```
ArneTR, line 176: The power unit should not change and should just be read one time on start

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -0,0 +1,139 @@
...
+        if (read_msr(hDev, MSR_AMD_RAPL_POWER_UNIT, cpu, &raw) != 0) return s;
+    s.energy_unit = 1.0 / (double)(1ULL << ((raw >> 8) & 0x1F));
+    if (read_msr(hDev, MSR_PKG_ENERGY_STATUS,  cpu, &raw) == 0) s.pkg_energy_j  = (raw & 0xFFFFFFFF) * s.energy_unit;
```
ArneTR: I think it is quite some overhead to read all MSRs if we in the end just want to output one.

I would keep the mode as default, but allow restriciting to only one particular MSR via a switch

ArneTR (changes requested):
The code is niceley readable and well structured. I could easily follow :)

Stuff that was linked to a commit line I commented inline.

Here follows my general questions:
- You mentioned that `GetTickCount() ` Only has a 15ms resolution. Can it easily be moved to something that has native microsecond resolution?
  - We are actually making quite some work in our providers to have a monotonic clock, that we are mapping to a wall-time. WOuld be nice to have something identical in windows to not be affected by DST and leap seconds. See: https://github.com/green-coding-solutions/green-metrics-tool/blob/6a8c3c3e37fbdd79f1aaf591233e7c4b627dee59/lib/c/gmt-lib.c#L53
- In https://github.com/MaximilianJahns/gmt-windows-rapl you mention that you put 2 uJ when value is too small. I cannot see that anymore. Has it been removed?
- You mention that "stdout-Weiterleitung für viele Metriken ungeeignet" and want to checked named pipes. Is this about the performance? Or about having merged metrics in the output? For the latter we actually just start multiple providers and restrict each to only one output metric and run fine with it ...

metric_providers/cpu/energy/rapl/scaphandre/component/__init__.py
MaximilianJahns: nope you are right, i will take it away.

metric_providers/cpu/energy/rapl/scaphandre/component/build_and_deploy.bat
MaximilianJahns: then we have to deal with it the other way around that we need some enviromen variable to locate the gmt folder. we could put it in the install script. to check if windows bare metal > create enviroment variable for gmt provider folder to locate it with the bash script. i will keep it for now like this integrate the rest of the comments and try to get a solution for it.

metric_providers/cpu/energy/rapl/scaphandre/component/provider.py
MaximilianJahns: the metric provider binary is just a bash script to activate the reader on windows maschiene

metric_providers/cpu/energy/rapl/scaphandre/component/provider.py
MaximilianJahns: i thougth it is helpful for NaN but i guess there wont be NaNs so i take it out

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: /metric_providers/base.py

    def _check_resolution_underflow(self, df):
        if self._unit in ['mJ', 'uJ', 'Hz', 'us']:
            if (df['value'] <= 1).any():
                raise ValueError(f"Data from metric provider {self._metric_name} is running into a resolution underflow. Values are <= 1 {self._unit}")

for dram i created a check if dram is supported (not 0  otherwise it is not take into account) but for Cpu_Gpu the support test is runnig througth, but sometimes an intervall is 0 uj and i get an error. not shure how to deal with it other then to make an exception for the gpu-cpu if 0 = 2uk.

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: ok i changed it

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: i thougth it is more constant with the delay, one solution that is faster then GetTickCount() is QueryPerformanceCounter() but a fixed sleep is also fine with me, so i changed that

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: ok

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: ok

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -99,41 +294,79 @@ int main(int argc, char *argv[]) {
...
+             * Only skip negative values (overflow edge cases).
+             * Matches GMT's native RAPL provider behavior.
+             * Use +1us timestamp offsets per domain so GMT can
+             * calculate per-domain sampling rates correctly.
+             */
+            if ((domains & DOMAIN_PKG)   && pkg_uj  >= 0)
+                printf("%llu %lld cpu_package\n", (unsigned long long)ts,     pkg_uj);
+            if ((domains & DOMAIN_CORES) && pp0_uj  >= 0)
+                printf("%llu %lld cpu_cores\n",   (unsigned long long)ts + 1, pp0_uj);
+            // * it is given a vailue for 0 of 2uk, otherwise he have to chnage the provider logik.
```
ArneTR: what does this comment mean? Is there a typo in there? I do not understand the "for 0 of 2uk"

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -99,41 +294,79 @@ int main(int argc, char *argv[]) {
...
+            /*
+             * Only skip negative values (overflow edge cases).
+             * Matches GMT's native RAPL provider behavior.
+             * Use +1us timestamp offsets per domain so GMT can
+             * calculate per-domain sampling rates correctly.
+             */
+            if ((domains & DOMAIN_PKG)   && pkg_uj  >= 0)
+                printf("%llu %lld cpu_package\n", (unsigned long long)ts,     pkg_uj);
+            if ((domains & DOMAIN_CORES) && pp0_uj  >= 0)
+                printf("%llu %lld cpu_cores\n",   (unsigned long long)ts + 1, pp0_uj);
```
ArneTR: why are you adding time to the timestamp?

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -99,41 +294,79 @@ int main(int argc, char *argv[]) {
...
-            long long pp0_uj  = delta_uj(curr.pp0_energy_j,  prev.pp0_energy_j,  curr.energy_unit);
-            long long pp1_uj  = delta_uj(curr.pp1_energy_j,  prev.pp1_energy_j,  curr.energy_unit);
-            long long dram_uj = delta_uj(curr.dram_energy_j, prev.dram_energy_j, curr.energy_unit);
-
-            /* Skip values <= 1 to avoid GMT resolution underflow and dont use same timestamp */              /* cpu_gpu immer ausgeben, auch kleine Werte */
-            if (pkg_uj  > 1) printf("%llu %lld cpu_package\n", (unsigned long long)ts,     pkg_uj);
-            if (pp0_uj  > 1) printf("%llu %lld cpu_cores\n",   (unsigned long long)ts + 1, pp0_uj);
-            printf("%llu %lld cpu_gpu\n", (unsigned long long)ts + 2, pp1_uj > 0 ? pp1_uj : 2);
-            if (dram_uj > 1) printf("%llu %lld dram\n",        (unsigned long long)ts + 3, dram_uj);
+            long long pkg_uj  = delta_uj(curr.pkg_energy_j,  prev.pkg_energy_j,  energy_unit);
```
ArneTR, line 451: You are still fetching all MSRs but only outputting them separately.

How is that reducing the latency? Or do you plan to address this a different way?

ArneTR (changes requested):
Thanks for updates! I added some remarks on the updates.

Still open is:
- Can PSYS be added
- You mention some problem with the GPU. I do not quite understand it. if it is not present it will return 0 values which will lead to an underflow. THis means the provider cannot be used. There is no sane way to fix this in the C code directly and atm this is considred domain knowledge. Does that help? If not please follow up

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: Thanks for veviewing the code already

The GPU domain PP1 is present and returns values on this CPU, but occasionally outputs 0 uJ at 99ms intervals when the iGPU is idle. 

 first If we output 0 uJ → GMT raises a resolution underflow error (`value <= 1 uJ`)
second . If we skip the 0 uJ samples → the time series has gaps → GMT raises a sampling rate error (~198ms instead of ~99ms, exceeding the 20% tolerance)

The current workaround is to substitute 0 uJ with 2 uJ to keep the time series gap-free.

My original approach was to detect inactive domains at startup by taking 5 samples × 100ms and disabling any domain that consistently returns 0. This works for DRAM (which is permanently 0 on this CPU). For GPU however, the 0 values occur only occasionally during measurement, not consistently at startup - so the startup check passes but 0 values still appear later.

Given that GMT has a 2-second startup delay, I could extend the startup check to 10-15 samples × 100ms to get a more reliable test. But even then, a GPU that is active at startup might go idle during measurement and produce 0 uJ values. there where around 15% samples missing if i skip the 0 vlaues compared to the rest of the data. 
not sure what is the best approach or to skip the cpu_gpu firstly.

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: The +1 microsecond offsets are just  workaround.

GMT calculates the sampling rate using time differences per domain. If two domains have the exact same timestamp, the difference is 0, which causes an error.

Since all domains are read at the same time, they would normally have identical timestamps. The tiny offsets make them slightly different so GMT works correctly.

i could also call the msr fpor eacht value seperatly to get a unique timestamp. i will implement it!

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: This is already addressed in the current commit. The  read_rapl()`function only reads MSRs for the requested domains


if (domains & DOMAIN_PKG)
    if (read_msr(hDev, MSR_PKG_ENERGY_STATUS, cpu, &raw) == 0)
        s.pkg_energy_j = (raw & 0xFFFFFFFF) * energy_unit;

if (domains & DOMAIN_DRAM)  // skipped if DRAM was auto-disabled
    if (read_msr(hDev, MSR_DRAM_ENERGY_STATUS, cpu, &raw) == 0)
        


So when DRAM is disabled at startup (auto-detection), its MSR is never read during the measurement loop. Same applies when using -d cpu_package - only the MSR package should be read.

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
ArneTR: GMT can split by detail_name and will evaluate timestamps separately. I am not sure why this is happening then as a lot of providers like the NetworkIO System have this kind of behaviour ... Maybe have a look at it's implementation or we call next week for details

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
ArneTR: We can just disable the check for the GPU then.

This however means we need the C-Code to work with switches that we can separate GPU, DRAM and CPU polling into separate threads.

This is also why I argued for having the polling of the MSRs optional per domain and not poll all at the same time as it would create overhead when we thread-split them.

Can also be discussed next week in a call if unclear

ArneTR:
Hey @MaximilianJahns 

just some perspective here: we brought in a new tool to AI review commits.

Don't be startled by the volume of comments. If you feel they are non-sensical to you ignore them for now :) But if you feel it pointed out something legit please push a fix.

@ribalba will follow up with a first look today and as it looks we will touch base on wednesday in a call

MaximilianJahns:
Hi Arne. yes i saw lots of comments created by the bot. i will check them aswell continuing with the code. i Think its helpful! any comment even by a bot is fine ;)

ribalba:
Just a heads up that I am still in the process of getting a windows machine.

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
MaximilianJahns: the code is done for all metrics, i used an x flag from the provider not shure if you like that. but in the cli the metrics are seperate domians. i used psysy for the Benchmark of Blender 3d. Cpu_Gpu is not supportet and dram nether 

#define DOMAIN_PKG   (1 << 0)  // 00001
#define DOMAIN_CORES (1 << 1)  // 00010
#define DOMAIN_GPU   (1 << 2)  // 00100
#define DOMAIN_DRAM  (1 << 3)  // 01000
#define DOMAIN_PSYS  (1 << 4)  // 10000
#define DOMAIN_ALL   (DOMAIN_PKG | DOMAIN_CORES | DOMAIN_GPU | DOMAIN_DRAM | DOMAIN_PSYS)
// DOMAIN_ALL = 11111 = 31

I dont have a dekstop maschiene. but i have an old very gigantic Hp Zbook, that could work. i will do the test with it today

metric_providers/cpu/energy/rapl/scaphandre/component/rapl_reader_cli.c
```diff
@@ -0,0 +1,341 @@
...
+            s.pp0_energy_j = (raw & 0xFFFFFFFF) * energy_unit;
+    if (domains & DOMAIN_GPU)
+        if (read_msr(hDev, MSR_PP1_ENERGY_STATUS, cpu, &raw) == 0)
+            s.pp1_energy_j = (raw & 0xFFFFFFFF) * energy_unit;
+    if (domains & DOMAIN_PSYS)
+        if (read_msr(hDev, MSR_PLATFORM_ENERGY_STATUS, cpu, &raw) == 0)
+            s.psys_energy_j = (raw & 0xFFFFFFFF) * energy_unit;
+
+    s.valid = 1;
+    return s;
```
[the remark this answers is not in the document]
MaximilianJahns: i will fix it per domain

ribalba:
I have copied a lot of code from this into 

https://github.com/green-coding-solutions/green-metrics-tool/pull/1648

I think we can close this PR and follow up in the other PR

PR #1644, opened 2026-04-16
(test-fix): Certain test ordering could result in wrong Docker image being used

ArneTR:


ArneTR:
@ribalba The error seen with the volume loading resulted on a specific ordering which did use a wrong Docker image.

This test forces the two tests that use that image to always rebuild.

Time impact is +10s

ArneTR:
Pipeline failed with HTTP errors in dependencies ... Tests for the functionality seems ok.

PR #1648, opened 2026-04-20
makes the GMT run under windows without WSL

ribalba:
This is a working version of the GMT under windows. There are a few small things that still need fixing but like this it is possible to run the GMT under windows without the WSL system. You need to install python from the windows app store and git.

Tomorrow I will port the code from @MaximilianJahns so we have a metric provider. Also I am quite sure that I can get a simple cpu utilization provider working with `psutils`.  

The next step we need to decide is how we want to run windows apps. There is no real reason we should benchmark docker containers under windows. So I think we will need to implement the `host` functionality. We will need to look into how we want to orchestrate windows apps. Maybe @MaximilianJahns knows more about this as they have a windows benchmarking lab at HTW. 

@ArneTR small ping


ArneTR:
Very interesting!

Let me know if I should review / test.

Also looping in @MaximilianJahns as possible reviewer once done

MaximilianJahns:
hello, yes i am keen to test. unfourtunally  i am not familiar with the benchmark lab for windows at the HTW, but i guess i can get in touch and get in to it. to have GMT native in windows would be way more lean, so hope it is runnig soon.

install_windows.ps1
```diff
@@ -0,0 +1,479 @@
...
+
+    if ((Split-Path -Leaf $python) -ieq "py.exe") {
+        return @($python, "-3")
+    }
+    return @($python)
+}
+
+function Invoke-Python {
+    param([string[]]$Arguments)
+    $command = $script:PythonCommand[0]
```
ArneTR: This seems to be the culprit. 

The variable $script:PythonCommand contains: "C:\Users\greencoding\AppData\Local\Microsoft\WindowsApp\python.exe"

So getting [0] of that only results in C ...

ArneTR (changes requested):
Cannot get it to work.

It errors with:

& : Die Bennenung "C" wurde nicht als Name eines Cmdlet, einer Funktion oder einer Sktipdatei oder eines ausführbaren Programmes erkannt.
Überprüfen Sie die Schreibweise ...
In C:\Users\greencoding\Green-metrics-tool\install_windows.ps1:131 Zeichen 7

I have tried to fix it myself but failed. I made a comment highlighting the correct line for you to drill down

ribalba:
@MaximilianJahns will test on his setup and see if he can get it to work

MaximilianJahns:
i am running into some issues using the runner.py but the install prcoess was smothly.   what test do you use for testing the Windows GMT version do you have a standard test?

ribalba:
I just tested some of the example applications. https://github.com/green-coding-solutions/example-applications

What are the issues you are facing?

MaximilianJahns:
**Root cause of `TypeError: can't concat NoneType to bytes`**

`set_nonblocking()` in `host_platform.py` calls `os.set_blocking()` – which does not exist on Windows. Because of the early `return` in the function, it silently does nothing and the stream stays in blocking mode.

Python's `subprocess` internally starts a `_readerthread` to continuously read from the pipe:
```python
# From Python's subprocess.py (simplified):
def _readerthread(fh, buffer):
    buffer.append(fh.read())  # crashes here
```

we could try to use an different method for windows like Queue+Thread (works for Linux and Windows) python or PeekNamedPipe in C but then it seems to just work for Windows. shall i try it with the python version Queue?

ribalba:
I can not replicate this. Can you give me a dump of your shell please?

@MaximilianJahns I fixed a few little things. Can you please pull again and reinstall. 

then I run something like
```
py runner.py --uri C:\Users\didi\example-applications\ --filename stress\usage_scenario.yml --dev-no-sleep --dev-no-system-checks
```

MaximilianJahns:
hey thats looking good! yes i will do! okay the runner is quit similar i hope to get it runnig aswell

MaximilianJahns:
it is running througth! i had to build the binary again forgot about that and had some user related Problems... but now the values are shown in frontend. the missing data issue is sitll there  but i guess it is a different story.

ribalba:
What missing data issue?

MaximilianJahns:
frontend

frontend2
 
that is the last test run on windwos, but it is just a displaying issue i guess, so the phases are shown in the specific columns at the top

ArneTR:
The "missing data" displays comes when no machine energy provider is connected.

You can solve this once you have either RAPL PSYS (does this exist on your machine?) or once we have a CPU Utilization provider and can run one of the "common" providers that estimate the machine energy (SDIA or XGBoost)

In short: This is not a bug, but a missing configured provider

MaximilianJahns:
ahh okay now i got it, thanks

MaximilianJahns:
okay i just replicated the issued from yesterday, it happens if the test is detached true(i guess thats not quiete a typical testcase...) but even then it goes trougth and is just displayed as follow:  Traceback (most recent call last):
  File "C:\Users\jahns\AppData\Local\Programs\Python\Python313\Lib\threading.py", line 1044, in _bootstrap_inner
    self.run()
    ~~~~~~~~^^
  File "C:\Users\jahns\AppData\Local\Programs\Python\Python313\Lib\threading.py", line 995, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jahns\AppData\Local\Programs\Python\Python313\Lib\subprocess.py", line 1615, in _readerthread
    buffer.append(fh.read())
                  ~~~~~~~^^
  File "<frozen codecs>", line 324, in decode
TypeError: can't concat NoneType to bytes
  File "C:\Users\jahns\AppData\Local\Programs\Python\Python313\Lib\threading.py", line 1044, in _bootstrap_inner
    self.run()
    ~~~~~~~~^^
  File "C:\Users\jahns\AppData\Local\Programs\Python\Python313\Lib\threading.py", line 995, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jahns\AppData\Local\Programs\Python\Python313\Lib\subprocess.py", line 1615, in _readerthread
    buffer.append(fh.read())
                  ~~~~~~~^^
  File "<frozen codecs>", line 324, in decode
TypeError: can't concat NoneType to bytes

ribalba:
Can you share the usage_scenario and the runner call string please?

MaximilianJahns:
sure it is the same as you used: so its the usage szenario but just with detached: true ---
name: Stress Container One Core 5 Seconds

author: Arne Tarara <arne@green-coding.berlin>
description: A simple 5 seconds stress on the system with one core. Expected display result is a rectangle shaped load.

compose-file: !include compose.yml

services:
  gcb-alpine-stress:
    cpus: 1
    mem_limit: "300MB"



flow:
  - name: Stress
    container: gcb-alpine-stress
    commands:
# Alpine does not have stress, so we use stress-ng
# We need the -q flag because otherwise it will write debug to STDERR
      - type: console
        command: stress-ng -c 1 -t 5 -q
        note: Starting Stress
        detach: true
 python runner.py --uri C:\Users\jahns\Documents\CASO\GMT-WIndows\example-applications --filename stress\usage_scenario2.yml --dev-no-sleeps --dev-no-system-checks

lib/utils.py
```diff
@@ -113,6 +114,9 @@ def df_fill_mean(group):
 
 
 def get_network_interfaces(mode='all'):
+    if host_platform.is_windows():
```
MaximilianJahns, line 117: i wonder if it would be helpful to create an windows/common check in the config file similar to linux/mac

MaximilianJahns (commented):
I went through, its clean. 

A couple of small things:

The URI detection in runner.py might produce a slightly confusing error if someone passes a file path instead of a directory, or if the path doesn’t exist. Pretty minor UX thing.

the mentioned set_nonblocking() is just crashing if the kerneldriver is down aswell so that shouldnt be a bigger problem.  hope i found most

ribalba:
@ArneTR pleasy try on your setup again

ArneTR:
Works like a charm now 😊

Let's bring it in

lib/scenario_runner.py
```diff
@@ -178,10 +180,13 @@ def __init__(self,
         self._measurement_wait_time_dependencies = measurement_wait_time_dependencies
         self._last_measurement_duration = 0
         self._phase_padding = phase_padding
-        self._phase_padding_ms = max(
-            utils.get_metric_providers(config, self._disabled_metric_providers).values(),
-            key=lambda x: x.get('sampling_rate', 0) if x else 0
-        ).get('sampling_rate', 0)
+        configured_metric_providers = utils.get_metric_providers(config, self._disabled_metric_providers)
```
ArneTR, line 183: is this just a clean version, or did you change something?

lib/host_platform.py
```diff
@@ -0,0 +1,158 @@
...
+        if len(parts) == 2:
+            rows.append((parts[0], parts[1]))
+    return rows
+
+
+def remove_gmt_tmp_images():
+    try:
+        image_rows = _docker_image_rows()
+    except subprocess.CalledProcessError:
+        return
```
ArneTR, line 127: this return does not seem right to me ... this is windows related, is it? Why not return only on Windows and keep error on linux?

lib/host_platform.py
```diff
@@ -0,0 +1,158 @@
...
+    system = system_name().lower()
+    if system == 'darwin':
+        return 'macos'
+    if system == 'windows':
+        return 'windows'
+    return system
+
+
+def get_tmp_root():
+    return Path(tempfile.gettempdir()).resolve(strict=True)
```
ArneTR, line 31: Does this really do what we want?

Before we used /tmp

This will now return for example `/var/folders/n4/qq26rns52yv6lwrp3zk_nhcc0000gn/T`

A lot of our code assumes /tmp/green-metrics-tool as static path ...?

Furthermore we use the advantange that /tmp is mounted on tmpfs for quicker storing .. so i foresee problems wiht metric provider sample accuracy ...?

lib/host_platform.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def clear_file_system_caches():
+    if is_windows():
+        print('Skipping filesystem cache clearing on Windows')
```
ArneTR, line 36: Why? Does it not have clearing of caches?

This will be very hard to debug if we continue working on windows later ...

lib/system_checks.py
```diff
@@ -139,6 +148,9 @@ def check_swap_disabled(*_, **__):
         return True
 
 def check_suspend(*, run_duration):
+    if host_platform.is_windows():
```
ArneTR, line 151: Would be nice to have a warning / todo here for later that we need to implement this

lib/system_checks.py
```diff
@@ -161,6 +173,8 @@ def check_suspend(*, run_duration):
     return 'Entering' not in ps.stdout and 'suspend' not in ps.stdout
 
 def check_steal_time(*_, **__):
+    if host_platform.is_windows():
```
ArneTR, line 176: Would be nice to have a warning / todo here for later that we need to implement this

lib/system_checks.py
```diff
@@ -126,6 +133,8 @@ def check_tty_attached(*_, **__):
 # This text we compare with indicates that no swap is used
 #pylint: disable=no-else-return
 def check_swap_disabled(*_, **__):
+    if host_platform.is_windows():
```
ArneTR, line 136: Would be nice to have a warning / todo here for later that we need to implement this

lib/system_checks.py
```diff
@@ -51,10 +54,12 @@ def check_one_energy_and_scope_machine_provider(*_, **__):
     return len(energy_machine_providers) <= 1
 
 def check_tmpfs_mount(*_, **__):
+    if host_platform.is_windows():
+        return True
     return not any(partition.mountpoint == '/tmp' and partition.fstype != 'tmpfs' for partition in psutil.disk_partitions())
```
ArneTR, line 59: This check makes also no sense anymore with the new tmp folder ...

ArneTR (commented):
Sorry to bring this up again after merge, but I was a bit too eager.

I noticed I did not review the PR so far and have made some remarks we need to address

The /tmp problem is for me only on macOS but can be on linux also to my understanding depending on system configuration

lib/scenario_runner.py
ribalba: The problem was that this was failing when there were no metric providers configured. Which I was testing stuff with under windows. It should be functionally the same

lib/host_platform.py
ribalba: Very good catch. I just assumed that under Linux it would be `/tmp/`. Changed it to be `/tmp` for Mac and Linux.

lib/host_platform.py
ribalba: I didn't find a nice way of doing this. There are tools that can do this but I didn't want to ship with an external tool. Then there is a way with powershell but here you need to be Admin. So I decided to keep it clean. I have now implemented the shell version in the PR so we can discuss.

lib/host_platform.py
ribalba: Removed the catch so that it always fails. This was from when there is no Docker. But for now we should enforce docker as the usage scenarios need it.

lib/system_checks.py
ribalba: Changed so we keep the test

lib/system_checks.py
ribalba: Added a power shell version

lib/system_checks.py
ribalba: Added a power shell version

lib/system_checks.py
ribalba: There is no steal time under Windows

lib/utils.py
ribalba: What do you mean?

ribalba:
I addressed the feedback here https://github.com/green-coding-solutions/green-metrics-tool/compare/windows-fixes?expand=1

PR #1658, opened 2026-04-26
Decrypt only on request

ArneTR:
This PR fixes some deployment related quirks with the SSH private key / RSA encryption implementation.

In the old form it was not possible to have the server only configured with a RSA public key only. Both keys needed to be set as the user.py was decrypting the private on object instantiation.

This has been now changed to a lazy encryption only on actual read of the key (which happens only in the runner).

Furthermore a documentation was missing. The experiences from setting this up are documented in https://github.com/green-coding-solutions/documentation/issues/134

## Overview
Implements lazy decryption of SSH private keys so the server no longer requires the RSA private key at startup — the private key is decrypted only when actually read.

## Changes
- **lib/user.py**
  - Switches SSH private key handling to lazy decryption: stores the encrypted DB value on initialization and defers decryption until get_ssh_private_key() is called; caches a SecureVariable after first decrypt.
  - Adds __encrypted_ssh_private_key and __decrypted_ssh_private_key internals, _save_key_to_db helper, and refactors update_ssh_private_key to validate, normalize, encrypt, cache, and persist keys. Clearing/None handling updated.
  - to_dict() and has_ssh_private_key() updated to avoid exposing internal key material and to derive _has_ssh_private_key from the encrypted value.
- **lib/encryption.py**
  - Adds is_valid_openssh_private_key(data: str) -> bool which checks OpenSSH private key PEM boundaries and attempts parsing via serialization.load_ssh_private_key(..., backend=default_backend()).
  - No change to existing RSA PEM load/encrypt/decrypt flows; existing error handling retained.
- **.gitignore**
  - Added .rsa/ to ignored paths.
- **tests**
  - **tests/test_functions.py**: Adds OPENSSH_EXAMPLE_PRIVATE_KEY constant used across tests.
  - **tests/lib/test_user.py**: Refactors tests to use the shared OPENSSH_EXAMPLE_PRIVATE_KEY constant; strengthens assertions to ensure both encrypted and decrypted internal fields are cleared/not exposed; adds/updates tests covering validation, clearing, persistence, and that neither plaintext nor encrypted markers leak via User.__repr__ or to_dict().
  - **tests/frontend/test_frontend.py**: Extends test_settings_does_not_expose_ssh_private_key to assert internal encrypted and decrypted SSH key fields are not present in the settings endpoint response.

### DIFF
--- a/lib/user.py
+++ b/lib/user.py
@@ -149,6 +149,7 @@ class User():
             WHERE id = %s
             """, params=(encrypted_value, self._id, ))
 
+        self.__encrypted_ssh_private_key = encrypted_value
         self.__decrypted_ssh_private_key = SecureVariable(normalized_value) if normalized_value else None
 
     def can_change_setting(self, name):
--- a/lib/user.py
+++ b/lib/user.py
@@ -40,7 +40,7 @@ class User():
         del values['_id']
         values.pop('_User__encrypted_ssh_private_key', None)
         values.pop('_User__decrypted_ssh_private_key', None)
-        values['_has_ssh_private_key'] = bool(self.__encrypted_ssh_private_key) is not None
+        values['_has_ssh_private_key'] = bool(self.__encrypted_ssh_private_key)
         return values
 
     def __repr__(self):
--- a/lib/user.py
+++ b/lib/user.py
@@ -108,9 +108,7 @@ class User():
         return bool(self.__encrypted_ssh_private_key)
 
     def get_ssh_private_key(self):
-        if self.__decrypted_ssh_private_key:
-            return self.__decrypted_ssh_private_key.get_value()
-        elif self.has_ssh_private_key():
+        if self.__decrypted_ssh_private_key is None and self.has_ssh_private_key():
             if self.__encrypted_ssh_private_key.startswith(ENCRYPTED_VALUE_PREFIX):
                 decrypted_ssh_private_key = decrypt_data(self.__encrypted_ssh_private_key)
             else:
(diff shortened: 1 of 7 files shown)


ArneTR:
@coderabbitai review

lib/user.py
```diff
@@ -107,13 +105,20 @@ def change_setting(self, name, value):
...
 
     def get_ssh_private_key(self):
-        if self._ssh_private_key:
-            return self._ssh_private_key.get_value()
-        else:
+        if self.__decrypted_ssh_private_key is None and self.has_ssh_private_key():
+            if self.__encrypted_ssh_private_key.startswith(ENCRYPTED_VALUE_PREFIX):
+                decrypted_ssh_private_key = decrypt_data(self.__encrypted_ssh_private_key)
+            else:
+                decrypted_ssh_private_key = self.__encrypted_ssh_private_key or None
```
ribalba: When do we have this case?

ribalba (approved):
looks good

lib/user.py
ArneTR: I thought the same thing :)

But If I am not crazy this was your code. Thus I opted for leaving it like this.

Now that we are talking about it though I would strongly vote for kicking this out as I believe it is a dead code path in our setup

ribalba: I agree. This must be left over from the olden days when we still had the unencrypted functionaility

ArneTR:
@coderabbitai review

ribalba:
Did this test? Because `main` is somehow failing now

PR #1659, opened 2026-04-26
Changing import of optimzations to happen earlier for efficiency reasons

ArneTR:
Optimizations import was moved to a different location as it was happening inside of the loop.

This was apparently needed in an old implementation but is not anymore. The comment was misleading and was updated.

## Summary

Moves `optimization_providers.base.import_reporters()` from inside the per-file loop to a one-time execution before the loop, conditioned on the same checks (`--skip-optimizations`, `--dev-no-save`, `--dev-no-metrics`). The `run_reporters()` call remains in the loop where it executes for each file. This avoids redundant imports when processing multiple scenario files in a single run.

The misleading comment explaining the prior placement has been clarified to indicate why the import still cannot occur at the top of the file: the configuration file may be overridden via command-line arguments.

**Files changed:** runner.py (+6/-3)


ArneTR:
@coderabbitai review

ArneTR:
@coderabbitai review

PR #1664, opened 2026-04-28
wrong value was saved

ribalba:


ribalba:
Waiting if this will fix the tests

ArneTR:
Interesting. The last change was done by me and seen as so minor (bc just reordering) that I merged after.

Bad day yesterday :/

PR #1665, opened 2026-04-29
Makes the get_private_key return a secure variable per default.

ribalba:
Makes it a little more secure as the default is a `SecureVariable` like this we avoid leaking it in the future per accident

## Overview
Refactors `User.get_ssh_private_key()` to return a `SecureVariable` wrapper by default rather than an unwrapped string, ensuring SSH private keys are handled as secure values throughout the codebase and reducing the risk of accidental leakage in logs or error messages.

## Changes

**lib/user.py**
- Modified `get_ssh_private_key()` to return `SecureVariable | None` instead of extracting and returning the raw string value
- Removed the `.get_value()` call that previously unwrapped the SecureVariable

**lib/job/run.py**
- Simplified `RunJob._process()` to pass the `SecureVariable` directly from `user.get_ssh_private_key()` to `ScenarioRunner`
- Removed prior conditional logic and explicit SecureVariable wrapping that handled None cases
- Removed unused `SecureVariable` import

**tests/api/test_api_base.py**
- Updated `test_can_update_ssh_private_key_setting` to expect `SecureVariable` return type
- Added verification of wrapped value using `.get_value()`
- Added explicit `SecureVariable` import

**tests/lib/test_user.py**
- Updated two SSH key tests to verify returned objects are `SecureVariable` instances
- Changed assertions to validate decrypted content via `.get_value()` instead of direct string comparison
- Maintained existing checks for encrypted storage and string representation security

### DIFF
--- a/lib/job/run.py
+++ b/lib/job/run.py
@@ -5,7 +5,6 @@
 import sys
 import faulthandler
 
-from lib.secure_variable import SecureVariable
 faulthandler.enable(file=sys.__stderr__)  # will catch segfaults and write to stderr
 
 import os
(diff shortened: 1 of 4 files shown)


ArneTR:
LGTM

@coderabbitai review

PR #1688, opened 2026-05-18
Adds more features to the windows part and fixes tmp path

ribalba:


ArneTR:
Looks all very reasonable to me.

My vote is to merge it in now, so we have the /tmp folder error for macOS out of the way.

Are the Windows functionalities currently covered by a test?
Or differently asked: Do we have a testing harness for Windows at the moment?

If not I see two possibilities:
- Creating one
- Asking Maximilian to test the features manually

ribalba:
No Windows has no tests to date. I looked into it for a few minutes and it shouldn't be too hard to port. This is also something Maximilan could do IMHO

ArneTR:
ok. can you make then two tasks for Maximilian in the project board (manual testing, porting the test suite to windows) and merge this in

PR #1697, opened 2026-05-26
Fixing tests

ribalba:
- [ ]  Still need to check if this also fixes the front end display errors

## Overview
Refactors CO2 intensity derivation to generate carbon metrics dynamically (replacing "_energy_" with "_carbon_"), removes a now-unused module constant, adjusts Elephant provider request parameters, and updates tests to match the new behavior and stricter error handling.

## Key changes
- lib/post_metric_providers/calculate_co2_intensity.py
  - Removed module-level DERIVED_METRIC constant.
  - Now derives metric names per energy metric by replacing "_energy_" → "_carbon_" and inserts derived metrics with unit "ugCO2e".
  - Builds derived metric detail_name as "{energy_detail_name}_{carbon_metric}_{carbon_detail_name}" and writes values via DB.copy_from.

- metric_providers/carbon/intensity/elephant/machine/provider.py
  - Removed the `update: "true"` request parameter when querying /carbon-intensity/history.
  - Error handling unchanged but tests updated to reflect parameter removal.

- Tests
  - tests/lib/test_phase_stats.py: Calls calculate_co2_intensity(run_id) before building phase stats; updated expected derived metric names/units and loosened one numerical assertion to pytest.approx.
  - tests/metric_providers/test_electricitymaps_provider.py: Removed pandas import; updated error-path tests to expect RuntimeError (status code/message) instead of returning empty DataFrame and checking _error_string.
  - tests/metric_providers/test_elephant_providers.py: Removed unused pandas import; rewrote request and error assertions to expect RuntimeError on failures and assert the absence of the `update` parameter; adjusted live test provider string.
  - tests/smoke_test.py: Adds derived carbon metric names (map _energy_ → _carbon_) to expected metric providers and skips stdout-count validation for those derived metrics.
  - tests/test_config_opts.py: Accepts additional configured provider `carbon_intensity_static_machine` in assertions.

## Testing / behavior impact
- Derived carbon metrics are now created per energy+carbon_intensity pairing and stored as microgram CO2 equivalents (ugCO2e).
- Several tests now expect RuntimeError for provider HTTP errors instead of empty DataFrames.
- Test expectations and metric-name mappings updated to reflect dynamic naming.
- Checklist: the PR contains an unchecked item to verify whether these changes also fix front-end display errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/green-coding-solutions/green-metrics-tool/pull/1697?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/post_metric_providers/calculate_co2_intensity.py
+++ b/lib/post_metric_providers/calculate_co2_intensity.py
@@ -52,7 +52,7 @@ def calculate_co2_intensity(run_id):
             if not energy_values:
                 continue
 
-            detail_name = f"{energy_metric}_{energy_detail_name}_{carbon_metric}_{carbon_detail_name}"
+            detail_name = f"{energy_detail_name}_{carbon_metric}_{carbon_detail_name}"
             derived_metric = energy_metric.replace('_energy_', '_carbon_')
             derived_metric_id = DB().fetch_one('''
                 INSERT INTO measurement_metrics (run_id, metric, detail_name, unit)
--- a/lib/post_metric_providers/calculate_co2_intensity.py
+++ b/lib/post_metric_providers/calculate_co2_intensity.py
@@ -8,8 +8,6 @@ from io import StringIO
 
 from lib.db import DB
 
-DERIVED_METRIC = 'psu_carbon_elephant_machine'
-
 def calculate_co2_intensity(run_id):
     carbon_intensity_metrics = DB().fetch_all('''
         SELECT id, metric, detail_name
@@ -55,11 +53,12 @@ def calculate_co2_intensity(run_id):
                 continue
 
             detail_name = f"{energy_metric}_{energy_detail_name}_{carbon_metric}_{carbon_detail_name}"
+            derived_metric = energy_metric.replace('_energy_', '_carbon_')
             derived_metric_id = DB().fetch_one('''
                 INSERT INTO measurement_metrics (run_id, metric, detail_name, unit)
                 VALUES (%s, %s, %s, %s)
                 RETURNING id
-            ''', params=(run_id, DERIVED_METRIC, detail_name, 'ugCO2e'))[0]
+            ''', params=(run_id, derived_metric, detail_name, 'ugCO2e'))[0]
 
             csv_buffer = StringIO()
             carbon_times = [entry[0] for entry in carbon_values]
@@ -81,4 +80,3 @@ def calculate_co2_intensity(run_id):
                 columns=('measurement_metric_id', 'value', 'time')
             )
             csv_buffer.close()
-
--- a/metric_providers/carbon/intensity/elephant/machine/provider.py
+++ b/metric_providers/carbon/intensity/elephant/machine/provider.py
@@ -150,7 +150,6 @@ class CarbonIntensityElephantMachineProvider(BaseMetricProvider):
             'region': self.region,
             'startTime': self._format_time(self._start_time),
             'endTime': self._format_time(self._end_time),
-            'update': 'true',
         }
 
         if self.provider_filter:
--- a/tests/metric_providers/test_electricitymaps_provider.py
+++ b/tests/metric_providers/test_electricitymaps_provider.py
@@ -1,6 +1,5 @@
 import os
 import shutil
-import pandas
 import pytest
 import requests
 import tempfile
@@ -214,9 +213,8 @@ def test_fallback_http_error_returns_none():
         return make_response(status_code=503)
 
     with patch('requests.get', side_effect=side_effect):
-        result = provider._read_metrics()
-    assert isinstance(result, pandas.DataFrame) and result.empty
-    assert provider._error_string != ''
+        with pytest.raises(RuntimeError, match='503'):
+            provider._read_metrics()
 
 
 # --- _read_metrics: error paths ---
@@ -224,18 +222,15 @@ def test_fallback_http_error_returns_none():
 def test_http_error_returns_none_and_logs():
     provider = profiled_provider()
     with patch('requests.get', return_value=make_response(status_code=500)):
-        result = provider._read_metrics()
-    assert isinstance(result, pandas.DataFrame) and result.empty
-    assert '500' in provider._error_string
+        with pytest.raises(RuntimeError, match='500'):
+            provider._read_metrics()
 
 
 def test_network_failure_returns_none_and_logs():
     provider = profiled_provider()
     with patch('requests.get', side_effect=requests.RequestException('timeout')):
-        result = provider._read_metrics()
-    assert isinstance(result, pandas.DataFrame) and result.empty
-
-    assert provider._error_string != ''
+        with pytest.raises(RuntimeError, match='timeout'):
+            provider._read_metrics()
 
 
 # --- _read_metrics: missing start/end times ---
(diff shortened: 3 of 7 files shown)


ArneTR:
@coderabbitai review

lib/post_metric_providers/calculate_co2_intensity.py
```diff
@@ -55,11 +53,12 @@ def calculate_co2_intensity(run_id):
                 continue
 
             detail_name = f"{energy_metric}_{energy_detail_name}_{carbon_metric}_{carbon_detail_name}"
```
ArneTR: can we now here remove "energy_metric" ? Is renduntant now I believe and having this strin shorter is quite desirable

metric_providers/carbon/intensity/elephant/machine/provider.py
```diff
@@ -150,7 +150,6 @@ def _read_metrics(self):
             'region': self.region,
             'startTime': self._format_time(self._start_time),
             'endTime': self._format_time(self._end_time),
-            'update': 'true',
```
ArneTR, line 153: what is the reason for that removal?

tests/metric_providers/test_electricitymaps_provider.py
```diff
@@ -214,28 +213,24 @@ def side_effect(url, **_kwargs):
         return make_response(status_code=503)
 
     with patch('requests.get', side_effect=side_effect):
-        result = provider._read_metrics()
-    assert isinstance(result, pandas.DataFrame) and result.empty
-    assert provider._error_string != ''
+        with pytest.raises(RuntimeError, match='503'):
```
ArneTR, line 216: You want a 503 error? Why is that?

I dont quite get the value of this test. It effectively tests if the patch method on requests.get works

Should it not test the electricity maps API somehow? Maybe use a sandbox token available as ENV?

tests/metric_providers/test_electricitymaps_provider.py
```diff
@@ -214,28 +213,24 @@ def side_effect(url, **_kwargs):
...
+            provider._read_metrics()
 
 
 # --- _read_metrics: error paths ---
 
 def test_http_error_returns_none_and_logs():
     provider = profiled_provider()
     with patch('requests.get', return_value=make_response(status_code=500)):
```
ArneTR: same question as above. feels like a test for patch()

tests/metric_providers/test_electricitymaps_provider.py
```diff
@@ -214,28 +213,24 @@ def side_effect(url, **_kwargs):
...
-        result = provider._read_metrics()
-    assert isinstance(result, pandas.DataFrame) and result.empty
-    assert '500' in provider._error_string
+        with pytest.raises(RuntimeError, match='500'):
+            provider._read_metrics()
 
 
 def test_network_failure_returns_none_and_logs():
     provider = profiled_provider()
     with patch('requests.get', side_effect=requests.RequestException('timeout')):
```
ArneTR: same question as above. feels like a test for patch()

tests/smoke_test.py
```diff
@@ -101,6 +101,17 @@ def test_db_rows_are_written_and_presented():
 
         metric_providers.extend(pm_additional_list)
 
+    # calculate_co2_intensity derives a carbon metric for each energy+carbon_intensity combo.
+    # Add those expected derived names so the assertion below can validate them.
+    if any('carbon_intensity_' in p for p in metric_providers):
```
ArneTR: since we control the test-config.yml file I guess you can be more explicit here. carbon intensity will always be present.
Actually it would be better to fail if it is not as this is what the smoke test should also test.

ArneTR (changes requested):
ty for the PR. The fixes for the phase stats look all sound to me. 

For the fix in smoke test I believe we can be more rigid.

Since I rarely check the tests in PRs for features some of the implementation I saw here was new to me. I feel I do not quite understand what you are testing for the electricitymaps providers and pointed out some questions on the test intent. Since it came up repeatedly I stopped commenting after a while. Would be better to chat about the idea here for all the "patch()" tests.

metric_providers/carbon/intensity/elephant/machine/provider.py
ribalba: With `update=True` the server updates all the sources. This can take quite long (30+ seconds on the live server) and just lets the test wait. For the tests we don't need the up to date number so removing it seems fine

tests/metric_providers/test_electricitymaps_provider.py
ribalba: I fixed the tests to the new raise behaviour. I chose 503 because it something I can see happening. 

Now with the new raise logic we are basically testing if the metric provider handles all the cases with a raise

tests/metric_providers/test_electricitymaps_provider.py
ribalba: I removed this as this is now testing the same error path as above

tests/metric_providers/test_electricitymaps_provider.py
ribalba: Also removed this as with the new raise Methodology we don't really need to test for timeout specifically anymore.

metric_providers/carbon/intensity/elephant/machine/provider.py
ArneTR: Could this be the reason for the timeouts we experience when querying the live system during measurements? 
How would this behave if 2-3 concurrent request would hit? The elephant would always trigger a refresh?

tests/metric_providers/test_electricitymaps_provider.py
ribalba: We test against the live system. There is a whole section
```
# --- Live integration test (hits the real Electricity Maps API) ---
```

tests/smoke_test.py
ribalba: changed to an assert

metric_providers/carbon/intensity/elephant/machine/provider.py
ribalba: Yes, that is why I fixed it to 4 worker threads now.

tests/smoke_test.py
```diff
@@ -103,14 +103,13 @@ def test_db_rows_are_written_and_presented():
...
     # Add those expected derived names so the assertion below can validate them.
-    if any('carbon_intensity_' in p for p in metric_providers):
-        derived_carbon_metrics = [
-            p.replace('_energy_', '_carbon_')
-            for p in metric_providers if '_energy_' in p
-        ]
-        metric_providers.extend(derived_carbon_metrics)
-    else:
-        derived_carbon_metrics = []
+    assert any('carbon_intensity_' in p for p in metric_providers), \
```
ArneTR, line 106: is this not the other way round? any -> not any?

ArneTR:
@coderabbitai review

ArneTR:
I have no objections here. If tests pass please merge in

tests/smoke_test.py
ribalba: Is it? Currently I want it to pass if the carbon intensity is present and fails when it's not

ArneTR: ja, true. I somehow had the thought assert fails if the condition is true

ArneTR:
Some of the tests are failing but they seem to be unrelated / GitHub Issues.

I will bring this in for now and create follow-up issues to dive deeper. I think it is paramount to have the functionality with the broken phase stats to be fixed

PR #1701, opened 2026-05-28
Fixes tests under Mac

ribalba:
We had the problem that tests under mac would corrupt the docker VM. Building on the work that I did some time back I could now home in. The problem is a stale inode cache in the docker demon. This PR fixes the tests under MacOS

## Problem
Tests running on macOS were corrupting the Docker VM due to stale inode cache in the Docker daemon. This occurred when directory cleanup operations (`rmtree` followed by `mkdir`) caused inode mappings to change across test runs.

## Solution
Introduces platform-aware directory cleanup logic to preserve inode stability on macOS while maintaining full cleanup on other systems.

## Changes
**lib/job/run.py**
- Imports `host_platform` from `lib`
- Modified `_process()` method's cleanup logic to conditionally handle macOS: deletes directory contents in place (removing symlinks and child directories) while preserving the directory itself, versus using full `shutil.rmtree()` on non-macOS platforms

**lib/scenario_runner.py**
- Updated `_initialize_folder()` with macOS-specific cleanup: removes contents in place and exits early, avoiding `rmtree()` + `mkdir()` operations that could change inode mappings under Docker Desktop virtiofs
- Modified Docker build artifact cleanup to call `_initialize_folder()` on macOS (preserving inode), while non-macOS continues using `shutil.rmtree()`

**Impact**: +35/-2 lines across both files. No public API changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/green-coding-solutions/green-metrics-tool/pull/1701?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/job/run.py
+++ b/lib/job/run.py
@@ -17,6 +17,7 @@ from lib.db import DB
 from lib.user import User
 from lib.terminal_colors import TerminalColors
 from lib.scenario_runner import ScenarioRunner
+from lib import host_platform
 import optimization_providers.base
 
 
@@ -94,6 +95,21 @@ class RunJob(Job):
                 )
 
         finally:
-            shutil.rmtree(runner._tmp_folder) # we see no sane reason for keeping tmp files on the cluster after a run
+            # we see no sane reason for keeping tmp files on the cluster after a run.
+            # On macOS, however, we wipe contents in place rather than removing
+            # _tmp_folder itself, so the directory's inode (and the inodes of subdirs
+            # like repo/) stay stable across runs. Docker Desktop on macOS caches
+            # inode<->path mappings in its virtiofs layer; deleting and recreating a
+            # bind-mount source between runs causes subsequent containers to see a
+            # stale/empty view until Docker Desktop is restarted.
+            if host_platform.is_macos():
+                if runner._tmp_folder.exists():
+                    for child in runner._tmp_folder.iterdir():
+                        if child.is_symlink() or not child.is_dir():
+                            child.unlink()
+                        else:
+                            shutil.rmtree(child, ignore_errors=False)
+            else:
+                shutil.rmtree(runner._tmp_folder)
             self._run_id = runner._run_id # might not be set yet due to error
             user.deduct_measurement_quota(self._machine_id, int(runner._last_measurement_duration/1_000_000)) # duration in runner is in microseconds. We need seconds
--- a/lib/job/run.py
+++ b/lib/job/run.py
@@ -102,14 +102,12 @@ class RunJob(Job):
             # inode<->path mappings in its virtiofs layer; deleting and recreating a
             # bind-mount source between runs causes subsequent containers to see a
             # stale/empty view until Docker Desktop is restarted.
-            if host_platform.is_macos():
-                if runner._tmp_folder.exists():
-                    for child in runner._tmp_folder.iterdir():
-                        if child.is_symlink() or not child.is_dir():
-                            child.unlink()
-                        else:
-                            shutil.rmtree(child, ignore_errors=False)
-            else:
-                shutil.rmtree(runner._tmp_folder)
+            if runner._tmp_folder.exists():
+                for child in runner._tmp_folder.iterdir():
+                    if child.is_symlink() or not child.is_dir():
+                        child.unlink()
+                    else:
+                        shutil.rmtree(child, ignore_errors=False)
+
             self._run_id = runner._run_id # might not be set yet due to error
             user.deduct_measurement_quota(self._machine_id, int(runner._last_measurement_duration/1_000_000)) # duration in runner is in microseconds. We need seconds
--- a/lib/job/run.py
+++ b/lib/job/run.py
@@ -8,7 +8,6 @@ import faulthandler
 faulthandler.enable(file=sys.__stderr__)  # will catch segfaults and write to stderr
 
 import os
-import shutil
 
 CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 
@@ -17,7 +16,6 @@ from lib.db import DB
 from lib.user import User
 from lib.terminal_colors import TerminalColors
 from lib.scenario_runner import ScenarioRunner
-from lib import host_platform
 import optimization_providers.base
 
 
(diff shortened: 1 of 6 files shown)


ArneTR:
If that fixes the issue I am down. However I am suprised that the tests work on this one on macOS because the test_client.py tests explicitely for the folder to be gone ... https://github.com/green-coding-solutions/green-metrics-tool/blob/2ccba5d0c11e4107fb4ae63014b9774c4082d643/tests/cron/test_client.py#L43

Did that work on your mac?

Furthermore: I would rather go for a unified behaviour to keep complexity low and clean the directory fully on all systems instead of unlinking it. 

Is there any argument against it?

ArneTR:
@coderabbitai review

lib/scenario_runner.py
```diff
@@ -274,20 +274,17 @@ def _custom_sleep(self, sleep_time):
...
-            # recreated between runs, subsequent containers can see a stale/empty
-            # view of the directory until Docker Desktop is restarted.
-            for child in path.iterdir():
-                if child.is_symlink() or not child.is_dir():
-                    child.unlink()
-                else:
-                    shutil.rmtree(child, ignore_errors=False)
-            return
-        shutil.rmtree(path, ignore_errors=False)
-        path.mkdir(parents=False, exist_ok=False)
```
ArneTR, line 278: the mkdir is now gone. is this intended?

lib/job/run.py
```diff
@@ -102,14 +102,12 @@ def _process(self, docker_prune=False, full_docker_prune=False):
...
-            if host_platform.is_macos():
-                if runner._tmp_folder.exists():
-                    for child in runner._tmp_folder.iterdir():
-                        if child.is_symlink() or not child.is_dir():
-                            child.unlink()
-                        else:
-                            shutil.rmtree(child, ignore_errors=False)
-            else:
-                shutil.rmtree(runner._tmp_folder)
+            if runner._tmp_folder.exists():
```
ArneTR: can this not use the helper function from scenario runner _initialize_folder? Otherwise we have a lot of code duplication and I see behaviours diveriging over time maybe

tests/test_config_opts.py
```diff
@@ -89,7 +89,7 @@ def test_provider_disabling_working():
 
     providers = list(run_data['configured_metric_providers'].keys())
     # or check bc in linux / macos there is a different result set of configured providers
-    assert providers == ['psu_energy_ac_sdia_machine', 'cpu_utilization_mach_system', 'psu_energy_ac_xgboost_machine'] or providers == ['disk_used_statvfs_system', 'network_io_procfs_system', 'memory_used_procfs_system', 'psu_energy_ac_sdia_machine', 'cpu_utilization_procfs_system', 'psu_energy_ac_xgboost_machine', 'carbon_intensity_static_machine'], 'Network Connections provider still in configured_metric_providers' #pylint: disable=consider-using-in
+    assert providers == ['psu_energy_ac_sdia_machine', 'cpu_utilization_mach_system', 'psu_energy_ac_xgboost_machine', 'carbon_intensity_static_machine'] or providers == ['disk_used_statvfs_system', 'network_io_procfs_system', 'memory_used_procfs_system', 'psu_energy_ac_sdia_machine', 'cpu_utilization_procfs_system', 'psu_energy_ac_xgboost_machine', 'carbon_intensity_static_machine'], 'Network Connections provider still in configured_metric_providers' #pylint: disable=consider-using-in
```
ArneTR, line 92: why is this in this PR? Is this a copy paste error?

runner.py
```diff
@@ -245,7 +245,17 @@
                 optimization_providers.base.run_reporters(runner._user_id, runner._run_id, runner._tmp_folder, runner.get_optimizations_ignore())
 
             if args.file_cleanup:
-                shutil.rmtree(runner._tmp_folder)
+                # Wipe contents in place rather than removing _tmp_folder itself, so the
+                # directory's inode (and the inodes of subdirs like repo/) stay stable
+                # across runs. Docker Desktop on macOS caches inode<->path mappings in
+                # its virtiofs layer; deleting and recreating a bind-mount source
+                # between runs causes subsequent containers to see a stale/empty view.
+                if runner._tmp_folder.exists():
```
ArneTR: can this not use the helper function from scenario runner _initialize_folder? Otherwise we have a lot of code duplication and I see behaviours diveriging over time maybe

ArneTR:
@ribalba Have you seen my comments?

ribalba:
Yes, but I didn't work over the weekend so I wanted to answer them today :)

lib/scenario_runner.py
ribalba: Yes, it wasn't really needed as the mkdir happened somewhere else. I have now added the case in the initalise folder to make it more future proof.

lib/job/run.py
ribalba: Good point. I just always replaced the `rmtree` to fix the error. Makes total sense to have one function that does this. Didn't see it down in the engine room

tests/test_config_opts.py
ribalba: One is for Mac and one for Linux. I only fixed the Linux version as Mac tests didn't work. Now the carbon provider is also checked in the Mac version

runner.py
ribalba: refactored

ArneTR:
Nice. works on my box and happy to bring this in.

During the process another Docker error that we had spotted before re-emerged:

```python3
    def test_carbondb_add():

        exp_data = ENERGY_DATA.copy()
        del exp_data['energy_uj']
        exp_data['energy_kwh'] = 2.7777777777777774e-13 # 1 uJ
        exp_data['carbon_kg'] = None # as we have no further carbon intensity supplied

        response = requests.post(f"{API_URL}/v2/carbondb/add", json=ENERGY_DATA, timeout=15)
        assert response.status_code == 202, Tests.assertion_info('success', response.text)

        data = DB().fetch_one('SELECT * FROM carbondb_data_raw', fetch_mode='dict')
        assert data is not None or data != []
        assert_expected_data(exp_data, data)

        assert data['ip_address'] is not None # some auto fill is happening
>       assert ipaddress.ip_address(data['ip_address']).is_private
E       AssertionError: assert False
E        +  where False = IPv4Address('160.79.104.10').is_private
E        +    where IPv4Address('160.79.104.10') = <function ip_address at 0x107b64a40>(IPv4Address('160.79.104.10'))
E        +      where <function ip_address at 0x107b64a40> = ipaddress.ip_address

```

The IP Bug, where fastapi does not resolve to a local IP anymore, but to a random one from Cloudfront / Cloudflare.

It is unrelated to the fixes in this PR, but just out of fun I wanted to log this. Really curious when we get to the bottom of this. Easiest fix for this: Restart Docker Desktop

PR #1705, opened 2026-06-01
Image Pruning Fix

ArneTR:
Image pruning was happening to all images with same tag. We wa…nt to delete only tags not all IDs on system

## lib/host_platform.py

Updated Docker image cleanup logic to remove only specific image tags rather than all images sharing the same tag.

**Changes:**
- `_docker_image_rows()`: Modified to return only `REPOSITORY:TAG` strings by removing the `{{.ID}}` field from the Docker format string
- `remove_gmt_tmp_images()`: Updated to filter and remove images using the tag name strings directly
- `remove_docker_images_except()`: Updated to filter images by tag name and pass names to `docker rmi`

The fix addresses a bug where image pruning would delete all images on the system that shared a tag, rather than just removing the specific tag references.

### DIFF
--- a/lib/host_platform.py
+++ b/lib/host_platform.py
@@ -35,7 +35,7 @@ def get_tmp_root():
         return Path(tempfile.gettempdir()).resolve(strict=True)
     else:
         return Path('/tmp/').resolve(strict=True)
-    
+
 def clear_file_system_caches():
     if is_windows():
         try:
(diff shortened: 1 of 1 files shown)


ArneTR:
@coderabbitai review

ArneTR:
@ribalba Somewhere in the refactor the logic how images are pruned on the system fell apart.

Can you double check that everything is according to your new idea of refactoring?

I check for the function use and found it only in this function. I changed it to output the names so we delete by tag.

Reason being: When stuff is in the whitelist (like the Kaniko image) we want to keep / delete only the tag. Since images with different tags can have same ID and thus would be deleted.
GMT tags all images by default with _gmt_run_tmp and we want to delete all thos. but not base images which might fly around

ribalba:
Looks ok to me

PR #1706, opened 2026-06-01
Fixes the colouring when rounded values are displayed in compare

ribalba:
Error reported here: https://github.com/green-coding-solutions/green-metrics-tool/pull/1702


ArneTR:
LGTM

PR #1713, opened 2026-06-04
Adds the --non-interactive arg to the installer

ribalba:
so that we don't need to hit enter like a wood pecker on install

### DIFF
--- a/lib/install_shared.sh
+++ b/lib/install_shared.sh
@@ -43,6 +43,7 @@ install_nvidia_toolkit_headers=false
 disable_path_security_checks=false
 ee_branch=''
 ask_elephant=true
+non_interactive=false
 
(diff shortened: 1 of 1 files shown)


ArneTR:
I do not see the value in this PR.

For me typing --non-interactive is the same as simply holding the enter key down until the blocking password prompt appears.

The downside of this flag is that we have to bring in every key that will be asked to the non-interactive group every time and maintain this functionality.

So my questions in short:
- Why not simply hold the enter key down?
- If you always to the same install why not simply have a shortcut in alfred or an alias "install_gmt" that resolves to 'install_linux.sh -X -P -T ' or whatever the values are that you constantly set? This is how the install is triggered on all our other boxes

PR #1714, opened 2026-06-04
Tries to fix the random ip issue under MacOs

ribalba:


tests/setup-test-env.py
```diff
@@ -239,7 +239,7 @@ def edit_etc_hosts():
 
 
 def build_test_docker_image():
-    subprocess.run(['docker', 'compose', '-f', test_compose_path, 'build'], check=True)
+    subprocess.run(['docker', 'compose', '-f', test_compose_path, 'build', '--no-cache'], check=True)
```
ribalba: I added this because I had a cache corruption locally and the tests didn't work anymore. Could also take this out again. We could also add it to the installer?

ArneTR: I also had the issue with the cache corruption lately. But never experienced it in 3 years before.

Unsure where this was coming from. 

Still I would argue to only bring this in if we have narrowed down why it happens, bc it can slow down builds tremendously. 

Vote to remove.

ArneTR:
PR looks good. See my comment. After resolve feel free to merge in

ArneTR:
Test error is unrelated to the fix

tests/setup-test-env.py
ribalba: removed

PR #1717, opened 2026-06-09
(fix): Creating carbon values for all componnents

ArneTR:
@ribalba I already created a branch for the fix that you wanted to look at.

Reason being: some of our verticals which rely on the carbon component value like webNRG and Green-Coding.ai started to fail.

Please have a look and please add tests once you get to it.

### DIFF
--- a/tests/api/test_api_software_add.py
+++ b/tests/api/test_api_software_add.py
@@ -151,11 +151,11 @@ def test_post_run_add_gitlab_tag():
 def test_post_run_add_gitlab_custom_api_base():
     run_name = 'test_' + utils.randomword(12)
     run = Software(name=run_name, repo_url='https://gitlab.rlp.net/green-software-engineering/oscar', email='testEmail', branch='', filename='', machine_id=1, schedule_mode='commit')
-    response = requests.post(f"{API_URL}/v1/software/add?no_url_check=true", json=run.model_dump(), timeout=15)
+    response = requests.post(f"{API_URL}/v1/software/add", json=run.model_dump(), timeout=15)
     assert response.status_code == 202, Tests.assertion_info('success', response.text)
 
     watchlist_item = utils.get_watchlist_item('https://gitlab.rlp.net/green-software-engineering/oscar')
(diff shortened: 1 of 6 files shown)


ribalba:
Sorry for all the runs. I can not reproce the error locally.

ribalba:
@ArneTR please review this as I changed the way the tests work as the URL didn't resolve anymore and we are getting rate limited more often by GitHub

tests/api/test_api_software_add.py
```diff
@@ -151,11 +151,11 @@ def test_post_run_add_gitlab_tag():
 def test_post_run_add_gitlab_custom_api_base():
     run_name = 'test_' + utils.randomword(12)
     run = Software(name=run_name, repo_url='https://gitlab.rlp.net/green-software-engineering/oscar', email='testEmail', branch='', filename='', machine_id=1, schedule_mode='commit')
-    response = requests.post(f"{API_URL}/v1/software/add", json=run.model_dump(), timeout=15)
+    response = requests.post(f"{API_URL}/v1/software/add?no_url_check=true", json=run.model_dump(), timeout=15)
```
ArneTR: Please Remove it from here as this tests the gitlab API and we check for the marker later explicitely for syntax also

ArneTR (commented):
I am generally fine with the change, but it should not be removed in the gitlab test. Made a comment.

2. Please add a test somewhere that the metric is actually generated. Display in the frontend so far was never a problem. Generation in the phase_stats is.

A good place would be the test_phase_stats.py in a test that actually imports CPU energy metrics. A cpu carbon metric should then be present also in the return.

tests/api/test_api_software_add.py
ribalba: This was the test that kept failing. rlp.net times out on GitHub (sometimes)

https://github.com/green-coding-solutions/green-metrics-tool/actions/runs/27762880298/job/82150508702

ArneTR:
Review ready?

ribalba:
yes

ArneTR:
LGTM. ty ❤️

PR #1728, opened 2026-06-15
(feat): Custom metrics can be any unit

ArneTR:
This PR allows for the metric value to be any character except newline (which would break parsing).

Furthermore frontend was checked to not have XSS escapes with the value anymore now with new allowed chars

## Custom Metrics Character Validation Expansion

This PR expands the character set allowed in custom metric values by changing validation logic to accept any character except newlines, which would break parsing.

### Changes

**Backend** (`lib/schema_checker.py`):
- Centralized allowed character rules into top-level constants (`VALID_CHARS`, `VALID_CHARS_SPACE`)
- Added new validation helper `no_newlines()` method to reject values containing newline characters
- Updated custom metrics `unit` field validation to use `no_newlines` instead of `is_valid_string_with_spaces`, effectively expanding the allowed character set from a restricted whitelist to allowing all characters except newlines

**Frontend** (`frontend/js/helpers/metric-boxes.js`):
- Consistently applied HTML escaping (`escapeString()`) to unit strings in tooltip/title attributes across `displaySimpleMetricBox` and `displayDiffMetricBox` functions to prevent XSS attacks with the expanded character set

### Security
The frontend was audited and verified to properly handle XSS escaping with the newly allowed characters, confirming no security vulnerabilities are introduced.

### DIFF
--- a/frontend/js/helpers/metric-boxes.js
+++ b/frontend/js/helpers/metric-boxes.js
@@ -135,8 +135,8 @@ const displaySimpleMetricBox = (phase, metric_name, metric_data, detail_name, de
             <td>${scope}</td>
             <td><span class="detail-name-ellipsis" title="${detail_name}">${detail_name}</span></td>
             <td>${metric_data.type}</td>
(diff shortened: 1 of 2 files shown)


ArneTR:
@coderabbitai review

PR #1729, opened 2026-06-15
(feat): Reboot happens now always on package update

ArneTR:
## Overview
This PR implements automatic system reboots when OS packages are installed during maintenance, with persistence of revalidation state across reboots.

## Database Schema Changes
- Adds `needs_revalidation` boolean column to `machines` table (defaults to `false`)

## Client Logic Changes
- **New `reboot()` function**: Sets client status to 'reboot', syncs filesystem, triggers systemctl reboot, then sleeps
- **Startup initialization**: Reads persisted `machines.needs_revalidation` flag from database to determine if validation is required after boot
- **Scheduled maintenance (12-hour interval)**: When `do_maintenance()` detects installed packages, sets `machines.needs_revalidation = true` and calls `reboot()` instead of just setting an in-process flag
- **Post-job maintenance (finally block)**: Similarly triggers reboot with database persistence when packages are installed
- **Database persistence**: Revalidation flag is stored in the database rather than relying on in-process state, ensuring the requirement survives reboots

## Impact
Ensures newly installed packages are loaded into the running system through automatic reboots, with the revalidation requirement persisting across the reboot cycle.

### DIFF
--- a/cron/client.py
+++ b/cron/client.py
@@ -94,7 +94,7 @@ def reboot_if_uptime_exceeded(reboot_after_s):
         print(f"Uptime {uptime_seconds:.0f}s exceeds reboot_after_seconds {reboot_after_s}s. Rebooting...")
         set_status('reboot')
         subprocess.check_output(['sync'], encoding='UTF-8', errors='replace')
-        subprocess.check_output(['sudo', 'systemctl', 'reboot'], encoding='UTF-8', errors='replace')
+        subprocess.check_output(['/usr/bin/sudo', '/usr/bin/systemctl', 'reboot'], encoding='UTF-8', errors='replace')
         time.sleep(86400) # reboot request might not be handled directly, thus we wait until this process gets killed.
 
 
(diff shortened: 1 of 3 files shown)


ArneTR:
@coderabbitai review

ArneTR:
Tests running here: https://github.com/green-coding-solutions/green-metrics-tool/actions/runs/27541048897

cron/client.py
```diff
@@ -224,6 +224,12 @@ def do_measurement_control():
         # endlessly in validation until manually handled, which is what we want.
         time.sleep(config['cluster']['client']['time_between_control_workload_validations'])
 
+def reboot():
+    set_status('reboot')
+    subprocess.check_output(['sync'], encoding='UTF-8', errors='replace')
+    subprocess.check_output(['/usr/bin/sudo', '/usr/bin/systemctl', 'reboot'], encoding='UTF-8', errors='replace')
+    time.sleep(86400)
```
ribalba, line 231: @ArneTR do you really want sleep here for a day? Wouldn't a shorter sleep with a following error be better to notice the failing reboot?

ArneTR: This has never failed so far. So if it does I suppose it is a bigger error.

If it does not reboot I have to trigger it again anyway as we want to force that behaviour. So the only sensical option that I see is making this an infinite loop ... which I want to avoid for now until I understand possible error cases

PR #1731, opened 2026-06-16
Added Carbon Simulation call also to test function loop

ArneTR:
@ribalba While refactoring some code I noticed you did not bring an update to the runner core loop to the mimic in the test_functions.py 
Was that intended?

Please have a look at this PR where I replayed the change to the test_functions ... is that ok? or will this cause issues?


ribalba:
No, this was an oversight on my end

ArneTR:
ok. then please put on list for next week to fix tests and merge after

PR #1734, opened 2026-06-17
Allow users to update their ssh private keys (migration script)

davidkopp:
PR https://github.com/green-coding-solutions/green-metrics-tool/pull/1625 introduced the feature that users can set SSH private keys on their own.

However, when I want to update the setting I get an error:

Image

The seed data in `docker/structure.sql:62` includes "ssh_private_key" in the updateable_settings list, but none of the existing migration scripts add it. This PR adds the missing migration script.

The issue was discovered on the SaaS version (https://metrics.green-coding.io/settings.html).

### DIFF
--- a/migrations/2026_06_17_ssh_private_key_updateable.sql
+++ b/migrations/2026_06_17_ssh_private_key_updateable.sql
@@ -7,4 +7,4 @@ SET capabilities = jsonb_set(
         '["ssh_private_key"]'::jsonb
     ),
     true
-);
+) WHERE id != 0;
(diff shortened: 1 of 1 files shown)


ArneTR:
Thanks for the PR. I have updated this for your user.

The migration looks good but should exclude user 0.

Can you add "WHERE id != 0;"

davidkopp:
Migration is updated. Thanks for updating the user :)

ArneTR:
ty for the PR ❤️

PR #1740, opened 2026-06-23
feat: add timestamp parsing for container execution logs (#1343)

### LINKED ISSUE #1343: Time-splitting of container logs
The PR https://github.com/green-coding-solutions/green-metrics-tool/pull/1341 enabled the storing of logs in a JSON structure. One advantage of this is that the different log types are now separated properly:

- logs from the container
- logs from setup commands
- logs from flow commands
- system related logs

While the logs from the setup commands and flow commands happen during a specific phase (setup commands in `[BOOT]` phase and flow commands in `[RUNTIME]` phase), the logs from the container usually span multiple phases (`[BOOT]`, `[IDLE]`, `[RUNTIME]`). Therefore the phase of the log type `container_execution` is stored as `[MULTIPLE]`.

@ArneTR proposed in the comment https://github.com/green-coding-solutions/green-metrics-tool/pull/1341#issuecomment-3295479744 to time-split the container logs to be able to separate between boot, idle, runtime and sub-runtime logs. For this, the logs need to be time-keyed. At least for the container logs that should be not problem ... not so much for the run/exec logs.

## Current Implementation

The container logs are collected using:

```python
log = subprocess.run(
    ['docker', 'logs', container_id],
    check=True,
    encoding='UTF-8',
    errors='replace',
    stdout=stdout_behaviour,
    stderr=stderr_behaviour,
)
```

Docker's `logs` command provides built-in timestamp support that could be used:

- **Flag**: `--timestamps` or `-t`
- **Format**: RFC3339Nano timestamp prefix
- **Example Output**:
  ```
  2025-09-17T06:46:17.013138795Z first log
  2025-09-17T06:46:19.014901570Z second log
  2025-09-17T06:46:20.016411547Z third log
  ```

Logs are currently stored as JSON objects with the following structure:

```python
log_entry = {
    'type': log_type.value,
    'id': str(log_id),
    'cmd': command_string,
    'phase': phase,
    'stdout': stdout,  # optional
    'stderr': stderr,  # optional
    'flow': flow,      # optional
    'class': exception_class  # optional
}
```

### Option 1: Integrated Timestamp Storage

**Approach**: Add timestamps to existing log structure

**Changes Required**:
1. Modify docker logs command in `_read_container_logs()`:
   ```python
   ['docker', 'logs', '--timestamps', container_id]
   ```

2. Parse timestamps in `_handle_process_output()`:
   ```python
   for line in log_output.split('\n'):
       if not line.strip():
           continue

     # Parse RFC3339Nano timestamp
     timestamp_match = re.match(r'^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+Z)\s+(.*)$', line)
     if timestamp_match:
         timestamp, content = timestamp_match.groups()
         entries.append({
             'timestamp': timestamp,
             'content': content
         })
     else:
         # Fallback for lines without timestamps
         entries.append({
             'timestamp': None,
             'content': line
         })
   ```

3. Update log entry structure for container execution only:

   ```python
   # For LogType.CONTAINER_EXECUTION only
   if log_type == LogType.CONTAINER_EXECUTION:
       log_entry = {
           'type': log_type.value,
           'id': str(log_id),
           'cmd': command_string,
           'phase': phase,
           'stdout_entries': parsed_stdout_entries,  # List of {timestamp, content}
           'stderr_entries': parsed_stderr_entries,  # List of {timestamp, content}
           # ... other fields
       }
   else:
       # Keep existing structure for other types like setup_command and flow_command
       log_entry = {
           'type': log_type.value,
           'id': str(log_id),
           'cmd': command_string,
           'phase': phase,
           'stdout': stdout,  # Plain string
           'stderr': stderr,  # Plain string
           # ... other fields
       }
   ```

### Option 2: Separate Timestamp-Aware Logs

**Approach**: Maintain existing logs and add parallel timestamp-aware storage

**Pros**:
- Full backward compatibility
- No impact on existing log consumers

**Cons**:
- Increased storage overhead
- Duplicate data maintenance

### 1. Log Type Differentiation

- **Container logs**: Will have per-line timestamps with nanosecond precision
- **Setup/Flow logs**: Will maintain current plain string format
- **Mixed handling**: Code must handle both timestamped and non-timestamped logs

### 2. Timezone Handling
- Docker timestamps are in UTC (Z suffix)
- Consider if local timezone conversion is needed for display

### 3. Performance Impact
- Timestamp parsing adds processing overhead
- Consider lazy parsing if logs are large

### 4. Backward Compatibility
- If the JSON structure is changed (approach 1), a migration script is needed
- Only affects container execution logs structure
- Setup/Flow command logs remain unchanged

### 5. Phase Detection
- The timestamps could be used to determine the correct phase to which the log message belongs
- Further analysis may be necessary to specify where and how this should be done

### 6. Frontend
- How to display the timestamps in the frontend is an open question for me

sufirehman:
## Description
Implements Option 1 (Integrated Timestamp Storage) from the issue, scoped specifically to LogType.CONTAINER_EXECUTION. Adds the --timestamps flag to the docker logs call in _read_container_logs(), parses each line's RFC3339Nano timestamp prefix into {timestamp, content} entries, and stores them as stdout_entries/stderr_entries for container logs only. SETUP_COMMAND and FLOW_COMMAND log types keep their existing plain stdout/stderr string structure completely unchanged.

## Related Issue
Closes #1343

## Motivation and Context
Container logs span multiple phases (BOOT, IDLE, RUNTIME) and are currently stored as a single undifferentiated block per phase. Docker's --timestamps flag gives per-line RFC3339Nano timestamps for free, which lets us time-key container log lines for future phase-splitting and analysis, as proposed by @ArneTR.

## How Has This Been Tested?
Added 6 tests: timestamp parsing on normal docker --timestamps output, fallback to {timestamp: None, content: line} for any line that doesn't match the expected format, empty input handling, and three tests confirming CONTAINER_EXECUTION entries use the new structure while SETUP_COMMAND and FLOW_COMMAND entries remain plain strings, completely unaffected. All 6 tests pass. (Note: the existing test teardown calls Tests.reset_db(), which requires the test Postgres container to be running; this is unrelated to the change and fails identically on main without the test Docker stack started.)

## Scope notes
Per discussion with @ArneTR, this PR does not touch frontend display. Per his guidance, the plan for a follow-up is to simply prepend each line with its timestamp in the Logs tab, similar to how `docker logs -t` displays output, with no interleaving logic for now.

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Summary

- Added Docker timestamp support for container execution logs.
- Parses RFC3339Nano prefixes into `{timestamp, content}` entries, with `None` for unmatched lines.
- Stores container logs as `stdout_entries`/`stderr_entries` while preserving plain strings for setup and flow commands.
- Added tests covering parsing, fallback behavior, empty input, and log-type-specific storage.

Frontend display and phase detection remain out of scope.


sufirehman:
Hi @ArneTR, this PR has been open since June 23 without any feedback yet, so just checking in. I've since resolved a merge conflict in lib/scenario_runner.py that came up from recent upstream changes, and the timestamp parsing tests still pass cleanly against the merged code.

No rush at all if you're busy, just wanted to make sure it hadn't slipped through unnoticed. Happy to adjust the approach if anything needs changing.

ArneTR:
Hey @sufirehman,

sorry for the this. The PR simply slipped through 😨 

Reviewing now

ArneTR:
@coderabbitai review

lib/scenario_runner.py
```diff
@@ -2046,22 +2060,46 @@ def _add_to_current_run_log(self, container_name, log_type, log_id, cmd, phase,
...
-            log_entry['stdout'] = utils.filter_sensitive_data(log_entry['stdout'])
-        if stderr is not None:
-            if isinstance(stderr, str):
-                log_entry['stderr'] = stderr.replace('\x00', '0x00') # Postgres cannot handle null bytes (\x00) in text fields or \u0000 in JSONB columns
-            elif hasattr(stderr, 'decode') and callable(getattr(stderr, 'decode')): # can happen if a timeout error has occured and stderr was thus not converted yet
-                log_entry['stderr'] = stderr.decode('UTF-8', errors='replace').replace('\x00', '0x00')
-            else:
-                log_entry['stderr'] = str(stderr).replace('\x00', '0x00') # we just force it to a string. This can garble output a bit though
-            log_entry['stderr'] = utils.filter_sensitive_data(log_entry['stderr'])
+        if log_type == LogType.CONTAINER_EXECUTION:
```
ArneTR: This is very duplicated code. I feel this whole function should be refactored to be more compact and not branch into two big branches with identical sub-branches

lib/scenario_runner.py
```diff
@@ -2046,22 +2060,46 @@ def _add_to_current_run_log(self, container_name, log_type, log_id, cmd, phase,
...
+            # filter_sensitive_data's PRIVATE_KEY_RE can match blocks spanning multiple lines.
+            if stdout is not None:
+                if isinstance(stdout, str):
+                    stdout_str = stdout
+                elif hasattr(stdout, 'decode') and callable(getattr(stdout, 'decode')):
+                    stdout_str = stdout.decode('UTF-8', errors='replace')
+                else:
+                    stdout_str = str(stdout)
+                stdout_str = utils.filter_sensitive_data(stdout_str)
+                log_entry['stdout_entries'] = _parse_timestamped_log_lines(stdout_str)
```
ArneTR: Storing in this element is useful. But in the end it still must be copied as string only int stdout / stderr. Otherwise the logs currently do not show in the frontend

ArneTR (changes requested):
Thanks for the PR. I now had time to review it.

I made remarks in the code directly where appropriate.

But overall this PR does not fully implement the idea intended. 
- It does not show in the frontend. Although discussed that we do not want to touch the interleaving and phase splitting in the frontend, it should still show in the legacy box. I made an idea in the code how to achieve this
- Not implemented is the fact that `docker exec` calls will not have timestamps. Since they typically make up the majority of our logs I feel this must be part of the implementation. Would be great if you supply some ideas here how to achieve this. As David said: This might be a trickier one though :) 
If you feel this is too tough to implement I can understand. But I feel for only having timestamps in the `docker logs` this PR is currently not enough to merge.

sufirehman:
Hey @ArneTR, thanks for the review!

Points 1 and 2 make sense, I'll fix the frontend display and clean up the branching in _add_to_current_run_log.

For the docker exec timestamps, I looked into it and the issue is structural: exec output bypasses the Docker daemon's log driver so there's no --timestamps equivalent available. I'll add a comment explaining this as a known gap rather than pushing a streaming workaround that would only give host-receipt timestamps with buffering issues. Let me know if you'd rather I take a different approach there.

Will push the fix shortly.

sufirehman:
Hi @ArneTR, pushed the updated implementation:

1. Fixed the frontend display issue, _add_to_current_run_log now populates both stdout_entries/stderr_entries (structured, timestamped) and plain stdout/stderr (content joined for the legacy log box). The existing test_logs_structure integration test, which already expected this behavior, now passes.

2. Refactored _add_to_current_run_log using a _stringify_process_output helper, collapsing the duplicated decode/str-fallback logic that appeared four times into one loop over stdout and stderr.

3. For docker exec timestamps: confirmed this is a structural limitation, exec output bypasses the Docker daemon log driver entirely so there's no --timestamps equivalent. Added comments at both exec call sites explaining this. Happy to discuss further if you'd prefer a different approach.

Let me know if anything else needs adjusting.

lib/scenario_runner.py
```diff
@@ -2046,22 +2075,24 @@ def _add_to_current_run_log(self, container_name, log_type, log_id, cmd, phase,
...
+            value_str = utils.filter_sensitive_data(_stringify_process_output(value))
+
+            if log_type == LogType.CONTAINER_EXECUTION:
+                # Frontend display: prepend each entry timestamp before its content,
+                # similar to `docker logs -t`. No stdout/stderr interleaving for now.
+                entries = _parse_timestamped_log_lines(value_str)
+                log_entry[f'{key}_entries'] = entries
+                # Must also be copied as plain string: the (legacy) frontend log box reads
+                # stdout/stderr directly and does not know about the *_entries structure.
+                log_entry[key] = '\n'.join(entry['content'] for entry in entries)
```
ArneTR, line 2094: why this operation? and not just have the following else-branch unconditional. joining the entries here should be the same as setting log_entry[key] = value_str or?

ArneTR (changes requested):
The function is a lot clear now, ty.

I have made a remark where I think it can further be improved.

Will follow up on how to extend the PR.

ArneTR:
The remark that docker exec simply cannot do what we need at the moment is nice to have in the comments.
However I feel in the end this PR needs to address this to fulfill the open issue.

To make this achievable I think GMT should inject the functionality we need into the container. Here an idea, happy for your thoughs and remarks:

- For every exec call we do not simply execute the program, but rather inject a program. Let's call it `ts_x` for now.
- The program will be very similar to `ts` from moreutils with two important distinctions:
  - It can run in a container without any dependencies. THis means also no dynamically linked *libc*. It should bring its own libc statically linked
  - It does not require a pipe on the CLI (which is shell feature ... we cannot assume a shell in the container), but does process inheritance (similar to how timeout or stdbuf work)

This way we can then run `ts docker exec MY_PROGRAM` and it will prepend everything with timestamps, similar to how `docker logs` works.

## Considerations
- Measurement of overhead in a heavy logging file
- Measurement of overhead in a file that does no logging at all

This makes the PR bigger, but would fulfill the whole functionality. It requires some C work, but I guess also something a bit out of the ordinary :)

Let me know how you feel about extending the PR like this and if you feel you want to take this on.

sufirehman:
Hey @ArneTR, good question on the join.

They're actually not equivalent in this case. _parse_timestamped_log_lines reads value_str but doesn't mutate it, so value_str at that point still contains the raw docker logs -t output with RFC3339Nano timestamps prefixed on every line (e.g. "2025-09-17T06:46:17.013138795Z hello from container"). Using value_str directly for the plain field would put raw timestamped text into stdout/stderr, which breaks the legacy frontend display. The join reconstructs the timestamp-stripped version, which is what the frontend actually needs. I've added a clarifying comment in the code to make this explicit.

On the ts_x idea: I think it's a genuinely interesting approach and worth doing properly. A statically linked C binary with process inheritance, no shell dependency, and overhead benchmarking is meaningfully different in scope from what this PR set out to address though. Rather than extending this PR further, would it make sense to merge what's here (docker logs timestamps working and tested end to end) and open a follow-up issue for ts_x with your design notes as the starting point? Happy to open that issue if that works for you.

PR #1744, opened 2026-06-24
Deleting cancelled jobs

ArneTR:
This PR forces deletion of cancelled jobs also.

To ensure no status is forgotten an enum is introduced into the database

This change makes job cleanup include cancelled jobs and tightens job state handling with a database enum.

- `Job.clear_old_jobs` now deletes `CANCELLED` jobs along with the existing old `FAILED` and `FINISHED` jobs, while preserving the other cleanup rules.
- The job state column is migrated from ազատ-form text to a `job_state` enum, with `jobs.state` now `NOT NULL` and defaulting to `WAITING`.
- The schema definition is updated to match the new enum-based state model.


ArneTR:
@coderabbitai review

PR #1745, opened 2026-06-24
Submit software and keep values

ribalba:


ArneTR:
OMG this is so good. I love it. Was annoyed many times by it but somehow made my own workaround with tabbing windows 0_o

PR #1749, opened 2026-06-26
(feat): Multiple Detail Names now aggregate in key metrics if possible

ArneTR:
Before it was only showing one value for the possible detail names.

This can be confusing especially with multiple containers for network traffic but also can be confusing if multiple RAPL sockets are in the system.

If possible the value is now aggregated.
Screenshot 2026-06-26 at 11 15 27 AM

On compare it will always show "Not compareable" and hint to the table view at the bottom where values can be individually compared
Screenshot 2026-06-26 at 11 17 01 AM

- Updated key metric aggregation so metrics with multiple detail names can show a combined value instead of a single source, improving cases like network traffic and RAPL-derived metrics.
- Changed comparison behavior to consistently show “Not comparable” for aggregated values that can’t be meaningfully compared, with tooltip guidance directing users to the metrics table for per-source values.
- Added demo data for network-related runtime metrics to support the new comparison behavior.
- Refactored and expanded frontend tests to assert the new aggregated / “Not comparable” UI states, including tooltip-bearing HTML.


ArneTR:
@coderabbitai review

PR #1755, opened 2026-07-02
Settings watcher

ArneTR:
PR Summary

- Added many checks to system_checks to validate machine hardware state and settings:
  - RAPL Power Limits
  - check_cpu_scaling_driver
  - check_cpu_frequency
  - check_cpu_turbo_boost
  - check_cpu_smt
  - check_cpu_governor
  - check_pci_devices
  - check_usb_devices
  - check_dram amount

-  Temperature checks moved into system_checks (lib/system_checks.py, cron/client.py, lib/configuration_check_error.py)

-  validate_temperature() in client.py is replaced by a new check_temperature() in system_checks.py (first entry in start_checks). Temperature failures raise TemperatureException, a new subclass of ConfigurationCheckError, which propagates through job.process() to a new except  TemperatureException block in client.py. Hot machines trigger cooldown status events; cold machines trigger warmup (CPU spin-up). Both write  the temperature reading to client_status.data. After 10 consecutive failures the process aborts. Counters reset on a successful job run.

##  Maintenance cleanup (tools/cluster/maintenance_original.py, install_linux.sh)

-  Removed check_systemd_timers() (dead code after it moved to system_checks_root.py) and the corresponding __GMT_USER__ sed substitution from
  the installer.

##  New test (tests/test_root_script_imports.py)

-  AST-based test asserting that system_checks_root.py, hardware_info_root_original.py, and maintenance_original.py import only stdlib modules — enforcing the invariant that root-owned scripts cannot pull in venv packages.

Expanded startup system validation to cover more machine configuration drift, including RAPL power limits, CPU governor/frequency/scaling driver, turbo boost, SMT, PCI/USB allowlists, DRAM, and Docker registry mirror settings.

Moved temperature enforcement into `system_checks`, added `TemperatureException`, and updated the client to handle hot/cold machines with cooldown/warmup flow, status updates, retry counting, and abort after repeated failures.

Added a root-only helper for privileged system checks, a stdlib-only import safety test for root-owned scripts, and removed the old systemd-timers maintenance path and installer substitution.


ArneTR:
@ribalba Do you have an idea for more checks?

This PR was born bc after some updates to the OS and BIOS and the plethora of changes we have done over time sometimes I spotted machines not configured as intended or a setting got lost after a re-install

ribalba (commented):
Looks good. 

- One could also check if cron is doing something?
- Could we also check if we are running in a VM?

ArneTR:
The cron is a great idea. I will bring that in an then merge after.

The VM I plan to have as a configurable setting. A warning if running inside a VM we have already if the reporters have trouble capturing data. I feel that is sufficient for now.

ArneTR:
@coderabbitai review

PR #1756, opened 2026-07-03
Docker login encryption and submitted jobs URI encryption

ArneTR:
Added encryption support for embedded repo credentials and Docker registry/login credentials across API, job execution, persistence, and UI.

- API: in `/v1/runs/add`, embedded credentials in `software.repo_url` are encrypted for storage, with `EncryptionConfigurationError` mapped to a `422`; tag/commit “last marker” resolution now uses the original unencrypted URL.
- Runner/checkout: `ScenarioRunner` now accepts `docker_credentials`, generates a temporary Docker `config.json` (0600) for private pulls/builds, mounts it into Kaniko, sets `DOCKER_CONFIG`, and cleans up stale docker/SSH artifacts between runs.
- Git URI credentials: extracted userinfo from HTTP(S) clone URIs; git fetch/checkout uses decrypted credentials via askpass/env rather than embedding secrets in URLs.
- User/DB: introduced `users.docker_credentials` + migration and updated user capabilities to allow setting it; `User` persists encrypted Docker credentials, exposes a presence flag, and supports updating/clearing via `change_setting`.
- CLI/cron/workflows: added `--docker-credentials` CLI option (JSON list -> `SecureVariable` passwords) and wired it into the job execution path; cron/watchlist marker lookups now skip items on `RuntimeError`.
- Frontend: added Measurements settings UI for managing Docker registry credentials and status/clear/save behavior; also improved URL/detail rendering safety (escaping/truncation/tooltip titles).
- Git host handling: tightened `get_git_api` construction for custom GitHub/GitLab hosts.
- Tests/CI: expanded coverage for encrypted repo URL storage, Docker credential persistence and non-leakage, private image pulls (with/without credentials), cron execution with DB-stored credentials, and passing Docker secrets through GitHub Actions workflows.


ArneTR:
@coderabbitai review

ArneTR:
tests are ok

ArneTR:
@coderabbitai review

api/scenario_runner.py
```diff
@@ -958,16 +961,29 @@ async def runs_add(software: Software, no_url_check: bool = False, user: User =
         except Exception as exc:
             raise HTTPException(status_code=422, detail=str(exc)) from exc
 
+    unencrypted_repo_url = software.repo_url
+    _parsed_repo = urlparse(software.repo_url)
+    if _parsed_repo.username:
```
ribalba: **🔴 High — password-only credential URLs bypass encryption *and* redaction**

This gate (and the mirror at `lib/scenario_runner.py:218`) only fires when a username is present. For a valid git/curl form like `https://:glpat-SECRET@gitlab.com/repo.git`, `urlparse().username == ''` (falsy — verified), so:
- credentials are **not** encrypted here → the raw token is stored in `jobs.url` and `watchlist.repo_url`;
- `ScenarioRunner.__init__` doesn't strip it → raw token in `runs.uri`;
- `URI_CREDENTIALS_RE` requires `[^\s/@:]+` for the username, so it doesn't match this form → the token is served by `GET /v2/jobs` and echoed into the admin "new run" email in plaintext.

Gate on `_parsed_repo.username or _parsed_repo.password` (and add a redaction case for empty userinfo). The common `https://token@host` form is handled correctly — it's specifically the empty-username form that leaks.

lib/scenario_runner.py
```diff
@@ -193,8 +209,20 @@ def __init__(self,
...
 
-        if 'ssh_private_key' in self._arguments:
-            del self._arguments['ssh_private_key']
+        # security related keys we never want to log
+        del self._arguments['ssh_private_key']
+        del self._arguments['docker_credentials']
+
+        # Strip HTTP basic auth credentials from URI; keep clean version for logging and DB storage
+        _parsed_uri = urlparse(self._uri)
+        if _parsed_uri.username:
```
ribalba: Mirror of the empty-username gap flagged in `api/scenario_runner.py:966`: for `https://:token@host`, `_parsed_uri.username` is `''`, so the `else` branch leaves `self._uri` unstripped and the plaintext token ends up in `runs.uri`. Gate on `username or password`.

lib/scenario_runner.py
```diff
@@ -2824,6 +2892,7 @@ def cleanup(self):
         self.__relations.clear()
         self.__custom_metrics.clear()
         self.__sci_metrics.clear()
+        self.__uri_userinfo = None
```
ribalba, line 2923: **🟠 Medium — a reused runner loses HTTP-auth credentials after the first run**

`__init__` permanently strips userinfo from `self._uri` and stashes it in `self.__uri_userinfo`, but this line (in `cleanup()`, called from `run()`'s `finally`) nulls it. `runner.py` builds **one** runner and loops `set_filename(); run()` over `filenames * iterations`, and every `run()` re-invokes `_checkout_repository()`, which only applies creds `if self.__uri_userinfo`. So for a private `https://user:pass@host/repo`, iteration 1 clones fine and iterations 2+ clone **without** credentials and fail. The SSH path isn't affected because `self._ssh_private_key` is never nulled here. Fix: don't null it in `cleanup()` (or re-derive it at the start of each `run()`).

lib/utils.py
```diff
@@ -40,16 +40,21 @@ def get_git_api(parsed_url):
     if parsed_url.netloc == '' and '@' in parsed_url.path: # this could be an SSH git shorthand, we allow this but cannot determine API
         return [None, None]
 
-    if parsed_url.netloc in ['github.com', 'www.github.com']:
+    hostname = parsed_url.hostname or ''
```
ribalba, line 124: **🟠 Medium — `netloc`→`hostname` drops in-URL credentials from the API request**

Good fix for host/port matching, but `hostname` also strips any `user:pass@`. `get_repo_last_marker` (which raises `RuntimeError` on any non-200) is called with the credentialed URL for tag/commit schedules, so a private self-hosted GitLab repo submitted as `https://user:token@git.internal/g/r` now gets an unauthenticated 401/404 → `runs_add` re-raises as HTTP 422. It previously authenticated (creds were in `netloc`).

Related: `runs_add` stores the *encrypted* URL in `watchlist`, and `cron/watchlist.py:52` calls `get_repo_last_marker` on it **unguarded**, so the `RuntimeError` propagates and aborts the whole scheduling pass. Worth a `try/except` there regardless.

lib/scenario_runner.py
```diff
@@ -414,7 +462,17 @@ def _checkout_repository(self):
...
+                clone_uri = self._uri
+                if self.__uri_userinfo:
+                    userinfo = decrypt_data(self.__uri_userinfo) if self.__uri_userinfo.startswith(ENCRYPTED_VALUE_PREFIX) else self.__uri_userinfo
+                    _p = urlparse(self._uri)
+                    _host = _p.hostname or ''
+                    if _p.port:
+                        _host += f":{_p.port}"
+                    clone_uri = urlunparse((_p.scheme, f"{userinfo}@{_host}", _p.path, _p.params, _p.query, _p.fragment))
+
+                command.append(clone_uri)
```
ribalba: **🟠 Medium — decrypted git credentials exposed in the `clone` argv**

`clone_uri` is `https://user:pass@host/…` and goes into the `subprocess.run` argv, so while `git clone` runs the plaintext password is visible via `ps auxww` / `/proc/<pid>/cmdline` to any local user on the worker — this undoes the at-rest encryption at runtime. Prefer a credential helper / `GIT_ASKPASS` / `http.extraHeader` over inline-URL auth.

lib/scenario_runner.py
```diff
@@ -311,6 +339,26 @@ def _delete_ssh_private_key_file(self):
         if self._ssh_private_key_file.exists():
             self._ssh_private_key_file.unlink()
 
+    def _prepare_docker_credentials(self):
+        if not self._docker_credentials:
+            return
+
+        self._docker_config_dir.mkdir(exist_ok=True)
```
ribalba, line 370: **🟠 Medium — config.json written to a predictable path with a non-atomic create→chmod**

`_docker_config_dir` resolves to the fixed, world-traversable `/tmp/green-metrics-tool/docker_client_config`, created via `mkdir(exist_ok=True)`. Below, `config.json` is created empty at 0644, then `chmod(0o600)`, then the secret is written — a local user can pre-create/symlink the predictable path, or hold a read fd on the 0644 empty file before the secret lands. Use `os.open(path, O_WRONLY|O_CREAT|O_TRUNC, 0o600)` for an atomic secure create (the SSH-key file shares this pattern).

lib/scenario_runner.py
```diff
@@ -148,11 +153,22 @@ def __init__(self,
         else:
             raise ValueError('ssh_private_key must be of type SecureVariable')
 
+        if docker_credentials is None:
+            self._docker_credentials = None
+        elif isinstance(docker_credentials, list):
+            for cred in docker_credentials:
+                if not isinstance(cred.get('password'), SecureVariable):
```
ribalba, line 158: **🟡 Low — validate registry/username here, not mid-run**

`__init__` only checks that `password` is a `SecureVariable`; `_prepare_docker_credentials` later reads `cred['username']`/`cred['registry']` unchecked, so a malformed list fails with a `KeyError` **after** measurement has started and containers/folders are set up. Validate `registry`/`username` presence here so it fails fast at construction.

lib/scenario_runner.py
```diff
@@ -1230,7 +1294,10 @@ def _build_docker_images(self):
                 # bc the docker pull command does not stream the interactive progress bar. Only the lines when a layer finished with downloading
                 # Since this information does not provide info how long the image download still will take we opted for keeping this pull call less complex
                 # So you have to stare at the "Pulling XYZ" command until it is finished :)
-                ps_pull = subprocess.run(['docker', 'pull', service['image']], stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='UTF-8', errors='replace', check=False)
+                pull_env = os.environ.copy()
+                if self._docker_credentials and self._docker_config_dir.exists():
+                    pull_env['DOCKER_CONFIG'] = self._docker_config_dir.as_posix()
```
ribalba, line 1327: **🟡 Low — this bypasses the host's ambient docker auth**

Once user creds are set, `DOCKER_CONFIG` points at a config listing only the user's registries; docker does not merge it with `~/.docker/config.json`, so a pull from a registry the host operator `docker login`'d to (but the user didn't supply) now fails. Niche, but a behavior change worth noting.

lib/scenario_runner.py
```diff
@@ -1153,6 +1211,12 @@ def _build_docker_images(self):
                     '--mount', f"type=bind,source={self._build_dir.as_posix()},target=/output"]
                 )
 
+                if self._docker_credentials:
+                    docker_build_command.extend([
+                        '--mount',
+                        f"type=bind,source={self._docker_config_dir.joinpath('config.json').as_posix()},target=/kaniko/.docker/config.json,readonly",
```
ribalba, line 1245: **🟡 Low — base64 auth token is mounted into the kaniko build and isn't redaction-covered**

A malicious `RUN cat /kaniko/.docker/config.json` can echo the token into build logs, and `filter_sensitive_data` only matches URI-userinfo and PEM blocks, not base64 docker auth tokens. Mostly same-principal (the user's own creds building their own repo), so low — but if build logs can ever be public, consider a redaction rule for `"auth":"…"`.

runner.py
```diff
@@ -184,12 +185,26 @@
     else:
         ssh_private_key_contents = None
 
+    if args.docker_credentials:
+        with open(args.docker_credentials, 'r', encoding='UTF-8') as f:
+            raw_creds = json.load(f)
```
ribalba, line 190: **🟡 Low — malformed `--docker-credentials` file raises an unguarded `KeyError`**

The comprehension below does `c['registry']`/`c['username']`/`c['password']` with no key checks, so a hand-written JSON file missing a field dies with a bare `KeyError` instead of the clean `ValueError` raised elsewhere. Validate keys (ideally via a shared validator reused with `User.update_docker_credentials`).

ribalba (commented):
## Automated review — credential-handling focus

Reviewed the authoritative PR diff (19 files). The encryption envelope, `SecureVariable` wrapping, deletion of secret keys from `self._arguments`, the SSH-parallel temp-file lifecycle, and the encrypted-at-rest tests all look good. Inline comments below flag **2 fix-before-merge** issues and several hardening items; severity is marked on each.

**Out-of-scope heads-ups** — these live in code already on `main` (not in this PR's diff), so I couldn't anchor them inline, but they're real and worth a look:
- **XSS** — `frontend/js/timeline.js` ~L367: `escapeString(lines.push(currentLine))` escapes the array *length* and discards it, so the changelog tooltip's trailing line (and any message shorter than 40 chars) is injected **unescaped** into the ECharts tooltip HTML.
- **Null deref** — `frontend/js/helpers/metric-boxes.js` ~L388: `valueNode.getAttribute('title').replace(...)` throws when a key-metric card was first populated by a std-dev metric (which never sets `title`).

_🤖 Automated review via [Claude Code](https://claude.com/claude-code)_

ArneTR:
@ribalba Hmm ... not really the kind of review I was looking for :) Although Claude Code found some points I will give a closer look it creates some very hypothetical and IMHO abstract scenarios that just create work for me to reason about.
The pure code level quirks where after all aalready reviewed by CodeRabbitAI

I was more looking for something in the real of:
- How does the PR read to you? Is it clear what the code dos
- Did you find potential for refactorings? Is there something that will hurt maintainability in the long run
- Did you spot any conceptual / architectural issues

api/scenario_runner.py
```diff
@@ -958,16 +961,29 @@ async def runs_add(software: Software, no_url_check: bool = False, user: User =
...
+        try:
+            _userinfo = f"{_parsed_repo.username}:{_parsed_repo.password}" if _parsed_repo.password else _parsed_repo.username
```
ribalba: Why are you doing the `else username` here?

api/scenario_runner.py
```diff
@@ -958,16 +961,29 @@ async def runs_add(software: Software, no_url_check: bool = False, user: User =
...
+            _encrypted_userinfo = encrypt_data(_userinfo)
+        except EncryptionConfigurationError as exc:
+            raise HTTPException(status_code=422, detail='Cannot store URL credentials: encryption is not configured on this server') from exc
+        _host = _parsed_repo.hostname or ''
```
ribalba: Can the hostname be empty?

ribalba:
Doing the manual review now. Just let claude do the first run

lib/scenario_runner.py
```diff
@@ -193,8 +209,20 @@ def __init__(self,
...
-            del self._arguments['ssh_private_key']
+        # security related keys we never want to log
+        del self._arguments['ssh_private_key']
+        del self._arguments['docker_credentials']
+
+        # Strip HTTP basic auth credentials from URI; keep clean version for logging and DB storage
+        _parsed_uri = urlparse(self._uri)
+        if _parsed_uri.username:
+            self.__uri_userinfo = f"{_parsed_uri.username}:{_parsed_uri.password}" if _parsed_uri.password else _parsed_uri.username
+            _clean_host = _parsed_uri.hostname or ''
```
ribalba: What is the case for hostname to be empty?

lib/scenario_runner.py
```diff
@@ -193,8 +209,20 @@ def __init__(self,
...
-        if 'ssh_private_key' in self._arguments:
-            del self._arguments['ssh_private_key']
+        # security related keys we never want to log
+        del self._arguments['ssh_private_key']
+        del self._arguments['docker_credentials']
+
+        # Strip HTTP basic auth credentials from URI; keep clean version for logging and DB storage
+        _parsed_uri = urlparse(self._uri)
+        if _parsed_uri.username:
+            self.__uri_userinfo = f"{_parsed_uri.username}:{_parsed_uri.password}" if _parsed_uri.password else _parsed_uri.username
```
ribalba: This looks very much like the code in `api/scenario_runner.py‎` so much so I thought my comments didn't save.

lib/scenario_runner.py
ribalba: This might be relevant if we add some sort of docker caching

lib/user.py
```diff
@@ -61,6 +66,13 @@ def _save_key_to_db(self, encrypted_value):
             WHERE id = %s
             """, params=(encrypted_value, self._id, ))
 
+    def _save_docker_credentials_to_db(self, encrypted_value):
```
ribalba, line 69: We could add a check here if the value is really encrypted? Just to be sure

ribalba:
Didn't read the tests. Reads pretty much how we described it. Looks good except the few comments I had

Why did you add the functionality to add docker credentials in `change_settings` but don't add it to the frontend?

lib/scenario_runner.py
ArneTR: wontfix as we currently have no protection against this anyway. Fix would only be a patchwork

lib/scenario_runner.py
ArneTR: yes. Will try to keep a reminder

api/scenario_runner.py
ArneTR: bc a valid HTTP-Auth can also only contain a username.

According to claude it can also contain only a password which I did not know. Will refactor it to include that also

api/scenario_runner.py
ArneTR: It can. If the URL simply does not contain it. In the case of the API it will fail beforehand in the GIT check. In the case of the lib/scenario_runner.py it will fail after in the clone.

So this check is more of a type guard to not mangle with None.

lib/scenario_runner.py
ArneTR: It can. If the URL simply does not contain it. In the case of the API it will fail beforehand in the GIT check. In the case of the lib/scenario_runner.py it will fail after in the clone.

So this check is more of a type guard to not mangle with None.

lib/scenario_runner.py
ArneTR: good one. will try to refactor

lib/user.py
ArneTR: 👍

ArneTR:
@coderabbitai review

PR #1759, opened 2026-07-05
(feat): ErrorLog and Email Stream filter to redact HTTP-AUTH and PRivate Keys

ArneTR:
This PR was specifically crafted to not patch sys.stdout/sys.stderr as the overhead would be tremendous and the gain would only to have clean server logs. Server logs thus can in cases of a code oversight still contain unredacted passwords. DB is save though

Added sensitive-data redaction for HTTP-AUTH credentials and private keys across ErrorLog, email/job payloads, and URL/command logging.

- Implemented `lib.utils.filter_sensitive_data()` (with shared redaction markers) to strip URI credentials and PEM private keys.
- Applied redaction at output/serialization points (error formatting, run/scenario runner logs and stored command/stdout/stderr, cron/watchlist scheduling status, and runner URI echo) rather than monkeypatching `sys.stdout`/`sys.stderr`.
- Extended `/v2/jobs` to redact `jobs.url` in API responses while keeping the DB value unchanged for backend cloning.
- Refactored job insertion/routing to dedicated job classes with explicit `JOB_TYPE` (`RunJob`, `EmailSimpleJob`, `EmailReportJob`, etc.) and sanitized email/message fields before persistence; added URI filtering to watchlist “show”/status output.

Tests added/updated to verify redaction in stderr/system logs, DB-stored job/email message fields, API job URL responses, watchlist stdout, and relations checkout/run-log persistence, including new credentialed scenario YAML.

### DIFF
--- a/api/scenario_runner.py
+++ b/api/scenario_runner.py
@@ -23,7 +23,8 @@ from api.api_helpers import (CustomORJSONResponse, ORJSONResponseObjKeep, add_ph
 from lib.global_config import GlobalConfig
 from lib.db import DB
 from lib.diff import get_diffable_rows, diff_rows
-from lib.job.base import Job
+from lib.job.run import RunJob
+from lib.job.email_simple import EmailSimpleJob
 from lib.user import User
 from lib.watchlist import Watchlist
 from lib import utils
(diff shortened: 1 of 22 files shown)


ArneTR:
@coderabbitai review

ArneTR:
@coderabbitai review

PR #1761, opened 2026-07-06
Drawing Changelog Markline at actual point in time - Between data measurements

ArneTR:
Marklines where always rendered on the datapoints itself. This was due to the category nature of the axis.

This was usually fine if runs happened daily and thus changes in a day would be mapped to the event of the following day.

However with long pauses in the data it seemend unplausible that a change 3 months ago would have influenced a measurement way later.

Thus the rendering is now done graphically - More overhead but more accurate - In between the datapoints.
Screenshot 2026-07-06 at 12 37 35 PM
Screenshot 2026-07-06 at 12 37 30 PM

Updated timeline charts to render changelog markers between datapoints instead of snapping them to the nearest measurement. The changelog logic now builds segment-based marker data from timestamps, aggregates messages per segment, and redraws the markers as graphic overlays on resize and data zoom.

Also replaced the old `markLine`-based “Cluster Changelog” series with a custom dashed line plus invisible tooltip hit area, clipped to the visible zoom window for accurate placement.

### DIFF
--- a/frontend/js/timeline.js
+++ b/frontend/js/timeline.js
@@ -4,6 +4,9 @@ let repository_uri = null; // Store unescaped URI for URL construction
 window.onresize = function() { // set callback when ever the user changes the viewport
     chart_instances.forEach(chart_instance => {
         chart_instance.resize();
+        // graphic elements (changelog markers) are positioned in pixels, so they need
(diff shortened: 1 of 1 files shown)


ArneTR:
@coderabbitai review

PR #1762, opened 2026-07-07
(fix): Overflow when computing weighted average. Swapping to early cast for anyway used Double Precision

ArneTR:
Adjusted the weighted-average calculation in `lib/phase_stats.py` so `diff` is cast to `DOUBLE PRECISION` before the `SUM(...)`, avoiding overflow by promoting the arithmetic earlier rather than after aggregation.


ArneTR:
@coderabbitai review

PR #1770, opened 2026-07-09
Removes the error message as this always came up

ribalba:
On a fresh install, with the default config, this error would be shown 8 times. Removed it as it doesn't really help and we don't want a default carbon provider enabled as default.


ribalba:
Closing this in favour for always enabling the carbon static provider

PR #1772, opened 2026-07-10
(fix): Test path used unstable regex

ArneTR:
[test-path: smoke_test.py]


ArneTR:
@ribalba 

You can now run again reduced test suites by putting [test-path: FILE FILE2 ....] in your commit message

PR #1775, opened 2026-07-14
Upgrading postgres memory parameters to 8 GB reasonable machine

ArneTR:
After running GMT in production for a while we see that some background jobs experience strong memory pressure and makes query partials spill to disk.

This PR addresses this by migrating from mini-developer-machine defaults to reasonable 8 GB dedicated defaults.

We assume every developer machine nowadays has at least 8 GB of memory and is not fully saturated when running GMT. Thus these configs should work everywhere out of the box and provide more long-term stability when deploying GMT in cluster mode without ever tuning configs by hand.

## Summary

- Increased PostgreSQL memory settings in `docker/compose.yml.example` for 8 GB machines.
- Updated `shared_buffers`, `work_mem`, and `maintenance_work_mem` to reduce disk spills and improve background job stability.

### DIFF
--- a/docker/compose.yml.example
+++ b/docker/compose.yml.example
@@ -27,8 +27,10 @@ services:
       -c ssl=on
       -c ssl_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
       -c ssl_key_file=/etc/ssl/private/ssl-cert-snakeoil.key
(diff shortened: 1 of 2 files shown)


ArneTR:
@coderabbitai review

PR #1776, opened 2026-07-14
(rework): CarbonDB worker jobs now use 60 day window; Optimized backfill queries

ArneTR:
We are now using a 60 day window as we have seen errors when processing carbondb sometimes when backfill apis are not available for longer than 24 hours.

to not miss data while copying and compressing we now add an internal 30 buffer / grace period

## Summary

- Expands CarbonDB backfill/processing windows from ~24 hours to **60 days** and uses a **30-day grace/buffer** to prevent missing data when backfill/copy APIs lag.
- Refactors **carbon intensity** and **geo** backfills into a two-phase flow: run **DB-cache updates** for a computed set of missing rows, then manually fill only the **remaining** rows still missing values.
- Optimizes backfill queries by passing **explicit row ID lists** into cached update helpers (and simplifying the “missing” fetch helpers accordingly) rather than relying on broader null filters.
- Limits **compression** and related transformations to the **last 60 days** by changing the source query to a 60-day time window.
- Widens CarbonDB copy/merge/dedup operations to the **60-day** window (function defaults, constraint validation, and dedup cutoff logic updated).
- Updates `docker/structure.sql` with **new/adjusted indexes**, including partial indexes for missing geo backfills and a new `carbondb_data_raw(time, type, user_id)` index to improve performance (including compression-related access patterns).

### DIFF
--- a/cron/backfill_carbon_intensity.py
+++ b/cron/backfill_carbon_intensity.py
@@ -9,8 +9,6 @@ import os
 import json
 import fcntl
 
-CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
-
 import requests
 from cachetools import cached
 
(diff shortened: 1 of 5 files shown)


ArneTR:
@coderabbitai review

ArneTR:
@coderabbitai review

PR #1777, opened 2026-07-15
(fix): Phase padding will only happen if phase had at least one sampl…

### LINKED ISSUE #1774: Runs in GMT issue warning in every run
Currently runs issue 'Very few samples (< 50% of observed duration or < 2) encountered in phase '[BOOT]', MEAN values might be inaccurate' in every run.

This seems to be related to the new phase padding.

An unrelated sample seems to get mapped into the Boot phase which is typically zero (starting providers in boot phase - cgroup ones)

Needs further investigation.

For people experiencing this issues: To our current understanding it is more confusing than critical. NO values in other phases are affected.

ArneTR:
…e present

We had the bug that the metric providers for cgroup, which come up later, get a metric mapped into their phase from the data coming after. This is misleading and should only be the case if the come up so early that they actually capture data from the phase.

With a long sleep it can be that the mapping for the next tick is multiple seconds away. Guarding that at least one sample occured in the phase helps as due to sampling rate watching by the provider itself we know that samples cannot be more than the expected sample rate apart from each other.

Closes https://github.com/green-coding-solutions/green-metrics-tool/issues/1774

## Summary

Updated phase metric aggregation to apply padding only when a phase contains in-phase samples. This prevents late-starting cgroup metrics from being incorrectly attributed to earlier phases and avoids false `[BOOT]` too-few-samples warnings.

## Changes

- Added in-phase sample tracking through SQL aggregation and derivative computation.
- Skipped result handling when phases have no in-phase samples.
- Updated low-sample warning text.

### DIFF
--- a/lib/phase_stats.py
+++ b/lib/phase_stats.py
@@ -135,14 +135,20 @@ def build_and_store_phase_stats(run_id, sci=None, sci_metrics=None):
 
         select_query = """
             WITH in_range AS (
-                SELECT time, value
+                SELECT
+                    time,
+                    value,
+                    1 as in_phase -- a dummy marker to tell us if the later joined samples are from inside our outside the phase boundary
                 FROM measurement_values
                 WHERE measurement_metric_id = %s
                   AND time > %s
                   AND time < %s
             ),
             next_one AS (
-                SELECT time, value
+                SELECT
+                    time,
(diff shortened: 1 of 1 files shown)


ArneTR:
@coderabbitai review

PR #1778, opened 2026-07-15
(feat): Power Hog now also does not accept data anymore from older than 30 days

ArneTR:
Technically it could also accept older values, but to date there is no valid use case for it.

Having this in place helps us maintain the 30 day merge window for CarbonDB way easier as we can blindly copy over

## Summary

- Enforces a 30-day merge/timestamp acceptance window for Power Hog data on `POST /v2/hog/add`, rejecting timestamps older than the limit or in the future with HTTP 422 (and including the submitted timestamp in the error).
- Refactors the `/v2/hog/add` database write path to use a single `DB().transaction_cursor()` transaction for inserting into `hog_simplified_measurements` and `hog_top_processes` (removing the prior `query_multi`-based insertion pattern).
- Updates `tests/api/test_api_hog.py` with `make_hog_payload(measurement_timestamp)` and corrects the existing payload compression; adds tests for outdated (422), boundary/valid (202), and future (422) timestamps.
- Removes `DB.query_multi` and replaces it with `DB.transaction_cursor` in `lib/db.py`.


ArneTR:
@coderabbitai review

ArneTR:
@coderabbitai review

PR #1779, opened 2026-07-16
Allows for grouping of error messages

ribalba:
We often have the same error repeating in `/cluster-status.html`. Now you can group them and delete them together. 

Screenshot 2026-07-16 at 15 23 06

### DIFF
--- a/frontend/cluster-status.html
+++ b/frontend/cluster-status.html
@@ -84,6 +84,15 @@
 
         <h3>System Logs</h3>
         <div id="system-logs-section">
+            <div class="ui form" style="margin-bottom: 1em;">
+                <div class="inline field">
+                    <div class="ui toggle checkbox">
+                        <input type="checkbox" id="system-logs-group-toggle">
+                        <label for="system-logs-group-toggle">Group similar messages</label>
+                    </div>
+                    <span data-inverted data-tooltip="Collapses entries whose message differs only in run specific data like Run-ID, name or filename. Click the counter to expand a group."><i class="ui question circle icon fluid"></i></span>
+                </div>
+            </div>
             <table id="system-logs-table" class="ui celled striped table"></table>
         </div>
         <div id="system-logs-no-access" class="ui info message" style="display: none;">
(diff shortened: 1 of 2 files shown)


ArneTR:
I am a bit sceptical of that feature.

At the moment the grouping happens on the parameters passed to the exception. I remember many cases where for similar runs we had similar error parameters but only the traceback really showed the differences of where they occured. For instance a subprocess call can fail in different places - Only key points are logged in the error directly - The traceback shows that it happened in different places.

2. Also I am not clear how it works on the frontend:
- When I clear the error. Will multiple be deleted?
- Since DataTables pages the table. Will the grouping work for the full result or for all results in the whole table?
- Currently the API only returns 20 results at a time. So to leverage grouping really we would have to return all rows which   makes loading a bit laggy on very big error logs

3. The other issue I have is that I find it very useful to see how errors are distributed over time. So I see myself clicking in 30-50% of the cases on the expand button anyway.

4. I feel for this functionality the code is also quite complex.
5. 
6. What is the real use case you are trying to solve?

I for instance sometimes have cases where I know that one particular error has really spammed the logs. I then connect to the DB and issue a call like this:

`DELETE FROM system_logs WHERE message LIKE '%New run added from Web Interface%';`

If you encounter this would it not be rather interesting to have a "Clear Log by regex" field where you can quickly prune the log with a definitive message that you know is unique to the error type you are seeing?

This would reduce the complexity quite a bit at least ...

Let me know.

## Final words
In general if you feel the feature is very useful to you I am happy to bring it in. But would have it toggle on/off. State saved als localstorage per user.

ribalba:
A lot to answer here. So the problem I have is that I submit 20 runs and then 5 fail because of some configuration error. Right now we have 6 out of 11 errors being the same. The rest can also be grouped. So the error list is mostly duplicated entries. 

- In the current implementation if you delete a grouped entry it will delete all
- There is a toggle which is deactivated per default and you need to switch it on to group when you see a lot of the same errors. 
- The 20 error restricition is a problem as I only do the grouping in the frontend 

As you are the person using this 99% of the time I would vote for leaving this out if you don't find it useful.

PR #1781, opened 2026-07-17
Fix/audit findings

ribalba:


lib/system_checks.py
```diff
@@ -326,12 +326,16 @@ def _check_rapl_domain(domain_key):
 
     expected_uw = int(expected_watts) * 1_000_000
     for entry in domain_entries:
+        checked_any = False
         for limit_key in ('long_term_uw', 'short_term_uw'):
             value = entry.get(limit_key)
             if value is None:
-                return False  # this constraint type not exposed on this domain — skip it
+                continue  # this constraint type not exposed on this domain — skip it
```
ribalba, line 333: @ArneTR Ai fond this and it looks logical but I am not sure as it does change the beahviour

metric_providers/cpu/time/cgroup/system/provider.py
ArneTR, line 1: @ribalba Is there a reason for this change?

The provider is clearly by name a child of the cgroup providers. Why remove it?

ArneTR (changes requested):
Patched some stuff in prune_db. 

Unclear about the metric provider intent. Made a comment

Rest looks good.

The patch for system_checks is ok. Solves an edge case that is IMHO not possible. But does not hurt to have it

metric_providers/cpu/time/cgroup/system/provider.py
ribalba: This is a classical AI review problem. When looking at this in my VSCode it looked like it changed it from a `BaseMetricProvider` to a `CgroupMetricProvider` and I was like "That makes sense as it is clearly a cgroup provider." But I only looked very briefly. So this clearly is wrong. I will revert

PR #1789, opened 2026-07-22
Cpu utilization windows native

ribalba:
This is a branch I copied from 

https://github.com/MaximilianJahns/green-metrics-tool_rapl_metric_provider/tree/cpu-utilization-windows-native

and this was the PR

https://github.com/green-coding-solutions/green-metrics-tool/pull/1757

as I couldn't write my fixes to Maxes branch I pushed it here. 

@ArneTR this seems to work under windows now and I could test it. 

@MaximilianJahns Thank you for all the work

## Summary

- Added native Windows CPU utilization providers using Win32 `GetSystemTimes()` and NTAPI `NtQuerySystemInformation`.
- Added sampling interval validation, Windows timer-resolution handling, GMT timestamps, and optional interrupt/DPC metrics.
- Added Windows provider wrappers and MSVC build scripts.
- Consolidated Windows native provider compilation into one parameterized installer function.
- Renamed providers to `cpu_utilization_win32_system` and `cpu_utilization_ntapi_core`.
- Updated configuration, metric mappings, PSU providers, phase statistics, and default settings.
- Added Windows software-view activation options to the installer.


ArneTR:
@coderabbitai review

frontend/js/helpers/config.js.example
```diff
@@ -572,6 +572,16 @@ METRIC_MAPPINGS = {
         "source": "mach",
         "explanation": "CPU Utilization of total system"
     },
+    "cpu_utilization_windows_system": {
+        "clean_name": "CPU %",
+        "source": "Windows",
```
ArneTR: Here it would be nice to see the actual source in the system. For linux we for instance do not list "linux" as the source, but the subsystem where we capture the metric, for instance "procfs".

lib/phase_stats.py
```diff
@@ -267,6 +267,8 @@ def build_and_store_phase_stats(run_id, sci=None, sci_metrics=None):
                 'lmsensors_fan_component',
                 'cpu_utilization_procfs_system',
                 'cpu_utilization_mach_system',
+                'cpu_utilization_windows_system',
```
ArneTR: same argument here: use subsystem if possible

lib/phase_stats.py
```diff
@@ -284,7 +286,7 @@ def build_and_store_phase_stats(run_id, sci=None, sci_metrics=None):
             ):
                 csv_buffer.write(generate_csv_line(phase['hidden'], run_id, metric, detail_name, f"{idx:03}_{phase['name']}", value_avg, 'MEAN', max_value, min_value, sampling_rate_avg, sampling_rate_max, sampling_rate_95p, unit))
 
-                if metric in ('cpu_utilization_procfs_system', 'cpu_utilization_mach_system'):
+                if metric in ('cpu_utilization_procfs_system', 'cpu_utilization_mach_system', 'cpu_utilization_windows_system'):
```
ArneTR: where is cpu_utilization_windows_core added?

metric_providers/cpu/utilization/ntapi/core/provider.py
```diff
@@ -0,0 +1,55 @@
...
+        super().__init__(
+            metric_name='cpu_utilization_windows_core',
+            metrics={'time': int, 'value': int, 'detail_name': str},
+            sampling_rate=sampling_rate,
+            unit='Ratio',
+            current_dir=os.path.dirname(os.path.abspath(__file__)),
+            metric_provider_executable='metric-provider-binary.exe',
+            skip_check=skip_check,
+            folder=folder,
+            disable_buffer=False,
```
ArneTR, line 26: why not disabling the buffer? How is windows default buffering? we would need to rewrite the base.py which currently uses stdbuf ... but all in all I see this also helpful for windows

config.yml.example
```diff
@@ -271,13 +271,21 @@ measurement:
...
-        #   cpu_gpu: False
-        #   dram: False
+#      cpu_energy_rapl_scaphandre_component:
+#        sampling_rate: 99
+    #--- Optionally disable individual RAPL domains (cpu_package, cpu_cores, cpu_gpu, dram, psys)
+#         domains:
+#           cpu_gpu: False
+#           dram: False
+    #--- CPU Utilization - System-wide CPU load via GetSystemTimes()
+#      cpu_utilization_windows_system:
```
ArneTR: I would activate this by default

metric_providers/cpu/utilization/win32/system/source.c
```diff
@@ -0,0 +1,172 @@
...
+#include <timeapi.h>    /* timeBeginPeriod / timeEndPeriod */
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include <limits.h>
+
+/* ---- CLI arg parsing (identical pattern to RAPL provider) ---- */
+static unsigned int parse_int(char *argument)
```
ArneTR, line 19: Can windows not import our libraries from lib/c ? Maybe compile them separately as windows importable object files?

metric_providers/cpu/utilization/win32/system/source.c
```diff
@@ -0,0 +1,172 @@
...
+{
+    LARGE_INTEGER qpc;
+    QueryPerformanceCounter(&qpc);
+    return qpc.QuadPart;
+}
+
+/* ---- CPU time reading (this replaces read_cpu_proc() from procfs) ---- */
+typedef struct {
+    uint64_t idle;
+    uint64_t kernel;   /* NOTE: kernel time INCLUDES idle time, same caveat as procfs system_time */
```
ArneTR, line 86: can we link a documentation here where it is said that Windows kernel only tracks these three counters. 

And what exactly is the caveat that system time includes idle time?

metric_providers/cpu/utilization/win32/system/source.c
```diff
@@ -0,0 +1,172 @@
...
+        cpu_time_t curr;
+        read_cpu_times(&curr);
+
+        /* same compute/non_compute split logic as procfs, mapped to Windows fields */
+        uint64_t total_delta = (curr.user - prev.user) + (curr.kernel - prev.kernel);
+        uint64_t idle_delta  = curr.idle - prev.idle;
+        uint64_t busy_delta  = (total_delta > idle_delta) ? (total_delta - idle_delta) : 0;
+
+        uint64_t ts_us = now_us(&clock);
+        long value = (total_delta > 0) ? (long)((busy_delta * 10000ULL) / total_delta) : 0;
```
ArneTR, line 239: Why is the multiplication with 10,000 ? What is the unit of the delta? I would have expected a multiplication with 1,000 or 1,000,000 but 10,000 is suprising

metric_providers/cpu/utilization/win32/system/source.c
```diff
@@ -0,0 +1,172 @@
...
+    while (1) {
+        LONGLONG deadline = now_qpc() + (LONGLONG)(interval_ms * qpc_ticks_per_ms);
+
+        cpu_time_t curr;
+        read_cpu_times(&curr);
+
+        /* same compute/non_compute split logic as procfs, mapped to Windows fields */
+        uint64_t total_delta = (curr.user - prev.user) + (curr.kernel - prev.kernel);
+        uint64_t idle_delta  = curr.idle - prev.idle;
+        uint64_t busy_delta  = (total_delta > idle_delta) ? (total_delta - idle_delta) : 0;
```
ArneTR, line 235: I have a hard time understanding this calulation and think this will be hard to maintain.

I would expect:
user+kernel/(idle+kernel+user) 

But this seems to user+kernel-idle ...?

ArneTR: if this is related to the fact that kernel contains idle I think it would be cleaner to create a transient var: "true_idle" or "true_system" or "system_no_idle" or something

ArneTR (commented):
Thanks for the PR @ribalba and @MaximilianJahns 

I put most of my comments in the Review directly attached to code lines.

Here some extras:

- Was this provider tested on low -i ?  What is the minimum floor this provider can handle?
- In the Linux ones we introduced a minmum floor from which it makes no sense anymore and can actually lead to wrong averages (bc of the way we calculate averages). Maybe it makes sense to do that also here? Would be nice to have something from the docs we can link or something empirical.
PR for reference: Minumum Floor: https://github.com/green-coding-solutions/green-metrics-tool/commit/be78f8d9ecc8ebdcf53f54efb00582950bc01440

ArneTR:
Please also have a look at the NitPick comments from Coderabbit. They look useful

metric_providers/cpu/utilization/windows/system/source.c
```diff
@@ -0,0 +1,172 @@
...
+
+        uint64_t ts_us = now_us(&clock);
+        long value = (total_delta > 0) ? (long)((busy_delta * 10000ULL) / total_delta) : 0;
+
+        printf("%llu %ld\n", (unsigned long long)ts_us, value);
+
+        prev = curr;
+
+        LONGLONG remaining_qpc = deadline - now_qpc();
+        double sleep_ms = (remaining_qpc > 0) ? (remaining_qpc / qpc_ticks_per_ms) : 0.0;
```
ArneTR: I also wanted to flag this. This sleep behaviour is very different to the code we have in the other GMT providers.

What is the reason for implementing it that differently?

The benefits we have from our current version should then somehow guaranteed another way:
- Currently we put fixed sleeps and check resulting total sample times. These are checked to not execeed a 20% overhead. This guarantees for us that the provider itself does not consume excessive CPU ticks and we can safely run quite some in parallel

When calculating time differences in the provider itself we can run into situations where the provider can run continously and almost never sleep. So overhead must be checked inline.

I would argue for simplifying the C code as by requirement we anyway can have no provider that takes longer than 1-5 ms for processing the capture loop anyway. It would create noise on the measurement.

ArneTR (changes requested):

frontend/js/helpers/config.js.example
max123654: Updated both providers to use the concrete subsystem instead of the OS name — win32 for the system-wide provider (GetSystemTimes(), documented Win32 API) and ntapi for the per-core provider (NtQuerySystemInformation(), undocumented native NT API), consistent with how Linux uses procfs/cgroup rather than linux. Re: the same comment on phase_stats.py — that block only lists the metric name for aggregation purposes, there's no source field there to rename; the name itself (cpu_utilization_windows_system) stays as-is, same as cpu_utilization_procfs_system did.

metric_providers/cpu/utilization/ntapi/core/provider.py
MaximilianJahns: Currently has no effect for this provider — disable_buffer is only consumed inside BaseMetricProvider.start_profiling() (wraps the command with stdbuf -o0, Linux-only), but both Windows providers override start_profiling() entirely and never call super().start_profiling(). The actual unbuffered-output requirement is instead handled directly in source.c via setvbuf(stdout, NULL, _IONBF, 0), which is portable and doesn't depend on an external tool.

config.yml.example
MaximilianJahns: done

metric_providers/cpu/utilization/win32/system/source.c
MaximilianJahns: gmt-lib.c/gmt-lib.h use POSIX-only headers (unistd.h, clock_gettime, access()), which aren't available under MSVC. The existing RAPL provider already established the pattern of not using lib/c on Windows and reimplementing the small set of needed helpers directly — we followed that for consistency. There are currently three Windows providers affected (RAPL, and these two). Splitting out a Windows-compatible subset (arg parsing + timing helpers, roughly what we already have as parse_int() and clock_state_t/clock_init()/now_us()) as a shared gmt-lib-windows.c looks like a reasonably contained follow-up — touches all three existing Windows providers though, so we'd suggest it as a separate PR rather than folding it into this one.

metric_providers/cpu/utilization/win32/system/source.c
MaximilianJahns: This isn't a time-unit conversion — it's a fixed-point precision choice for the ratio, unrelated to the FILETIME 100ns units used elsewhere. busy_delta / total_delta is a value in [0,1]; multiplying by 10000 instead of 100 encodes it as an integer with 0.01% precision (matching the existing cpu_utilization_procfs_system convention on Linux) instead of losing precision to whole-percent integer rounding.

metric_providers/cpu/utilization/win32/system/source.c
MaximilianJahns: GetSystemTimes() is documented by Microsoft to return exactly three values: idle, kernel, and user time (https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getsystemtimes). The "kernel includes idle" caveat: per the same documentation and confirmed against psutil's own Windows implementation (which applies the identical kernel - idle correction), Windows' kernel-mode time accounting counts idle-thread execution as kernel time — there's no separate "system-busy" counter the way Linux's /proc/stat separates system from idle. We've made this explicit in code now via named intermediate variables (kernel_delta, user_delta, idle_delta) plus an inline comment. I changed it to be better readable.

metric_providers/cpu/utilization/windows/system/source.c
MaximilianJahns: Good point, agreed — simplified to a fixed Sleep(interval_ms) after each snapshot, same as the other providers, in both source.c files. The deadline-compensation logic was carried over from the RAPL provider pattern without re-evaluating whether it was actually needed here; you're right that it isn't, given the measurement itself is just a syscall (~1-5ms). Verified afterwards that intervals stay within ~2-6ms of the target with the simplified version, well inside the 20% tolerance check.

lib/phase_stats.py
MaximilianJahns: cpu_utilization_machine is a single scalar used downstream (e.g. by the XGBoost energy provider). cpu_utilization_windows_core emits 8 separate per-core values per tick from a single process call; including it in that same assignment would mean each core's value overwrites the previous one in turn, silently keeping only the last core processed. cpu_utilization_windows_system already provides the single system-wide value that's the correct semantic fit for cpu_utilization_machine. Happy to add a short comment at that line to make this explicit for future readers, if useful.

MaximilianJahns:
@ribalba If you want, feel free to copy this branch again for a merge... see the comments below for what I changed, hope you agree with the approach

metric_providers/cpu/utilization/ntapi/core/provider.py
ArneTR: I foresee this something tricky to debug when code is copied. In the current case if the C code contains a guard that disables buffering it is sufficient.
But if we ever use shell scripts or non-C callers in Windows this will fail. 

So my question: Is there a CLI way for windows to disable buffering when redirecting to files?

lib/phase_stats.py
ArneTR: I don't understand this. GMT supports having multiple detail_names per metric. WE have this for instance with containers / network interfaces etc. Is that not the same case here too?

Does this mean we currently do not have cpu_utilization_windows_core as phase_stat?

MaximilianJahns: cpu_utilization_windows_core is present as a phase_stat — it's already in the outer allowlist a few lines above the one in your diff, so it goes through the same detail_name-per-row thing as containers/network interfaces: 8 MEAN rows (core_0...core_7), one per core.

This specific line is just doing something different: cpu_utilization_machine is a single number, used downstream by things like the XGBoost energy provider, which just wants one system-wide utilization value. I left cpu_utilization_windows_core out of this one on purpose — since it produces 8 values per tick instead of 1, stuffing it into this single variable would just mean whichever core gets processed last silently "wins" and overwrites the rest.

One alternative I considered: instead of a single variable, use 8 separate variables (or a dict keyed by core). I didn't go that way because nothing actually consumes per-core values there right now — and the per-core data is already fully available anyway, just query the detail_name rows in phase_stats. Adding 8 parallel variables would just be duplicating data I already have in a cleaner form, without anyone actually using it.

metric_providers/cpu/utilization/ntapi/core/provider.py
MaximilianJahns: stdbuf is Linux-specific — it works via LD_PRELOAD, intercepting setvbuf() in the child process before main() even runs. Windows has no equivalent for arbitrary executables, since it doesn't have that kind of interposition mechanism built into the OS.

I searched a bit and one possible method could be ConPTY (CreatePseudoConsole, Windows Pseudoconsole API, available since Windows 10 1809). It's a pseudo console — making a child process think it's attached to a real terminal instead of a redirected file, which would trigger Windows' default unbuffered behavior automatically, no setvbuf() needed in the child at all.

But it has some real drawbacks:

ConPTY makes the child act like it's on a real terminal, so it can inject VT/ANSI escape sequences (cursor control, colors) into the output — those would need to be stripped before GMT's line parser sees them, risks silently corrupting data if done wrong.

There's a known Microsoft issue where CREATE_NEW_PROCESS_GROUP (already used via host_platform.terminate_process_group() for clean shutdown) doesn't play well with ConPTY-hosted processes — could break the shutdown path that works fine right now.

ribalba:
@ArneTR is this good to go?

ArneTR:
Yes, good to go.

All the arguments from Max were very reasonable. I do not see implementing the stdbuf for Windows with the fake console atm.

However I would love a separate shared gmt-lib-windows.c in a new PR as refactoring at some point

MaximilianJahns:
give me a minute. i guess i havent changed the provider names from windows to win32 for the system provider  and ntapi for the core . i would liek to fix that. sry have to get cleaner with my git code. will try to get more tidy

MaximilianJahns:
okay i changed the names to more specific ones then windows and made a run with gmt. i also changed it in js.config to avoid a clean-name tag in the frontend. hope that is fine now

ribalba:
@MaximilianJahns Is this ready to merge? I now pulled in your changes and fixed the two things that were forgotten. the sampling floor and the merge of build functions. Can you please check everything and then report if we can merge this.

ArneTR:
@coderabbitai review

config.yml.example
```diff
@@ -273,10 +273,17 @@ measurement:
     #--- RAPL - Requires the ScaphandreDrv kernel driver to be installed and running.
       cpu_energy_rapl_scaphandre_component:
         sampling_rate: 99
```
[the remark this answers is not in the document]
ribalba: Good point @MaximilianJahns can you please disable it

metric_providers/cpu/utilization/win32/system/source.c
```diff
@@ -0,0 +1,231 @@
...
+ * GetSystemTimes() is the only system-wide CPU time accounting the Win32 API
+ * exposes and it reports exactly three counters - idle, kernel and user:
+ * https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getsystemtimes
+ *
+ * Per that documentation the kernel counter "includes all threads in all
+ * processes, in kernel mode" - and the idle thread runs in kernel mode, so
+ * kernel time already contains idle time. Windows has no separate
+ * "kernel-busy" counter the way /proc/stat separates system from idle on
+ * Linux, so busy time has to be derived as (kernel + user) - idle. psutil's
+ * Windows implementation applies the same correction.
```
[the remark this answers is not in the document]
ribalba: I think we don't need to take into account a machine with more then 64 cores for Windows. @MaximilianJahns can you please add a check if we are over the 64 we should error.

config.yml.example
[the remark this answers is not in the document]
max123654: yes i will fix it

metric_providers/cpu/utilization/win32/system/source.c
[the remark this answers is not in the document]
MaximilianJahns: ok now it i added an error handling for 64 systems

config.yml.example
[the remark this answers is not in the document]
MaximilianJahns: done

ribalba:
@ArneTR @MaximilianJahns I have now addressed the last open points and merged with @MaximilianJahns branch. This should now be ready for final review and merge

PR #1803, opened 2026-07-29
when executing a shell command now fail on everything

ribalba:
---
<a href="https://app.blacksmith.sh/green-coding-solutions/codesmith/green-metrics-tool/pr/1803"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787941822&installation_model_id=428550&pr_number=1803&repository=green-coding-solutions%2Fgreen-metrics-tool&return_to=https%3A%2F%2Fgithub.com%2Fgreen-coding-solutions%2Fgreen-metrics-tool%2Fpull%2F1803&signature=1ee283680a70bd435906b10ebdaff476b9d46e9b78ceb07ba0c621dad2ae071f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


.gitignore
```diff
@@ -48,3 +48,5 @@ gha-creds-*.json
 # Window executables
 *.exe
 *.obj
+*.backup
```
ArneTR: this seems outdated. All backup files are now stored in central .config_backups which is already ignored.

Please delete your old backup files with git clean -f if still present and remove this

lib/scenario_runner.py
```diff
@@ -2322,7 +2322,10 @@ def _run_flows(self):
 
                         if shell := cmd_obj.get('shell', False):
                             docker_exec_command.append(shell)
```
ArneTR, line 2391: good call. But we have more than one -ec call in the lib/scenario_runner.py

Why not use this for all occurences?

lib/scenario_runner.py
```diff
@@ -2322,7 +2322,10 @@ def _run_flows(self):
...
-                            docker_exec_command.append('-ec')
+                            docker_exec_command.append('-o pipefail')
```
ArneTR: According to ChatGPT not all shells support pipefail I just noticed :/

I still feels this makes life tremendously easier for us. Maybe we just force it?

In every container that I have played around where there is a shell, there is also a shell that supports pipefail. Example: Ubuntu sh does not support pipefail but has bash, alpine has ash etc.

Shall we:
- Always set the flags and parse the error for "illegal option" ? 
- Shall we not set it on sh shell and issue a warning that some subtle errors cannot be caught?
- Shall we probe first before every call? (see quite some overhead here) ...
- Shall we forbid calling sh, unless an extra unsecure flag is set?

ArneTR (changes requested):

lib/scenario_runner.py
ribalba: 1) when running setup commands this always felt like a one commenad sort of thing. But we could also add it here
2) we control the playwright so there is no pipe there

lib/scenario_runner.py
ribalba: This is a good one. Especially as the most numbers for shell usage are desktop usage and this will be different to servers. We could also include an option called something like `shell_options` that we put before the `-c`. Just another idea. 

I do think a lof our jobs work on `sh` in Debian so  very good catch. 

I would probably add it and catch the error and then introduce a new flag that disables it.

Screenshot 2026-07-30 at 09 21 21

rp330114-del:
Pm disable-user --user 0 com.sec.enterprise.knox.cloudmdm.smdms0 com.

lib/scenario_runner.py
rp330114-del: pm disable+user -- 0 com.sec.knox.kccagent 0

ArneTR: But there is also the setup-commands which can call stuff in a shell

lib/scenario_runner.py
ArneTR: Seems like support is actually better than expected.

if we really feel that many more shells need support we can have custom injects for tcsh etc.

But I really like the idea of having an optional argument "shell options" that can be overridden. It defaults to `-o set pipefail -o ...`
I would argue this works in 99% of the cases and we can catch the error on top and tell the user to rather use a a different shell

ribalba:
@ArneTR I have now implemented the default shell options. Once this is ok I will edit the documentation

lib/scenario_runner.py
```diff
@@ -60,6 +60,47 @@
...
+# Options that are set for every command that is run through a shell.
+# They make the shell fail on the first error, on unset variables and on errors in a pipe.
+# Can be overridden per command with the 'shell-options' key in the usage_scenario.yml
+DEFAULT_SHELL_OPTIONS = ('-o', 'errexit', '-o', 'nounset', '-o', 'pipefail')
+
+# Not every shell supports all of the options we set by default (POSIX only mandates errexit and nounset).
+# Since we cannot probe every shell before every call we rather run into the error and then give a helpful hint.
+SHELL_OPTION_ERROR_MARKERS = ('illegal option', 'invalid option', 'unrecognized option', 'unknown option', 'bad option')
+
+def get_shell_options(cmd_obj):
```
ArneTR: I would move these two functions get_shell* and the two constants to lib.process_helpers

They seem to have the same context and do not pollute the lib.scenario_runner then

ArneTR (changes requested):
LGTM. Just had a remark regarding where to place the functions.

Good to merge after then

lib/scenario_runner.py
ribalba: done

ArneTR:
Merging main in with Pipeline fixes. Let's see ...

ArneTR:
ok, lets bring it in!

@ribalba Please make a docs update after

PR #1830, opened 2026-08-19
Increases the playwright timeout

ribalba:
As we now allow more complex playwright scripts I would want to change the way we handle timeouts. The
```
const contextOptions = {
  viewport: { width: 1280, height: 800 },
  ignoreHTTPSErrors: true, // <--- disables SSL check as we funnel requests through proxy
  timeout: 5000,
};
```
`timeout` here didn't seem to have an effect.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,6 +37,10 @@ The canonical local stack definition is `docker/compose.yml`.
 
 - Usage scenario syntax is enforced in `lib/schema_checker.py`.
 - The system distinguishes Git / URL measurements from local folder measurements via `uri_type`.
+- Usage scenario variables named `__GMT_VAR_SECRET_*__` hold secrets. They are encrypted on submission
+  (`lib/utils.py::encrypt_secret_usage_scenario_variables`), stored and displayed only in that encrypted
+  form, and decrypted just-in-time by the ScenarioRunner. Their plaintext must never reach the DB, the
+  API, or the frontend - see `lib/utils.py::register_sensitive_values` for the redaction safety net.
 - Many frontend tables consume positional SQL result arrays, so changing backend `SELECT` column order can silently break the UI.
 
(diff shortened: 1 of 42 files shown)


ribalba:
Would also be ok for me to just increase this to 5 min and not the 24 hours. Or introduce a new playwright timeout paramter

ArneTR:
1. timeout in the contextOptions indeed seems useless. Unclear why it is there. Can go
2. Mapping playwright timeout tot self._measurement_flow_process_duration makes quite some sense. But on the other hand it will absurdely clog our cluster with very long jobs if something hangs as all runs currently go through free users with no quota
3. Setting the navigation timeout to 1500 makes it 1.5 seconds. Is that not a bit too short?

ribalba:
@ArneTR 
2) I now introduced the optional parameter `playwright_process_duration` that we can set if the server is clogged. 
3) increase to 3000

ArneTR:
Setting the timeout to 3 s instead of 1.5 s is probably a good idea. But why limit it that hard anyway? Why can it not be unlimited and we just enforce the limit on the total duration of the playwright execution?

2. Thanks for introducting a `playwright_process_duration ` parameter. I re-thought my previous statement and I believe we actually do not need that.
We can just use _measurement_flow_process_duration and create a dedicated user for the webNRG service. I think that makes i easier to maintain. If you feel so too please revert this part of the change. I will setup the user then. What total time do you want? 5 Minutes? 10 Minutes?

ribalba:
@ArneTR so now this has gone a full round. Now only remove the wrong parameter and changes the 60 seconds. I think for now 5 minutes should be enough for anything coming from webNRG

ArneTR:
Ok, I have released it on all machines.

Did 10 Minutes, as the user is merged with green-coding.ai and this timeout is sensible for both

PR #1837, opened 2026-08-25
fixing an error in which the disk provider would output 0

ribalba:
See comment in code for details

## Summary

- Preserve the last valid disk I/O counters when `io.stat` contains no usable rows.
- Add a regression test for empty `io.stat` output.
- Report zero only when no valid reading exists.
- Fail with an error if state allocation fails.


ArneTR:
@coderabbitai review

ArneTR:
The fix looks reasonable to me given that this is a real world scenario.

Where have you observed this? On the cluster? Or locally on your laptop?

I am not sure what would force a docker container to evict a disk volume from its cgroup ... why did it happen? What was the test you ran?

2. The test looks extremely flaky to me. Did you write this yourself?

I find this extremely hard to maintain given that it is in bash and creates cgroups on the local machine which are not properly cleaned up in case.
Also it is not integrated into the test suite. Where do you imagine this to run?

@ribalba

ribalba:
It happened on my machine. I just added the .sh file so you could reproduce it. As soon as you were ok with the PR I would have deleted it. I should have added that to the description

ArneTR:
Following up our discussion on the phone:

You said: Totem video player creates sub-threads which then also allocate disk space on their own. When these sub-threads are terminated also their entry in the io.stat is removed

So the fix works fine for one entry addded and then removed, which keeps  the initially accounted budget.

But my argument was: If there are constantly groups overlapping added and removed if at least one group still remains the total amount will always change and the guard clause here implemented will never trigger. we will then constantly have changing values.

In essence: the provider is not designed for rows being removed. It will mal-calculate. if that is really a real scenario we need to rework the provider.

Plese check with totem how many lines are created and pos some debug data of the contents of the io.stat file

ribalba:
Ok, so I put some more work into this. The problem is not that rows are added or deleted but that the `fscanf` gets an `EOF` at some stage which makes the loop exit. My fix did fix this as it would then just count as 0 but the problem is somewhere else. 

Pushed a new version that should fix this.

We should also check if other providers have the same problem

ArneTR (commented):
Thanks for looking up in detail what the issue is.

I sadly to not undertand it fully. Maybe a call is better.

But here some questions:
- You say fscanf came back with 4 results read, but they were empty bc the file was torn. How can that be?
- It is a startup problem only? If so I feel the check atm is heavy overhead then as we do this every time we read from the file. The fopen handle should be kept and only checked once
- The logic itself is very very hard to read. I understand it is complex but I would love to have it hidden behind a function in gmt-helpers that is idiomatic through it's name: fopen_debounce or fopen_with_retry or something

But as said: still what really is the problem is unclear to me so this is not a full review, just a comment. 
What would have helped me to understand is a test case with a static file that clearly shows a "torn" file. The tests atm are very opaque to me at least visually

ribalba:
Here is a script that tries to create the file
```
#!/usr/bin/env bash
#
# Captures the raw io.stat states a container's cgroup goes through, and saves
# every distinct one. This is how the fixtures next to this script were made.
#
# WHY A TORN FILE EXISTS AT ALL
#
#   io.stat is a kernel seq_file, rendered on read, one line per block device
#   the cgroup has touched. While a cgroup is being wired up, a device can be
#   attached to it before its I/O counters are. Reading in that window returns
#   a line that has the "major:minor" prefix and nothing behind it:
#
#       252:0
#
#   The file is not empty and not corrupt - it is a valid snapshot of a cgroup
#   that is halfway through being created. It lasts well under a second, which
#   is why it is only ever seen by something sampling continuously.
#
#   Restarting a container is the easiest way to force it: docker destroys the
#   scope and creates a new one at the SAME path, so a collector that resolved
#   that path at startup reads straight through the rebuild.
#
# Usage:  ./capture-torn-io-stat.sh [output_dir]
set -uo pipefail

OUT_DIR="${1:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/captured}"
NAME=gmt-io-stat-capture
WORKLOAD='while true; do dd if=/dev/urandom of=/tmp/f bs=1M count=8 oflag=direct 2>/dev/null; sleep 0.3; done'

command -v docker >/dev/null || { echo "SKIP: docker not available"; exit 0; }

cleanup() { docker rm -f "$NAME" >/dev/null 2>&1; return 0; }
trap cleanup EXIT

mkdir -p "$OUT_DIR"
docker rm -f "$NAME" >/dev/null 2>&1
docker run -d --name "$NAME" alpine sh -c "$WORKLOAD" >/dev/null 2>&1 \
    || { echo "SKIP: could not start container"; exit 0; }
CID="$(docker inspect --format '{{.Id}}' "$NAME")"
STAT="/sys/fs/cgroup/system.slice/docker-${CID}.scope/io.stat"

for _ in $(seq 1 50); do [ -e "$STAT" ] && break; sleep 0.1; done
[ -e "$STAT" ] || { echo "SKIP: $STAT never appeared (rootless docker uses a different path)"; exit 0; }

echo "watching $STAT"
( sleep 2; docker restart -t 0 "$NAME" >/dev/null 2>&1 ) &

seen=""
n=0
end=$((SECONDS + 8))
while [ "$SECONDS" -lt "$end" ]; do
    content="$(cat "$STAT" 2>/dev/null)"
    sum="$(printf '%s' "$content" | md5sum | cut -d' ' -f1)"
    case "$seen" in
        *"$sum"*) ;;
        *)
            seen="$seen $sum"
            n=$((n + 1))
            printf '%s' "$content" > "$OUT_DIR/state-$(printf '%02d' "$n").io.stat"
            rows="$(printf '%s' "$content" | grep -c 'rbytes=' || true)"
            lines="$(printf '%s' "$content" | grep -c . || true)"
            torn=$([ "$lines" -gt "$rows" ] && echo "  <<< TORN: $((lines - rows)) line(s) with no counters" || echo "")
            printf 'state %02d: %4d bytes, %d line(s), %d with counters%s\n' \
                   "$n" "${#content}" "$lines" "$rows" "$torn"
            ;;
    esac
    sleep 0.01
done
wait
echo
echo "wrote $n distinct state(s) to $OUT_DIR"
```

ribalba:
Claude just gave me a nice text 

`io.stat` is a kernel `seq_file`. It does not sit on disk; it is rendered at the
moment you read it, one line per block device the cgroup has touched:

```
259:0 rbytes=1449984 wbytes=3145728 rios=18 wios=24 dbytes=0 dios=0
252:0 rbytes=1449984 wbytes=3145728 rios=18 wios=24 dbytes=0 dios=0
```

While a cgroup is being wired up, a device can already be attached to it before
its I/O counters are. A read landing in that window returns the `major:minor`
prefix with nothing behind it, and then EOF — `torn.io.stat` in this directory,
six bytes, no trailing newline:

```
$ hexdump -C torn.io.stat
00000000  32 35 32 3a 30 20                                 |252:0 |
00000006
```

That file is not corrupt and it is not empty. It is a valid snapshot of a cgroup
that is halfway through being created. It lasts well under a second, so only
something sampling continuously ever sees it.

**How the provider used to fail on it.** The old parser drove the whole file
from one `fscanf` format that had to match a full row. On this content it gets
two conversions (`major`, `minor`) instead of four, the loop condition
`== 4` is false immediately, and the function returns `{0, 0}`. `rbytes`/`wbytes`
are cumulative, so publishing that zero rewound the series and
`disk_io_parse.py` then rejected the whole measurement for negative intervals.

This is a *different* failure from `no_discard_counters.io.stat`. There `fscanf`
returns 4 — all four assignments really did happen — but the tail of the format
does not match, so the stream is left mid-row and every later device is silently
dropped. One mode fabricates a zero; the other under-reports.

ArneTR:
Hmm, I understand it even less now. Maybe we should call here.

The code speaks very often of cgroups re-created. But that should never be the case. If a cgroup gets re-created it means the docker container gets re-created. But we do not have that. 
I am unsure if that is just a plain false debug analysis by Claude or legacy ...

metric_providers/disk/io/cgroup/container/source.c
```diff
@@ -25,20 +26,57 @@ static int user_id = -1;
...
+    ssize_t line_len;
 
-    FILE * fd = fopen(path, "r");
-    if ( fd == NULL) {
-        fprintf(stderr, "Error - Could not open path %s (%s) for reading. Maybe the container is not running anymore? Errno: %d\n", path, container_name, errno);
-        exit(1);
-    }
+    *torn = false;
 
-    while (fscanf(fd, "%u:%u rbytes=%llu wbytes=%llu rios=%*u wios=%*u dbytes=%*u dios=%*u", &major_number, &minor_number, &rbytes, &wbytes) == 4) {
```
ribalba, line 41: @ArneTR So the problem is that when it can not find the 4 values it exits the while loop and then it returns `0` and then whey the metric provider does the `.diff()` if generates negative values

ArneTR:
Looks good.

ArneTR:
Do you need this available on a server right now, or with the next roll-out in 1-2 weeks? Let me know

PR #1842, opened 2026-08-28
(feat): Adding longer timeouts and re-runs for frontend tests bc of flakyness

ArneTR:
This PR tries to combat the flakyness in the pipeline currently by two approaches:
- Increasing network timeouts in playwright
- Doing re-runs on frontend and API tests

We see strong latency spikes sometimes when the blacksmith servers are under high load. Since CPUs are oversubscribed this performance penalty leads to playwright / network requests failing.


ArneTR:
@coderabbitai review

PR #1854, opened 2026-09-07
(fix): Validate flow container targets

### LINKED ISSUE #1148: Check that service really exits
Currently if you use a container in the flow section that does not exist the benchmark will fail once you hit that step in execution. We should check this when checking the yml file.

barislabs:
## What changed

`SchemaChecker` now rejects a non-null `flow.container` unless it matches one of the runtime container names resolved from `services`. This reports the mistake during YAML validation instead of later during execution.

The check keeps ordinary service names valid when no override exists, accepts explicit Docker `container_name` values, and intentionally does not treat an overridden service key as a runtime container name. It guards `null` for compatibility with #1585 but does not add host-execution schema support here.

## Verification

Run from `tests/` unless noted otherwise:

- `PYTHONPATH=.. ../venv/Scripts/python.exe -m pytest --confcutdir=lib -q lib/test_schema_checker.py` — 13 passed
- `../venv/Scripts/python.exe -m py_compile ../lib/schema_checker.py lib/test_schema_checker.py` — passed
- `PYTHONPATH=. venv/Scripts/python.exe -m pylint lib/schema_checker.py tests/lib/test_schema_checker.py --reports=no --score=no` (from the repository root) — passed
- `git diff --check` — passed

The Docker-backed project harness was not run locally because Docker is unavailable on this Windows host; the targeted schema tests were isolated from Docker fixtures.

## AI assistance

Agentic AI Tools were used to help implement this change and write the tests. Verification results and limitations are documented above.

Fixes #1148

---
<a href="https://app.blacksmith.sh/green-coding-solutions/codesmith/green-metrics-tool/pr/1854"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791402198&installation_model_id=428550&pr_number=1854&repository=green-coding-solutions%2Fgreen-metrics-tool&return_to=https%3A%2F%2Fgithub.com%2Fgreen-coding-solutions%2Fgreen-metrics-tool%2Fpull%2F1854&signature=744fae61eb841c4bc467ad1bc56ddc967125e42e550c8fcbeb8821a088e9eae8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/lib/schema_checker.py
+++ b/lib/schema_checker.py
@@ -303,6 +303,11 @@ class SchemaChecker():
                 raise SchemaError(f"The 'name' field in 'flow' must be unique. '{flow['name']}' was already used.")
             known_flow_names.append(flow['name'])
 
+            if flow['container'] is not None and flow['container'] not in known_container_names:
+                raise SchemaError(
+                    f"Flow '{flow['name']}' references unknown container '{flow['container']}'."
+                )
+
             for command in flow['commands']:
                 if command.get('read-sci-stdout', False) and not command.get('log-stdout', True): # log-stdout is by default always on. This is why we set default to True
                     raise SchemaError(f"You have specified `read-sci-stdout` in flow {flow['name']} but set `log-stdout` to False, which prevents log capturing.")
--- a/tests/data/usage_scenarios/duplicate_phase_name.yml
+++ b/tests/data/usage_scenarios/duplicate_phase_name.yml
@@ -11,21 +11,21 @@ services:
 
 flow:
   - name: This phase is ok
-    container: highload-api-cont
+    container: number-1
     commands:
       - type: console
         command: echo "asd"
         shell: bash
         note: Starting a little stress
   - name: This phase is also ok
-    container: highload-api-cont
+    container: number-1
     commands:
       - type: console
         command: echo "asd"
         shell: bash
         note: Starting a little stress
   - name: This phase is ok
-    container: highload-api-cont
+    container: number-1
     commands:
       - type: console
         command: echo "asd"
--- a/tests/data/usage_scenarios/include_example.yml
+++ b/tests/data/usage_scenarios/include_example.yml
@@ -2,7 +2,3 @@
 name: Name as is
 author: Author as is
 description: Description as is
-
-services:
-  test-container-1:
-    image: alpine
--- a/tests/data/usage_scenarios/include_example.yml
+++ b/tests/data/usage_scenarios/include_example.yml
@@ -2,3 +2,7 @@
 name: Name as is
 author: Author as is
 description: Description as is
+
+services:
+  test-container-1:
+    image: alpine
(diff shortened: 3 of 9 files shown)


ribalba (approved):
looks like a good addition to me. Weird that we don't check this already

ArneTR:
Hey @barislabs 

thanks for handing in a PR ❤️

The fix looks good. I applied a merge to make it compatible to the Shell Command PR just merged. Please have a look, but I think functionality should still be given.

2. Just out of curiosity: Did you ever encounter this issue while using the tool? Or did you just stumble over the issue and wanted to contribute?

We are always curious how people use our tool. So if want we are happy if you share some stuff that you are currently doing with GMT and where we can improve :)

barislabs:
Thanks, Arne! I found the issue while looking for somewhere useful to contribute, rather than running into it in my own usage. I don’t have a real-world GMT workload to share yet, but measuring software’s energy use looked like an interesting project to help with.

I checked the merged changes, and the container validation still works alongside host execution. I also caught some existing test fixtures that referenced missing or overridden container names. Those are fixed now without relaxing the validation.

The schema tests and targeted compatibility checks pass locally. I couldn’t run the Docker-backed integration tests on this Windows machine, and the new CI run needs maintainer authorization. Could you approve it when you have a moment?

ArneTR:
Hey @barislabs 

thanks for the feedback. Nice that you picked us for a open source contribution in your free time. Many thanks!

The fixes you pushed are mostly all reasonable. They would not break the tests per-se as tests trigger before the inconsistenscy is probed, but def. increase internal consistency if one just reads the tests!

One note though: Can you please revert changes on tests/data/usage_scenarios/include_example.yml
This is by design an incomplete usage_scenario file which is included in a different file. Then the services get merged in.
(for reference: the include happens in tests/data/usage_scenarios/overwrite_string_from_include.yml FYI)

Will run the tests after that!

barislabs:
Oh, don’t worry about it—this is a worthy cause! I’ve put `include_example.yml` back the way it was. The two scenarios that use it now define their own service instead. Both work with the container validation, and the local checks pass.

ArneTR:
Thanks for the PR! merging in :)

PR #1855, opened 2026-09-08
(fix): Time-weight step function MEANs by the value at each interval

ribalba:
Fixes the -1 "error" in the 99ms pull request Github is currently not showing me
Here it is https://github.com/green-coding-solutions/green-metrics-tool/pull/1825/


ArneTR:
Very very hard to read. Did you write the comment yourself?

I feel it is super verbose, has no examples, and somehow repeats concepts. Feels extremly AI and hard to read:

- If it is: I would really argue to limit AI comments. I feel we have brought in some in the last PRs and they hardcore clobber the code. Code becomes less idiomatic and rather relies on heavy comments, which will slow us down in the future when we refactor code.
- If it is not: Then I would be really suprised :). But still would then argue for more examples in such complex cases.

Happy to discuss. Code looks fine to me.

PR #1864, opened 2026-09-14
Adds the option to authenticate github requests

ribalba:
https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2026-03-10#authenticating-in-a-github-actions-workflow-using-curl


ArneTR:
I am unsure if we should bring this in.

The reason being is that the GitHub API only doubles the request with authenticated requests. Typically we never run out of requests with normal usage but when we make many measurements we also run out of budget even with the API.

So although I see it as a nice addition I also see the complexity and minimal risk of the token getting in the wild.
My gut feeling would prefer to keep the tool rather simple and not have this included as well. However the code change is also minimal ...

Let me know your thoughts after this comment.

In any case: if we bring it in the key should live under `cluster` as it is not related to security.

ribalba:
On: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2026-03-10#primary-rate-limit-for-unauthenticated-users
```
The primary rate limit for unauthenticated requests is 60 requests per hour.
```
and then for [Primary rate limit for authenticated users](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2026-03-10#primary-rate-limit-for-authenticated-users)
```
All of these requests count towards your personal rate limit of 5,000 requests per hour.
```

so it would be quite an improvement. What we can do is that the user holds the ssh key (as we already support) and like this the git pulls don't fail when someone has queued more than 60 jobs.

Will move it to cluster

ArneTR:
Oh my, thanks for the update.

Then either my info about the GitHub API was outdated or I mixed up something.

5k requests is really worth bringing this in.

Last change looks good. ty!

